PR #79, opened 2015-01-07
Polyline

patriciogonzalezvivo:
First round


bcamper:
Note: we lost some functionality in this change that needs to be restored before it can be merged. Namely:

* Handling of closed polygons outlines, which form a continuous loop (instead of having caps on the first and last points). See `closed_polygon` flag.
* Automatic removal of line segments falling on tile edges, which is needed for correctly rendering polygon outlines, since many logical polygons have been split into several polygons across tile boundaries (we are basically visually reassembling it to appear as a single polygon). See `remove_tile_edges` flag.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,178 @@ GLBuilders.buildPolylines = function (
...
+        if (lineSize < 2) {
+            continue;
+        }
 
-        vertices.push(
-            pb_inner, pb_outer, pa_inner,
-            pa_inner, pb_outer, pa_outer
-        );
+        //  Initialize variables
+        var normPrevCurr = [0.0,0.0,0.0],  // Right normal to segment between previous and current m_points
```
bcamper: 0.0 not needed here, can just be 0.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,178 @@ GLBuilders.buildPolylines = function (
...
+            normCurrNext = [0.0,0.0,0.0],  // Right normal to segment between current and next m_points
+            rightNorm = [0.0,0.0,0.0];     // Right "normal" at current point, scaled for miter joint
+
+        var prevCoord = [], // Previous point coordinates
+            currCoord = [], // Current point coordinates
+            nextCoord = []; // Next point coordinates
+
+        //  Compute first pair of vertexes based on the second point of the line
+        currCoord = Vector.set( line[0] );
```
bcamper: For style consistency, please remove extra spaces in parens.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,178 @@ GLBuilders.buildPolylines = function (
...
+        //  b) add the pair vertex into the vertex buffer
+        if (scaling_index) {
+            //  b.a If scaling is on add the vertex (the currCoord) and the scaling Vecs (normals pointing where to extrude the vertexes)
+            vertices.push(currCoord);
+            vertices.push(currCoord);
+            scalingVecs.push( rightNorm );
+            scalingVecs.push( Vector.neg(rightNorm) );
+        } else {
+            //  b.b Add the extruded vertexes
+            vertices.push([ currCoord[0] + rightNorm[0] * halfWidth,
```
bcamper: Ditto extra spaces in parens.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,178 @@ GLBuilders.buildPolylines = function (
...
+        }
 
-        // Add UVs
+        // c) Add uv's if they are enable
         if (texcoord_index) {
-            texcoords.push(
-                [min_u, min_v], [max_u, min_v], [min_u, max_v],
-                [min_u, max_v], [max_u, min_v], [max_u, max_v]
-            );
+            texcoords.push( [max_u, min_v], [min_u, min_v] );
```
bcamper: Is this right? 3 references to min but only 1 to max? (I might just not see what's happening here.)

src/gl/gl_builders.js
```diff
@@ -145,213 +145,178 @@ GLBuilders.buildPolylines = function (
...
+            addVertex(2*i+2);
+            addVertex(2*i+1);
+            addVertex(2*i+0);
+
+            addVertex(2*i+2);
+            addVertex(2*i+3);
+            addVertex(2*i+1);
         }
 
+        // // Add triangles indices
```
bcamper: Can remove this commented code.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,178 @@ GLBuilders.buildPolylines = function (
...
+        //     indices_data.push(vertexData_Offset + 2*i+2);
+        //     indices_data.push(vertexData_Offset + 2*i+1);
+        //     indices_data.push(vertexData_Offset + 2*i);
+        //     indices_data.push(vertexData_Offset + 2*i+2);
+        //     indices_data.push(vertexData_Offset + 2*i+3);
+        //     indices_data.push(vertexData_Offset + 2*i+1);
+        // }
+
+        // Clean all
+        vertices.length = 0;
```
bcamper: See other comments, please use `= []` instead (and move to top of this loop).

src/vector.js
```diff
@@ -2,6 +2,107 @@
...
+        // Divide to vectors
+        var len = Math.min(v1.length,v2.length);
+        for (i = 0; i < len; i++) {
+            v[i] = v1[i] / v2[i];
+        }
+    }
+    return v;
+};
+ 
+// Get 2D perpendicular FAST
```
bcamper: 1. Is this used anywhere?
2. Why the FAST comment?
3. It looks like it won't work, since the arguments are vectors and you're doing subtraction on the objects directly (remember, this isn't C++ :)

src/vector.js
```diff
@@ -2,6 +2,107 @@
...
+    return [ v2[1] - v1[1], 
+             v1[0] - v2[0] ];
+};
+
+// Get 2D vector rotated 
+Vector.rot = function (v, a) {
+    var V = v;
+    var vr = Vector.length(v);
+    var va = Vector.angle(v);
+    V[0] = vr * Math.cos(va+a);
```
bcamper: Ditto as above, these are JS objects, you don't have any overloaded operators.

src/vector.js
```diff
@@ -2,6 +2,107 @@
...
+};
+ 
+// Get 2D perpendicular FAST
+Vector.perp = function (v1, v2) {
+    return [ v2[1] - v1[1], 
+             v1[0] - v2[0] ];
+};
+
+// Get 2D vector rotated 
+Vector.rot = function (v, a) {
```
bcamper: Objects are passed by reference, this is going to overwrite the contents of the argument vector `v`. Set `V = []` to make a new vector instead.

src/vector.js
```diff
@@ -53,6 +154,17 @@ Vector.cross  = function (v1, v2)
     ];
 };
 
+// Dot product of two vectors
+Vector.dot = function (v1, v2)
+{
+    var n = 0.0;
```
bcamper: 0.0 is superfluous (very nitpicky I know, just FYI).

src/vector.js
```diff
@@ -2,6 +2,107 @@
...
+    var V = v;
+    var vr = Vector.length(v);
+    var va = Vector.angle(v);
+    V[0] = vr * Math.cos(va+a);
+    V[1] = vr * Math.sin(va+a);
+    return V;
+};
+
+// Get 2D heading angle
+Vector.angle = function (v) {
```
iwillig: Destructing allows you to write this function like this in more compact manner,
```javascript
Vector.angle = function ([x, y]) {
    return Math.atan2(x, y);
};
```
There should be no overhead to form.

src/vector.js
```diff
@@ -53,6 +154,17 @@ Vector.cross  = function (v1, v2)
     ];
 };
 
+// Dot product of two vectors
+Vector.dot = function (v1, v2)
+{
+    var n = 0.0;
+    var lim = Math.min(v1.length, v2.length);
```
bcamper: While in some ways better to check min length of each argument (avoids any invalid accesses), it's also kind of weird, because it's not really valid to call this with arguments of mismatching length anyway, is it? For functions that might be called frequently in inner loops, we would often not carefully check arguments, but instead assume that the calling code is going to call it correctly, and just let the function throw an error if there is an invalid access. So in that case you could just use the length of the first argument and assume they are the same. I think the same may apply to some other functions above where you're checking the min of multiple arguments too?

bcamper: @iwillig your opinion here?

src/vector.js
bcamper: Good point - @patriciogonzalezvivo destructuring is a new feature of ES6 (you may have seen it in python): https://github.com/lukehoban/es6features#destructuring

src/vector.js
iwillig: I agree with Brett, it seems fine to just use the length of the first argument. One thing to mention is that JavaScript Arrays do not throw out of bound errors for element they don't have. They return undefined, so 

```javascript
var a = [1,2,3,4];
a[10] === undefined; // true
```
So if you do have a shorter Array as the first argument, the multiplication operation will produce a 'NaN'.

iwillig: It actually gets better, because testing to see if you have a NaN value is hard. In JS
```javascript
NaN === NaN; // false
NaN == NaN; // false
```
The correct way to detect if you have a NaN value is to use Object.is, now. I think this is new.
```javascript
Object.is(NaN, NaN); // true
```

src/gl/gl_builders.js
```diff
@@ -145,213 +145,149 @@ GLBuilders.buildPolylines = function (
...
-                    else if (!edge1) {
-                        buildSegment(anchors[p][0], anchors[p][1]);
-                    }
-                    else if (!edge2) {
-                        buildSegment(anchors[p][1], anchors[p][2]);
-                    }
-                }
-            }
+    // Add to equidistant pairs of vertices
+    function addVertexPair ( coord, normal, v_pct ) {
```
iwillig: Spacing, normally we don't put spaces around the function arguments.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,149 @@ GLBuilders.buildPolylines = function (
...
-            if (len_sq > (width * width * miter_len_max * miter_len_max)) {
-                line_debug = 'distance';
-                intersect_outer = Vector.normalize([intersect_outer[0] - joint[0], intersect_outer[1] - joint[1]]);
-                intersect_outer = [
-                    joint[0] + intersect_outer[0] * miter_len_max,
-                    joint[1] + intersect_outer[1] * miter_len_max
-                ];
-            }
+        // Do this with the rest (except the last one)
+        for(i = 1; i < lineSize - 1 ; i++){
```
iwillig: Missing a space here.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,149 @@ GLBuilders.buildPolylines = function (
...
+                            [ min_u, (1-v_pct)*min_v + v_pct*max_v ]);
         }
     }
 
-    // Add vertices to buffer
-    for (var v=0; v < vertices.length; v++) {
-        vertex_template[0] = vertices[v][0];
-        vertex_template[1] = vertices[v][1];
+    // Add a vertex based on the index position into the VBO 
+    function addIndex ( index ) {
```
iwillig: Not sure if this is important, but defining a function within another function means that on every function invocation we are going to be creating a new function object. 

Normally this is not a huge deal in JS, but you mentioned that some of this code is in a tight inner loop.

src/gl/gl_builders.js
```diff
@@ -145,213 +145,149 @@ GLBuilders.buildPolylines = function (
...
-
-                pb_inner[1], pb_outer[1], pb_inner[0],
-                pb_inner[0], pb_outer[1], pb_outer[0]
-            );
+
+        rightNorm = Vector.normalize( normCurrNext );
+        addVertexPair(nextCoord, rightNorm, 1 );
+
+        // Add vertices to buffer acording their index
+        for (i = 0; i < lineSize - 1; i++) {
```
iwillig: I noticed that you reuse the `i` variable in this function. Feel free to use `let` instead of `var`. `Let` respects block scoping while `var` only understands function scope. 

https://github.com/google/traceur-compiler/wiki/LanguageFeatures#block-scoped-binding
I believe this is turned on by default now in traceur

patriciogonzalezvivo: great tip! thanks

bcamper: Except I think that traceur emulates this with try/catch blocks, and they warn it is slow? :/  I've used it for non-performance-critical code, but have left it out of frequently called functions.

iwillig: Yeah I agree that we should not use features that could cause tangram to be slow. I looked into how they implemented `let` in traceur. For the following code, 
```javascript
let a = [1, 2, 3, 4];
let i = 10;
for (let i in a) {
    console.log(i);
}

function do_something() {
  console.log(foo); // ReferenceError
  let foo = 2;
}

if (true) {
    let x = 'a';
    console.log(x);
    if (true) {
        let x = 'b';
        console.log(x);
        if (true) {
            let x = 'c';
            console.log(x);
            if (true) {
                let x = 'd';
                console.log(x);
            }
        }
    }
}
```
traceur generates 
```javascript
System.registerModule("test.js", [], function() {
  "use strict";
  var __moduleName = "test.js";
  var a = [1, 2, 3, 4];
  var i = 10;
  for (var i$__0 in a) {
    console.log(i$__0);
  }
  function do_something() {
    console.log(foo);
    var foo = 2;
  }
  if (true) {
    var x = 'a';
    console.log(x);
    if (true) {
      var x$__1 = 'b';
      console.log(x$__1);
      if (true) {
        var x$__2 = 'c';
        console.log(x$__2);
        if (true) {
          var x$__3 = 'd';
          console.log(x$__3);
        }
      }
    }
  }
  return {};
});
System.get("test.js" + '');
```
I need to make sure that traceur always uses this method for compiling `let`, but if so then it should be no or minimal overhead.
 
I can look into this more tomorrow as there might be situations where they need try and catch.

bcamper: Oh nice, that's great. I thought I read the catch/try thing on the traceur docs, but that was also before they enabled it by default. Perhaps they switched to this method.

demos/styles.yaml
```diff
@@ -257,7 +257,7 @@ layers:
             filter: water
 
         style:
-            # name: water
+            name: water
```
bcamper: Not sure why this change ended up here but could you revert? I turned water animation off by default in the main demo.

src/gl/gl_builders.js
```diff
@@ -134,9 +134,59 @@ GLBuilders.buildExtrudedPolygons = function (
...
+
+    // c) Add uv's if they are enable
+    if (texcoords) {
+        texcoords.push( [ max_u, (1-v_pct)*min_v + v_pct*max_v ],
+                        [ min_u, (1-v_pct)*min_v + v_pct*max_v ]);
+    }
+}
+
+// Add a vertex based on the index position into the VBO 
+    function addIndex (index, { vertex_data, vertex_template, halfWidth, vertices, scaling_index, scalingVecs, texcoord_index, texcoords }) {
```
bcamper: Remove extra indentation.

src/gl/gl_builders.js
```diff
@@ -134,9 +134,59 @@ GLBuilders.buildExtrudedPolygons = function (
...
+        //  Add vertex to VBO
+        vertex_data.addVertex(vertex_template);
+    }
+
 // Build tessellated triangles for a polyline
-// Basically following the method described here for miter joints:
-// http://artgrammer.blogspot.co.uk/2011/07/drawing-polylines-by-tessellation.html
 GLBuilders.buildPolylines = function (
     lines,
     z, width,
```
bcamper: Remove `z` argument (not used anymore).

src/gl/gl_builders.js
```diff
@@ -145,213 +195,99 @@ GLBuilders.buildPolylines = function (
...
-
-                pb_inner[1], pb_outer[1], pb_inner[0],
-                pb_inner[0], pb_outer[1], pb_outer[0]
-            );
+            addIndex(2*i+2, constants);
+            addIndex(2*i+3, constants);
+            addIndex(2*i+1, constants);
         }
 
+        // for (var i = 0; i < lineSize - 1; i++) {
```
bcamper: Remove commented code.

src/gl/gl_builders.js
```diff
@@ -134,9 +134,59 @@ GLBuilders.buildExtrudedPolygons = function (
     }
 };
 
+// Add to equidistant pairs of vertices
+function addVertexPair (coord, normal, v_pct, { halfWidth, vertices, scalingVecs, texcoords, min_u, min_v, max_u, max_v }) {
```
bcamper: Maybe add a comment for these functions indicating that they are internal helper methods for the polyline builder. Maybe put them after the main polyline builder method.

src/gl/gl_builders.js
```diff
@@ -145,213 +195,99 @@ GLBuilders.buildPolylines = function (
...
-        pb,
+        scalingVecs = scaling_index && [],   //  vertices directions for then (on GLSL vertex shader) extrude
+        texcoords = texcoord_index && [],
+        halfWidth = width/2,
         num_lines = lines.length;
-
+        
     var [[min_u, min_v], [max_u, max_v]] = texcoord_scale || [[0, 0], [1, 1]];
 
+    var constants = { vertex_data, vertex_template, halfWidth, vertices, scaling_index, scalingVecs, texcoord_index, texcoords, min_u, min_v, max_u, max_v };
```
bcamper: Maybe add a comment like:
`// Values that are constant for each line and are passed to helper functions`

src/gl/gl_builders.js
```diff
@@ -145,213 +195,99 @@ GLBuilders.buildPolylines = function (
...
     }) {
 
-    // Build triangles
+    // Build variables
     var vertices = [],
-        texcoords = [],
-        p,
-        pa,
-        pb,
+        scalingVecs = scaling_index && [],   //  vertices directions for then (on GLSL vertex shader) extrude
```
bcamper: Initialize these inside the top of the line loop instead.

src/gl/gl_builders.js
```diff
@@ -145,213 +195,99 @@ GLBuilders.buildPolylines = function (
...
+        // for (var i = 0; i < lineSize - 1; i++) {
+        //     indices_data.push(vertexData_Offset + 2*i+2);
+        //     indices_data.push(vertexData_Offset + 2*i+1);
+        //     indices_data.push(vertexData_Offset + 2*i);
+        //     indices_data.push(vertexData_Offset + 2*i+2);
+        //     indices_data.push(vertexData_Offset + 2*i+3);
+        //     indices_data.push(vertexData_Offset + 2*i+1);
+        // }
+
+        // Clean all
```
bcamper: Re-initialize these to `[]` at the top of the line loop instead.

src/gl/gl_builders.js
patriciogonzalezvivo: you mean line and lineSize?

patriciogonzalezvivo:
Polyline ready for review. Not caps and joins implemented yet

meetar:
So does this just silently replace the old line style? Do we have or want a way to specify eg endcap style in the scene def?

bcamper:
Yep this is just a different way of implementing what we had before. There
aren't any additions for caps and joints yet, but there should be soon!

PR #80, opened 2015-01-08
Automatically inject dependent uniform definitions into shaders

bcamper:
This PR is about simplifying the knowledge and amount of code needed to write custom shaders in the scene definition file.

The scene definition allows users to easily define uniforms in a YAML/JSON/JS object format, e.g.:

```
uniforms:
  scale: 0.5
  velocity: [1, 0.7, 0]
  texture: '/path/to/image.png'
```

Tangram will infer the correct types of each uniform and call the appropriate GL setter (e.g. a 3-element array like `velocity` is assumed to be a `vec3`). However, until now, users also had to separately define those uniforms within their shader. This PR adds code to automatically inject those uniforms into the shader code (also checking to see if they have already been defined, to avoid duplicate definitions). For example, when called for each of the uniforms in the block above, the `GLSL.defineUniform` method would generate the following GLSL code:

```
uniform float scale;
uniform vec3 velocity;
uniform sampler2D texture;
```

The system handles the following types:

- `float` (JS numeric value)
- `vec2`, `vec3`, `vec4` (JS array of length 2-4 holding numeric values)
- texture `sampler2D` (JS string, assumed to be a URL to an image)
- custom `structs` (arbitrary JS objects, parsing each individual property), including nested `structs`
- arrays of any of the above types, e.g. array of `vec4`, array of `struct`, etc.

Note: because the GLSL is not actually parsed into an AST or other tokenized form, but is instead simply manipulated as a string, it is possible that some of these operations are brittle (although so far, so good). For example, the `GLSL.isUniformDefined` and `GLSL.isSymbolReferenced` functions use regular expressions. This implementation can be replaced with a full-on parsing approach if the need is demonstrated.


test/style_spec.js
```diff
@@ -1,124 +1,119 @@
 import chai from 'chai';
 let assert = chai.assert;
-//import {parseLayers, walkRuleTree} from '../src/rules';
```
iwillig: Thanks for cleaning this up, I should have deleted it before merging.

bcamper: We can/should still add tests for those style rules, but in the meantime I added these tests that are more about style compilation and config.

iwillig: Yeah I agree and those test look great.

bcamper:
cc @meetar if you want to check this out

bcamper:
Planning to merge this today unless any concerns.

meetar:
looks good to me – standing by to update the demos

PR #81, opened 2015-01-09
Restore error messages from the worker threads

iwillig:


src/worker_broker.js
```diff
@@ -275,6 +275,7 @@ function setupWorkerThread () {
         }
         catch(e) {
             // Thrown errors will be passed back (in string form) to main thread
+            console.error(e);
```
bcamper: Did you mean to keep this here?

src/gl/shader_sources.js
```diff
@@ -1,22 +1,6 @@
 // Generated from GLSL files, don't edit!
```
bcamper: Weird this file changed, I think it just re-compiled in a different order this time? There aren't/shouldn't be any changes as far as I know.

bcamper:
Sweet!

src/worker_broker.js
iwillig: No, good catch.

src/gl/shader_sources.js
iwillig: Yeah that is strange... let me try to fix it.

iwillig:
Okay fixed the sharder source file

PR #82, opened 2015-01-10
Continuous zoom (naive version)

bcamper:
This adds continuous zoom support, but is a fairly naive implementation -- since there is no effort to improve the transition between zoom levels, you're more likely to see stark transitions, e.g. when the next zoom level loads and/or rebuilds. This should be improved in the future with a proxy tiles implementation, to draw co-located tiles from higher/lower zooms until the appropriate tiles have loaded. The Leaflet integration may also be a bit unstable, since it explicitly disables some leaflet zooming options to get a smoother frame-rate. But... it works fairly well currently.

This also adds up/down input controls to the demo, to test zooming.


bcamper:
cc @meetar

bcamper:
Planning to merge this today unless any concerns. Still work to be done for this to be the best experience but having the library have basic support for fractional zoom levels in the meantime is helpful.

meetar:
:shipit: excited to test this one out.

PR #83, opened 2015-01-13
Support multiple tile/data sources

iwillig:


iwillig:
@bcamper 

Hey Brett, I think this PR cover all or most of what we talked about. An open issue which I just discovered is that these changes broke the way we where handing the tile sources in the demo. We allowed people to pass in a different tile source object for testing purposes, but that is not going to work anymore. 

This issue is made complicated by the way that now the style objects themselves specify the source. In order to support dynamically swapping sources we would need an override or something. Not sure what the best solution is.

iwillig:
@bcamper 

Okay I believe that the demo is fix.

demos/main.js
```diff
@@ -4,74 +4,65 @@
...
-        rS;
-
-    var tile_sources = {
-        'mapzen': {
-            source: {
+    var
+        tile_sources = {
+            mapzen: {
                 type: 'GeoJSONTileSource',
-                url:  appendProtocol('//vector.mapzen.com/osm/all/{z}/{x}/{y}.json')
```
bcamper: I think we were doing this to properly support https iframing? It's annoying and messy, but will need to make sure we didn't break anything.

iwillig: Right.... K will fix.

demos/main.js
```diff
@@ -4,74 +4,65 @@
...
-        },
-        'mapzen-topojson': {
-            source: {
+            'mapzen-topojson': {
                 type: 'TopoJSONTileSource',
-                url: appendProtocol('//vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson')
-            },
-            scene: 'demos/styles.yaml'
-        }//,
-        // 'osm': {
```
bcamper: I'd like to keep these somewhere for testing.

demos/main.js
```diff
@@ -4,74 +4,65 @@
...
+            scene: scene_url,
+            preUpdate: preUpdate,
+            postUpdate: postUpdate,
+            logLevel: 'debug',
+            attribution: 'Map data &copy; OpenStreetMap contributors | <a href="https://github.com/tangrams/tangram" target="_blank">Source Code</a>'
+        });
+
+    layer.scene.subscribe({
+        loadScene: function (config) {
+            config.sources['osm'] = tile_sources[default_tile_source];
```
bcamper: If we move styles under sources, I guess we could still make this work by just updating specific keys instead of overwriting the whole object, yes?

demos/styles.yaml
```diff
@@ -397,7 +406,7 @@ layers:
 
     buildings:
         geometry:
-            source: osm
+            source: default
```
bcamper: I think this is left over from debugging?

demos/styles.yaml
```diff
@@ -453,7 +462,7 @@ layers:
 
     pois:
         geometry:
-            source: osm
+            source: default
```
bcamper: Ditto -- leftover?

demos/styles.yaml
```diff
@@ -9,6 +9,18 @@ camera:
 
     # type: flat
 
+sources:
```
bcamper: Can we not put this block in between camera and lighting? Those kinda go together. It could be before or after both.

makefile
```diff
@@ -34,6 +34,7 @@ clean:
 
 lint:
 	$(JSHINT) src/gl/*.js
+	$(JSHINT) src/styles/*.js
```
bcamper: ah, good catch - it's not recursive? :(

src/leaflet_layer.js
```diff
@@ -124,9 +123,12 @@ if (Utils.isMainThread) {
             });
 
             // Use leaflet's existing event system as the callback mechanism
-            this.scene.init().then(() => {
-                log.debug('Scene.init() succeeded');
+           this.scene.init().then(() => {
```
bcamper: Indentation is off.

src/scene.js
```diff
@@ -24,20 +24,22 @@ var vec3 = glMatrix.vec3;
 if (Utils.isMainThread) {
     // On main thread only (skip in web worker)
     Utils.requestAnimationFramePolyfill();
+
```
bcamper: Can we kill this extra line.

src/scene.js
```diff
@@ -24,20 +24,22 @@ var vec3 = glMatrix.vec3;
...
 // Load scene definition: pass an object directly, or a URL as string to load remotely
-export default function Scene(source, config_source, options) {
-
+export default function Scene(config_source, options) {
     options = options || {};
     subscribeMixin(this);
 
     this.initialized = false;
-    this.tile_source = source;
+    this.sources = new Map();
```
bcamper: Haven't checked yet, but there is a good chance this is going to break iOS / Safari. If so we'll need to avoid (unfortunately).

src/scene.js
```diff
@@ -78,11 +80,8 @@ export default function Scene(source, config_source, options) {
     log.setLevel(this.logLevel);
 }
 
-Scene.create = function ({source, config}, options = {}) {
-    if (!(source instanceof TileSource)) {
-        source = TileSource.create(source);
-    }
-    return new Scene(source, config, options);
+Scene.create = function ({config}, options = {}) {
```
bcamper: May be silly to have `config` inside an option at this point?

bcamper: Actually is this function still needed/desirable if it has no extra functionality?

src/scene.js
```diff
@@ -149,7 +149,7 @@ Scene.prototype.destroy = function () {
         });
         this.workers = null;
     }
-
+    this.sources.clear();
```
bcamper: Might have to ditch Map.

src/scene.js
```diff
@@ -270,9 +270,13 @@ Scene.prototype.setZoom = function (zoom) {
         this.removeTilesOutsideZoomRange(below, above);
     }
 
+
     this.last_zoom = this.zoom;
     this.zoom = zoom;
-    this.capped_zoom = Math.min(Math.round(this.zoom), this.tile_source.max_zoom || Math.round(this.zoom));
+
+    this.capped_zoom = Math.min(Math.round(this.zoom), this.findMaxZoom() || Math.round(this.zoom));
```
bcamper: Weill have to revisit max zoom for multiple sources, but this probably makes sense for now.

src/scene.js
```diff
@@ -270,9 +270,13 @@ Scene.prototype.setZoom = function (zoom) {
...
+    this.zooming = false;
```
bcamper: This duplicates the line at the top of the function, and I believe to work correctly there's a state order dependence that requires it to be there before `removeTilesOutsideZoomRange` is called.

src/scene.js
```diff
@@ -677,9 +681,28 @@ Scene.prototype.forgetTile = function (key) {
...
+};
+
+
+
 // Load a single tile
 Scene.prototype._loadTile = function (coords, options = {}) {
-    var tile = Tile.create({ coords: coords, tile_source: this.tile_source, worker: this.nextWorker() });
+    var tile = Tile.create({
+        coords: coords,
+        max_zoom: this.findMaxZoom(),
```
bcamper: Unsure of the ramifications of this, probably OK for now but I will look into it.

src/scene.js
```diff
@@ -899,6 +922,16 @@ Scene.prototype.reload = function () {
 
 };
 
+Scene.prototype.loadDataSources = function () {
+
+    for (var name in this.config.sources) {
+        var source = this.config.sources[name];
+        this.sources.set(name,
```
bcamper: Depends on if Map is OK

src/scene.js
bcamper: Max zoom is not really a "tile" property, it is a per source property.. but we don't have a clean way to handle different max zooms of multiple sources that will be co-mingled.

src/scene.js
```diff
@@ -911,6 +944,9 @@ Scene.prototype.preProcessSceneConfig = function () {
         }
     }
 
+    // TODO, find a home for this
+    this.loadDataSources();
```
bcamper: I think to function properly, this needs to go in `updateConfig`, which instantiates various objects after the scene config is parsed.

src/scene.js
```diff
@@ -983,12 +1019,11 @@ Scene.prototype.updateConfig = function () {
 Scene.prototype.syncConfigToWorker = function () {
     this.config_serialized = Utils.serializeWithFunctions(this.config);
     this.selection_map_worker_size = {};
-
     // Tell workers we're about to rebuild (so they can update styles, etc.)
+
```
bcamper: extra line

src/scene_worker.js
```diff
@@ -11,14 +11,20 @@ import {parseRules} from './rule';
 import {GLBuilders} from './gl/gl_builders';
 import GLTexture from './gl/gl_texture';
 
-export var SceneWorker = {};
+export var SceneWorker = {
+    sources: new Map(),
```
bcamper: Another Map warning.

src/scene_worker.js
```diff
@@ -31,19 +37,19 @@ if (Utils.isWorkerThread) {
     };
 
     // Starts a config refresh
-    SceneWorker.worker.updateConfig = function ({ tile_source, config }) {
-        SceneWorker.config = null;
```
bcamper: Where do these get reset now? They need to be cleared whenever `updateConfig` is called (it's basically a re-initialization step).

src/styles/polygons.js
```diff
@@ -5,9 +5,7 @@ import {StyleParser} from './style_parser';
 import gl from '../gl/gl_constants'; // web workers don't have access to GL context, so import all GL constants
 import GLVertexLayout from '../gl/gl_vertex_layout';
 import {GLBuilders} from '../gl/gl_builders';
-import GLTexture from '../gl/gl_texture';
```
bcamper: Thanks

demos/main.js
iwillig: Yeah i think we would have to do something along those lines.

src/tile.js
```diff
@@ -32,7 +31,7 @@ export default class Tile {
             }
         }, spec);
 
-        this.coords = this.tile_source.calculateOverZoom(this.coords);
+        this.coords = this.calculateOverZoom();
```
bcamper: TODO: better max zoom / overzoom treatment

src/tile.js
```diff
@@ -75,31 +74,39 @@ export default class Tile {
             { tile: this.buildAsMessage() })
         .then(message => {
             scene.buildTileCompleted(message);
+        }).catch(error => {
+            throw error;
```
bcamper: I wish promises would just to this automatically.

makefile
iwillig: Back when I setting this up I was having issues with the recursive option, but it seems to work now. 

Will fix.

src/tile.js
```diff
@@ -173,18 +180,19 @@ export default class Tile {
...
-            tile.vertex_data[m] = vertex_data[m].end().buffer;
-        }
+            // Finalize array buffer for each render style
+            tile.vertex_data = {};
+            for (var m in vertex_data) {
+                tile.vertex_data[m] = vertex_data[m].end().buffer;
+            }
 
-        tile.debug.rendering = +new Date() - tile.debug.rendering;
+            source.debug.rendering = +new Date() - tile.debug.rendering;
```
bcamper: mixes source. and tile. references

src/tile.js
```diff
@@ -75,31 +74,39 @@ export default class Tile {
...
 
     // Process geometry for tile - called by web worker
     // Returns a set of tile keys that should be sent to the main thread (so that we can minimize data exchange between worker and main thread)
     static buildGeometry (tile, layers, rules, styles) {
-        var name, layer, feature, style, feature_style;
+        var feature, style, feature_style;
         var vertex_data = {};
         var style_vertex_data;
 
-        tile.debug.rendering = +new Date();
```
bcamper: It's actually more important to me to track these metrics at a tile level. We could just track the tile, or track the sources and sum them at the end of this method.

src/scene.js
iwillig: Yeah, lets kill it.

src/scene.js
iwillig: I think that this was just a bad merge, let me look into it.

src/scene.js
iwillig: Cool.

src/tile.js
```diff
@@ -125,7 +132,7 @@ export default class Tile {
                         // Parse style
                         rule.name = rule.name || StyleParser.defaults.style.name;
                         style = styles[rule.name];
-                        feature_style = style.parseFeature(feature, rule, tile, context);
+                        feature_style = style.parseFeature(feature, rule, source, context);
```
bcamper: I was wondering why `source` was being substituted for `tile` here... then I went and looked at `parseFeature` and it seems like neither is needed anymore (the tile is spit out in an error message, but that could be omitted, or moved here).

src/scene_worker.js
iwillig: Good point, I will add them back.

src/tile.js
```diff
@@ -173,18 +180,19 @@ export default class Tile {
...
+
             }
-        }
 
-        // Finalize array buffer for each render style
-        tile.vertex_data = {};
-        for (var m in vertex_data) {
-            tile.vertex_data[m] = vertex_data[m].end().buffer;
-        }
+            // Finalize array buffer for each render style
```
bcamper: This should be outside this loop, after all sources are complete. Having it here may cause undefined results.

src/tile.js
```diff
@@ -195,27 +203,24 @@ export default class Tile {
     /**
         Retrieves geometry from a tile according to a data source definition
     */
-    static getGeometryForSource (tile, source) {
+    static getGeometryForSource (source, sourceConfig) {
```
bcamper: Might be clearer to called these `sourceData` and `sourceConfig`

src/tile.js
```diff
@@ -282,14 +287,29 @@ export default class Tile {
     // TODO: pass bounds only, rest of scene isn't needed
     updateVisibility(scene) {
         var visible = this.visible;
-        this.visible = this.isInZoom(scene) && Geo.boxIntersect(this.bounds, scene.bounds_meters_buffered);
+        this.visible = this.isInZoom(scene.capped_zoom) && Geo.boxIntersect(this.bounds, scene.bounds_meters_buffered);
         this.center_dist = Math.abs(scene.center_meters.x - this.min.x) + Math.abs(scene.center_meters.y - this.min.y);
         return (visible !== this.visible);
     }
 
     // TODO: pass zoom only?
```
bcamper: can remove this comment

src/tile.js
```diff
@@ -282,14 +287,29 @@ export default class Tile {
...
-    isInZoom(scene) {
-        return (Math.min(this.coords.z, this.tile_source.max_zoom || this.coords.z)) === scene.capped_zoom;
+    isInZoom(zoom) {
+        return (Math.min(this.coords.z, this.max_zoom || this.coords.z)) === zoom;
+    }
+
+    // TODO fix the z adjustment for continuous zoom
```
bcamper: I will verify if this works, if so we can remove this comment

src/tile.js
```diff
@@ -282,14 +287,29 @@ export default class Tile {
...
+    calculateOverZoom() {
```
bcamper: Hmm... overzoom really is a source-specific thing. Let's discuss how to keep it there.

src/tile_source.js
```diff
@@ -8,28 +8,15 @@ import log from 'loglevel';
 export default class TileSource {
 
     constructor (source) {
+        this.name = source.name;
         this.url_template = source.url;
         // overzoom will apply for zooms higher than this
         this.max_zoom = source.max_zoom || Geo.max_zoom;
     }
 
-    // TODO fix the z adjustment for continuous zoom
```
bcamper: See comment in Tile class... not sure we should move this out of here if possible.

src/tile_source.js
```diff
@@ -51,42 +38,42 @@ export default class TileSource {
...
-                var feature = tile.layers[t].features[f];
+                var feature = source.layers[t].features[f];
                 feature.geometry.coordinates = Geo.transformGeometry(feature.geometry, Geo.latLngToMeters);
             }
         }
 
-        if (tile.debug !== undefined) {
-            tile.debug.projection = +new Date() - timer;
+        if (source.debug !== undefined) {
+            source.debug.projection = +new Date() - timer;
```
bcamper: Since we have multiple places like this where we have a `source` object to put debug info on but we don't have a `tile` object, we should probably just sum (for those where sum is appropriate) all source debug values and roll them up into the tile.

src/tile_source.js
```diff
@@ -51,42 +38,42 @@ export default class TileSource {
...
-        for (var t in tile.layers) {
-            var num_features = tile.layers[t].features.length;
+    /**
+     Re-scale geometries within each tile to the range [0, scale]
+     TODO: clip vertices at edges? right now vertices can have
+     values outside [0, scale] (over or under bounds); this would
+     pose a problem if we wanted to binary encode the vertices in
+     fewer bits (e.g. 12 bits each for scale of 4096)
+    */
+    static scaleData (source, {coords: {z}, min}) {
```
bcamper: nice destructuring

src/tile_source.js
```diff
@@ -51,42 +38,42 @@ export default class TileSource {
...
-    // Re-scale geometries within each tile to the range [0, scale]
-    // TODO: clip vertices at edges? right now vertices can have
-    // values outside [0, scale] (over or under bounds); this would
-    // pose a problem if we wanted to binary encode the vertices in
-    // fewer bits (e.g. 12 bits each for scale of 4096)
-    static scaleTile (tile) {
-        for (var t in tile.layers) {
-            var num_features = tile.layers[t].features.length;
+    /**
+     Re-scale geometries within each tile to the range [0, scale]
```
bcamper: This comment is kind of out of date, since `scale` is no longer passed (`Geo.units_per_meter` reflects a value computed based on scale). I should update it.

src/tile_source.js
```diff
@@ -111,19 +98,31 @@ export class NetworkTileSource extends TileSource {
...
+        var url = this.formatTileUrl(tile);
+
+        if (tile.sources == null) {
+            tile.sources = {};
+        }
 
+        var source = tile.sources[this.name] = {};
+
+        source.url = url;
+        source.debug = {};
```
bcamper: source/tile debug info reminder, how to best combine

src/tile_source.js
```diff
@@ -131,29 +130,21 @@ export class NetworkTileSource extends TileSource {
...
+            }).catch((error) => {
+                source.error = error.toString();
                 reject(error);
             });
         });
     }
 
     // Sub-classes must implement this method:
-    parseTile (tile) {
+    parseSource (tile, source, reponse) {
```
bcamper: `parseSourceData`?

test/subscribe_spec.js
```diff
@@ -0,0 +1,98 @@
+import subscribeMixin from '../src/subscribe';
```
bcamper: Thanks :)

bcamper:
@iwillig thanks, seems this is all working (using the "minimum max zoom" can be a temporary thing until we decide how to properly handle that). I made a small change to the demo so that it only overrides the source if one is not already set in the scene definition -- it was confusing that it was overwriting settings in the scene def and I didn't want that to trip up anyone else. Now for testing you can define any sources you want in the scene def and it won't override those.

Looks like there's a test failure, other than that feel free to merge when you're ready.

iwillig:
Cool, 

thanks @bcamper

PR #84, opened 2015-01-14
Minimal event system

bcamper:
Looks like we are finally succumbing to the need to have some kind of event subscribe system. Here's an attempt at a minimal one.

### DIFF
--- a/src/subscribe.js
+++ b/src/subscribe.js
@@ -1,0 +1,25 @@
+export default function subscribeMixin (target) {
+
+    var listeners = new Set();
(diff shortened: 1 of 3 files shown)


src/subscribe.js
```diff
@@ -0,0 +1,25 @@
+export default function subscribeMixin (target) {
+
+    var listeners = new Set();
```
iwillig: Sets ftw

iwillig:
Looks, want to merge this?

PR #85, opened 2015-01-16
Upgrade pbf & mapbox vector tile parser to fix float endian encoding

bcamper:
Older versions of the `pbf` library were encoding floats with reversed endian order. We were compensating for this by also encoding in incorrect order! The dependent libraries have been fixed, so I'd like to upgrade and also change our own server-side encoding to  be correct.


bcamper:
This also drops use of a fork of mapbox's vector tile JS lib, and does the GeoJSON conversion inline instead.

PR #86, opened 2015-01-20
Cameras: support multiple named cameras, refactor matrices, support FOV param

bcamper:
- Support multiple cameras: `cameras` is a hash of camera configs
- If only one `camera` is defined (singular `camera: ...` syntax, it will be set to the default camera, e.g. `cameras.default` will point to the value of `camera`).
- An initial active camera can be designated with `active: true`.
- An initial `position` and `zoom` can be specified by the camera, this override's the map's existing view on scene load.
- Field-of-view (`fov`) is an alternative to `focal_length`. If the latter is present, it will take priority.
- Matrices are closer to traditional model-view-projection division, but still not quite in alignment with that model due to precision limitations (the translation of camera + tile needs to happen in one step, or we don't have enough 32-bit precision for mercator).


bcamper:
cc @meetar

PR #89, opened 2015-01-22
Add base to relative data source URLs

bcamper:
Appears to fix a cross-origin problem that only happens with data sources (not with other relative URLs) -- possibly because these requests are made from a worker?


bcamper:
The error without this is:

`XMLHttpRequest cannot load . Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https, chrome-extension-resource.`

meetar:
Working here: http://tangrams.github.io/osm-webgl-demos/demo1/

PR #91, opened 2015-01-26
Meetar/overzoom fix

meetar:
with future min_zoom option commented out


bcamper:
@meetar updated to use the tile.max_zoom that was already being calculated by the scene (missed this before, whoops).

PR #92, opened 2015-01-26
Use Mapbox's earcut library for tessellation

bcamper:
It's faster than libtess.


bcamper:
This exposed an issue with our mapbox vector tile encoding of multipolygons, which is incorrect. See https://github.com/mapzen/mapbox-vector-tile/issues/4.

libtess was able to figure this out and correct the data, but earcut requires correct rings.

bcamper:
Multipolygon issues resolved, good to merge.

mourner:
Nice! Did you experience any other issues/artifacts?

bcamper:
@mourner oh yes sorry, I've been meaning to let you know we switched! So far, so good :) I have not had any artifacts, but simply may not have exposed it to as much data as you have yet (it's also mostly used with our vector tiles which have a known ring order). Figured I would give it a try and will definitely let you know if/when I encounter any problems... thanks again for the library, so fast.

mourner:
@bcamper thanks! I've also added indexed output today, give it a try. :) May improve rendering performance (we didn't find a way to measure this yet though).

PR #94, opened 2015-02-03
Rename & move

bcamper:
This is a wide-scale renaming and file re-org. Been wanting to do it for a long time, but have been concerned about merging it with outstanding branches... so I've kept putting it off. But this seems to merge pretty cleanly with the ones I've tried, so going to just get it over with.

This also hopefully solves a long-standing `make` annoyance, where browserify would fail trying to generate a list of dependencies when the stringified shaders file is missing (even though it's an earlier ilsted build dependency, `make` calls out to browserify as part of its dependency generation; the easy solution was to create a temporary file and then delete it).

@meetar @iwillig this moves lots of things around but doesn't change functionality. Let me know if you see anything off.


bcamper:
This merged cleanly into one of @meetar's other branches, so I'm going to go ahead with it now.

PR #95, opened 2015-02-06
De-couples tile loading from Leaflet

bcamper:
Let Tangram calculate which tiles to load internally, to reduce coupling to Leaflet. 

This was motivated (at the current moment) by a desire to try different fractional zoom logic than what Leaflet does (which is to round to the nearest zoom level). Rounding is a good default, but other alternatives (like taking the zoom floor) should also be explored. 

De-coupling also makes future view features (rotation, tilt, etc.) possible (or at least easier to implement).

### DIFF
--- a/demos/styles.yaml
+++ b/demos/styles.yaml
@@ -299,7 +299,7 @@ layers:
                 width: |
                     function () {
                         return (
-                            zoom >= 16 &&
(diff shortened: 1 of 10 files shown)


bcamper:
Need to fix tests.

PR #99, opened 2015-02-23
Use unruly lib for feature rule matches

bcamper:
This is working properly now and as fast or faster than the previous implementation, so it looks good to merge to me.


bcamper:
@iwillig?

iwillig:
Lets do it !

PR #100, opened 2015-02-24
Generalize data sources + add non-tiled sources

bcamper:
Note: WIP, not ready to merge yet.

This generalizes data sources, no longer assuming that all sources are tiled. Everything inherits from a `DataSource` class, but tiled sources have an additional `tiled` flag set. This basically splits sources into two types: `tiled` sources, and what I called `object` sources.

Currently, `object` sources are assumed to be used for things like single complex objects or overlays, which are generally expected to be loaded once and then persistently displayed. This is the most naive approach, where object sources will *always* be rendered once loaded. It would however make sense to add some visibility/scene graph support to only render them when they are in the current view. 

Because object sources represent entire objects, there's no need for each worker thread to load every source (this is different than with tiled sources, where each worker owns a different set of tiles, but those tiles need to be loaded from all sources on each worker). The simple approach I came up with (for now) was to simply evenly distribute the object sources across the workers, e.g. if a scene has 4 object sources and 2 workers, each worker would be assigned 2 of the sources.


bcamper:
@iwillig would appreciate getting your thoughts on this.

bcamper:
Note: this does the initial reorg of data sources, introduces the concept of object sources, and loads those sources in the worker.

However, as of yet, no styling of geometry construction is being done for object sources. (For one thing, need to decide how we want to associate styles with object sources, and since they will often be added programmatically / on the fly, an appropriate JS API for them.)

src/scene_worker.js
```diff
@@ -44,8 +49,21 @@ if (Utils.isWorkerThread) {
...
         for (var name in config.sources) {
-            let source = config.sources[name];
-            SceneWorker.sources[name] = TileSource.create(Object.assign(source, {name}));
+            let source = DataSource.create(Object.assign(config.sources[name], {name}));
+            if (source.tiled) {
+                SceneWorker.sources.tiles[name] = source;
+            }
+            else {
+                // Distribute object sources across workers
+                if (source.id % SceneWorker.num_workers === SceneWorker.worker_id) {
```
bcamper: Thinking about this now, might make more sense to do this filter on the main thread and never send sources to the workers that they don't need.

src/data_source.js
```diff
@@ -1,28 +1,20 @@
 /*jshint worker: true */
-/*globals TileSource, topojson */
+/*globals DataSource, topojson */
 import Geo from './geo';
 import {MethodNotImplemented} from './utils/errors';
 import Utils from './utils/utils';
 import log from 'loglevel';
 
-export default class TileSource {
+export default class DataSource {
```
iwillig: I think DataSource make sense.

src/data_source.js
```diff
@@ -146,9 +121,65 @@ export class NetworkTileSource extends TileSource {
...
+        super(source);
+        this.type = 'GeoJSONSource';
+    }
+
+    formatUrl (dest) {
+        return this.url;
+    }
+
+    parseSourceData (tile, source, response) {
+        source.layers = { _default: JSON.parse(response) };
```
iwillig: This is new. Is the idea to hide and keep the original source data. I could see that being useful.

PR #101, opened 2015-02-25
Sprite & Labels

karimnaaji:
First stable version of labels, there is some exposed stylesheet properties concerning fonts available [here](https://github.com/tangrams/tangram/blob/labels/demos/styles.yaml#L559_L563). 

Non-discarded Labels gets rasterized once and stored into texture per tile. Two main rules have been introduced in order to render something consistent:
 - move in tile : the logic will try to move the label inside of the tile boundaries in order to reduce labels collisions. Bounding box test only for labels of the same tile keeps deterministic map design since tiles are fetched in arbitrary order.
 - keep in tile : whether the label should be directly discarded if it is outside of the tile, without trying to move it.

Point labels can't be moved for now, this would be extended to move them around their point. 
Line labels can move along the lines they are sticking to trying to find the most appropriate segment.

WIP rules : minimum distance to keep between labels, reduce redundant information & label priority (those would be exposed to the stylesheet), break long labels into several lines.


bcamper:
Nice :)

What about collision between point + line labels? Or between labels + sprites? Are we currently supporting that?

karimnaaji:
There should be collision between point and line labels but just noticed there's something wrong about it. There's not yet collision test with sprites but will add this to the new features.

src/styles/builders.js
```diff
@@ -584,6 +584,67 @@ Builders.buildQuadsForPoints = function (
...
+    points,
+    width, height, angle, scale,
+    vertex_data, vertex_template,
+    scaling_index,
+    { texcoord_index, texcoord_scale }) {
+
+    let w2 = width / 2;
+    let h2 = height / 2;
+    let scaling = [
+        [-w2, -h2],
```
bcamper: Just a reminder that if we need, we can add an `anchor` property that will make the sprite/label rotate around a position other than the center point. That can be added later though.

src/styles/text/text.js
```diff
@@ -0,0 +1,371 @@
...
+
+        style.discarded = discarded;
+
+        if (!discarded) {
+            let text_info = this.texts[tile][style_key][text];
+            this.texcoord_scale = text_info.texcoords;
+
+            style.text = text;
+
+            style.angle = rule_style.angle || 0;
```
bcamper: I think this is leftover from the old sprites implementation, now that the `Label` class is setting the angle, right?

src/styles/text/text.js
```diff
@@ -0,0 +1,371 @@
...
+            if (typeof style.angle === 'function') {
+                style.angle = style.angle(context);
+            }
+
+            // factor by which sprites scales from current zoom level to next zoom level
+            style.scale = rule_style.scale || 1;
```
bcamper: Ditto I think this is unused and would be set automatically in the future, I think we can remove here.

src/styles/text/text.js
```diff
@@ -0,0 +1,371 @@
...
+
+                for (let text in text_infos) {
+                    let text_info = text_infos[text];
+                    let label;
+                    let keep_in_tile;
+                    let move_in_tile;
+                    let geometry = this.geometries[tile][style][text];
+
+                    if (geometry.type === "LineString") {
+                        let exceed_heuristic = this.label_style.lines.exceed;
```
bcamper: `let` variables are scoped to the block, so `exceed_heuristic` isn't defined when referenced in the `discard` call below. You can move the `let exceed_heuristic` declaration outside this block, like `keep_in_tile` etc.

src/styles/text/text.js
karimnaaji: Yes! This is not used.

src/styles/text/text.js
```diff
@@ -0,0 +1,371 @@
...
+
+                return this.super.endData.call(this, tile_data);
+            });
+        });
+    },
+
+    // Override to queue features instead of processing immediately
+    addFeature (feature, rule, context, tile_data) {
+        // Collect text
+        let text = feature.properties.name;
```
bcamper: One thing I **would** like to do for v1 is to allow the field to use for the text to be specifiable in the stylesheet -- `name` was just an initial test, though it will probably be the most common value, and would be a good default. Following our typical style logic, I think a good behavior would be something like a `style.text_source` (name TBD, unsure what is best), that:

- defaults to feature `name` (`feature.properties.name`) when unset
- uses the value as the property name when set to a string, e.g. `style.text_source: value` would look for `feature.properties.value`
- can be set to a dynamic function, e.g. `style.text_source: function() { return (feature.value > 50 ? 'high' : 'low'); }` would set the text to either 'high' or 'low' depending on the feature.

@karimnaaji @meetar what do you think

karimnaaji: This seems like an really interesting feature for me, we could talk about this a little more to see some particular use cases but maybe @meetar would have more thoughts about it. Maybe also something like sprite replacement for specific words is something we could consider.

bcamper: Cool - let me know if you want to discuss more, it should be pretty easy to do this by extending our current style pattern, where we often have function-based properties.

src/styles/text/text.js
```diff
@@ -0,0 +1,377 @@
...
+            if (!this.texts[tile]) {
+                this.texts[tile] = {};
+            }
+
+            if (!this.geometries[tile]) {
+                this.geometries[tile] = {};
+            }
+
+            let style = this.constructFontStyle(rule);
+            let style_key = this.constructStyleKey(style);
```
bcamper: This could get messy as we add more style params, but might be the only reasonable solution anyway.

bcamper:
@karimnaaji I think this is good to merge for now, and track on-going additions separately!

karimnaaji:
Nice!

PR #103, opened 2015-02-26
Lights spherical att

patriciogonzalezvivo:


bcamper:
I don't think we want to merge this to master, but rather to `lights`, correct?

bcamper:
(Though it looks like the target branch can't be changed once the PR is opened, or at least I can't change it, so maybe you want to close and re-open.)

PR #105, opened 2015-02-27
Lights

patriciogonzalezvivo:
Ambient light
Directional
Point/Sphere light ( attExp, inner and outer radius )
Spotlight ( attExp, inner and outer radius )


patriciogonzalezvivo:
Hooray!!

PR #108, opened 2015-03-02
Iwillig/cancel xhr request

iwillig:


iwillig:
@bcamper 

I think this works, but it would be good if we had a second pair of eyes. I left the abort call in the freeResources method to keep the worker and main thread code synced up. This might not be the best long term solution. 

Thanks

src/scene.js
```diff
@@ -936,6 +936,8 @@ Scene.prototype.removeTile = function (key) {
         tile.remove(this);
     }
 
+    this.workers.forEach(x => WorkerBroker.postMessage(x, 'removeTile', key));
```
bcamper: We don't want to call `removeTile` for every worker because the tile only resides in one worker. The tile knows its worker (`tile.worker`), and has a method for posting messages to its worker (https://github.com/tangrams/tangram/blob/iwillig/cancel-xhr-request/src/tile.js#L112). It also looks like `Tile.remove()` already calls `removeTile` on the correct thread (https://github.com/tangrams/tangram/blob/iwillig/cancel-xhr-request/src/tile.js#L286). Interestingly though, `Tile.remove()` doesn't appear to call `Tile.freeResources()`. So looks like we don't need a worker-specific call here, and the `Tile.remove()` call above should cover it, but another look at how `freeResources` fits in may be needed? (Sorry if I missed something here.)

iwillig: Got it, did not notice that method. Yeah you are right that is much better.

src/tile.js
```diff
@@ -80,12 +80,17 @@ export default class Tile {
...
+
         if (this != null && this.meshes != null) {
             for (var p in this.meshes) {
                 this.meshes[p].destroy();
             }
         }
         this.meshes = {};
+
+        if (this.request != null) {
+            this.request.abort();
```
bcamper: If `freeResources` is only ever called on the main thread, then I'm pretty sure we can leave this out. This is one of those tensions around the main thread/worker thread relationship and how the object state is "split" between the two, so this is just another one of those places where it's a little weird (and could possibly be replaced with a better global pattern).

iwillig: Makes sense.

iwillig:
Okay, updated based on your feedback. From manually testing it seems like http requests are canceled. It would be good to have a second pair of eyes on it. 

Thanks.

bcamper:
@iwillig added some more aggressive logic, it seems to be working well. I think we're good to merge this whenever. (Ideally we'd add some tests, though getting the right events to trigger in order could be tricky.)

iwillig:
The tests are broken, but that is because of a sauce labs error. But the tests pass locally. I check locally and sauce labs is broken as well. 

I also check with their demo [test suite]( https://docs.saucelabs.com/tutorials/node-js/) and that seems to be broken as well.

I am going to bug their support about it.

PR #111, opened 2015-03-04
Babel

iwillig:
@bcamper, 

I upgraded 6to5 to babel, but that besides that everything seems to work. I don't know which is better, but it does seem like there is a strong community around babel.js. One thing I like about babel is that they support a nice node repl out of the box.

```shell
ivan@ceres ~/dev/mapzen/tangram (git)-[6to5] % ./node_modules/babel/bin/babel-node 
> let x = 10;
undefined
> let c = () => 10
undefined
> c
[Function]
> c.toString()
'function () {\n  return 10;\n}'
> import u from 'underscore';
undefined
> u
{ [Function]
```


bcamper:
@iwillig hey, lingering issues here with how this will play with `match-feature` and `unruly`, which use traceur and expect the runtime (at least I think unruly does, it looks like I got an error there).

Also, the traceur bug that was breaking iOS 8 is actually in unruly, so it turns out switching just here wouldn't help :/

Should we look at including the ES6 files from those modules and transpiling them here? (This would also raise the issue of how Babel should transform files in node_modules though?)

bcamper:
BTW I just merged mater to this branch to get some recent changes there

iwillig:
Right, of course. I looked into this other day, but got distracted. `match-feature` is written in plain JavaScript (funny note there.. I almost wrote pain javascript), so that should not be an issue. 

I guess we kind of have three options with unruly. I can convert it to plain JS, use babel or merge it back into tangram. Any of them seem fine to me. Let me know what you think is best.

PR #113, opened 2015-03-06
Move style textures to the scene level

bcamper:
This removes some style-oriented texture support that I was never entirely comfortable with up to the scene level. This should be more compatible with @patriciogonzalezvivo's forthcoming material textures (https://github.com/tangrams/tangram/pull/112), and follows a suggestion from @meetar to provide a centralized way to define textures in the scene config file.

For cases where the standard material texture isn't enough (e.g. styles like sprites, that aren't doing normal texture mapping + lighting), there are now 3 ways a texture can be added to a style:

## URL in Style
the simplest way to add a new texture is to simply refer to it by URL. For an example `icons` style, this can be done like this:

```
styles:
    icons:
        texture: path/to/image.png
    ...
```

This will automatically load `path/to/image.png`, with default texture settings, and make it available as a shader uniform named `u_texture`.

## Object in Style
If more texture configuration is needed (for instance, specifying sprite UVs or overriding the default texture filtering mode), an object can be provided directly in the style definition instead:

```
styles:
    icons:
        texture:
            url: path/to/image.png
            filtering: mipmap
            sprites:
                # each sprite is defined as: [x origin, y origin, width, height]
                bunny: [0, 0, 32, 32]
                fox: [0, 32, 32, 32]
```

## Centralized at Scene Level
If the same textures will be referenced in multiple places across the scene definition, it's also possible to define textures centrally, at the same top-level as `styles`, `sources`, etc. In the scene config, then later referenced by name by a style:

```
# define textures centrally
textures:
    icon_texture:
        url: path/to/image.png
        filtering: mipmap
        sprites:
            # each sprite is defined as: [x origin, y origin, width, height]
            bunny: [0, 0, 32, 32]
            fox: [0, 32, 32, 32]

# reference the texture by name in a style
styles:
    icons:
        texture: icon_texture
        ...
```

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -490,9 +490,9 @@
                     scene.config.layers.pois.style.sprite = 'tree';
                     scene.config.layers.pois.style.size = [[13, '16px'], [14, '24px'], [15, '32px']];
 
-                    this.state.bouncy = this.uniforms().bouncy;
-                    this.folder.add(this.state, 'bouncy').onChange(function(value) {
-                        this.uniforms().bouncy = value;
+                    this.state.rotate = this.uniforms().rotate;
+                    this.folder.add(this.state, 'rotate').onChange(function(value) {
+                        this.uniforms().rotate = value;
                         scene.requestRedraw();
                     }.bind(this));
                 }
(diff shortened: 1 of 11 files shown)


bcamper:
@meetar I don't think this should break any existing demos (except for the icon stuff which is updated in this PR), so I'm inclined to merge it soon.

meetar:
Sounds good to me, we can always patch if needed

PR #116, opened 2015-03-10
update build

meetar:


meetar:
color and spec haven't worked properly together since https://github.com/tangrams/tangram/commit/e345c4f28f3e9f51c1cea0a6725ec1f39f1f3bea -- with this build, the spec seems to be working as before, but when specular is defined, color information is apparently not picked up by the shader.

PR #124, opened 2015-03-20
Adding editor config file.

zzolo:
This is a real simple change to add an .editorconfig file.
http://editorconfig.org/

Basically, it just sets some code editing guidelines, like spacing that many people have different and standardizes it for editors that support it (most have plugins for it).  It is helpful with contributions so people don't waste time with talking about or fixing 2 spaces vs 4 spaces sort of issues.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# http://editorconfig.org
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


bcamper:
Nice, thanks :)

PR #125, opened 2015-03-20
Adding basic mechanism for altering tile data.

zzolo:
This is mostly a proof of concept.  Feel free to close after checking it out.

This just adds the ability to add an `alter_data` function to a data source that will be run before the data gets projected and rendered.

It's currently only for GeoJSON Tile layers since its not necessarily the best way to do this.

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -12,7 +12,11 @@
                 // You can edit the tile data here before it gets projected
                 // and rendered
                 return data;
-            }
+            },
+            scripts: [
+                // importScripts doesn't like the agnostic //example.com proto
+                'http://api.tiles.mapbox.com/mapbox.js/plugins/turf/v2.0.0/turf.min.js'
+            ]
         },
         'mapzen-dev': {
             type: 'GeoJSONTileSource',
--- a/demos/main.js
+++ b/demos/main.js
@@ -7,7 +7,12 @@
     var tile_sources = {
         mapzen: {
             type: 'GeoJSONTileSource',
-            url: window.location.protocol + '//vector.mapzen.com/osm/all/{z}/{x}/{y}.json'
+            url: window.location.protocol + '//vector.mapzen.com/osm/all/{z}/{x}/{y}.json',
+            alter_data: function(data) {
+                // You can edit the tile data here before it gets projected
+                // and rendered
+                return data;
+            }
         },
         'mapzen-dev': {
             type: 'GeoJSONTileSource',
--- a/src/data_source.js
+++ b/src/data_source.js
@@ -12,13 +12,30 @@ export default class DataSource {
         this.name = source.name;
         this.url = source.url;
 
-        // Get alter data.  Comes in as string, even if using actual
-        // function
+        // Get alter data.  Comes in as string when a web worker
         if (typeof source.alter_data === 'string' && source.alter_data.indexOf('function') !== -1) {
             source.alter_data = Utils.stringsToFunctions({ a: source.alter_data });
             source.alter_data = source.alter_data.a;
         }
         this.alter_data = source.alter_data;
+
+        // Since this process is done in a web worker, we can't actually
+        // pass data at time of processing, so handle it here.
+        this.extra_data = source.extra_data;
+
+        // Import any extra scripts
+        if (typeof importScripts === 'function' && source.scripts) {
+            source.scripts.forEach(function(s, si) {
+                try {
+                    importScripts(s);
+                    log.info('DataSource: loaded library: ' + s);
+                }
+                catch (e) {
+                    log.error('DataSource: failed to load library: ' + s);
+                    log.error(e);
+                }
+            });
+        }
 
         // overzoom will apply for zooms higher than this
         this.max_zoom = source.max_zoom || Geo.max_zoom;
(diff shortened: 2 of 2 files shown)


bcamper:
Thanks for following up on this! I'll take a look and think on how it would fit with other planned use cases. You're right that we'd probably want to make this generic so it would be available to any data source, but that should be doable and your approach does follow the patterns we use elsewhere (glad you figured out the whole function serialization thing, which is needed to "transfer" these functions to the web worker... it's not exactly heavily documented yet :)

Have you tried it on your particular use case?

zzolo:
I had never worked with Web Workers before.  This pull request is before I realized I could not share data or libraries with the web worker which is kind of what is needed for our use case.  Getting around the data part was easy enough, but trying to inject a library into the web worker is something I have not figured out yet.

I think we will be able to make it work for our for project, though.

The ideal for us would be a way to hook into a shared tile load from multiple sources, have access to something like Turf.js, alter the data, then render.  And though I think this could be a fairly good use case in general, given that most people don't want to host their own OSM tiles but might have some custom data to inject, I totally get that this is not a priority.

I think your serialization makes a lot of sense, btw; not sure if I used it correctly.  I noticed Parallel.js had some good techniques for communicating with workers too.
http://adambom.github.io/parallel.js/

bcamper:
Welcome to workers :)

Workers do provide a standard mechanism for importing scripts at run-time (similar to what a `<script>` tag injection would do): `importScripts(url)`. We use it here when instantiating the TopoJSON tile source type, so we don't need to include all of topojson in the library by default: https://github.com/tangrams/tangram/blob/master/src/data_source.js#L229

In theory, I think it would be OK to use `importScripts` inside a tile data transform callback (I'd have to read the spec again, but hopefully it is idempotent and could safely be called multiple times). If you're up for it, give it a try on your branch and let me know.

I have a couple ideas for simple changes to the PR, I'll get back to you with those soon (in the midst of trying to nail down some v0.1 features and work on documentation...). It would be great to get this integrated and have more people involved :)

zzolo:
Interesting.  I did try to user `importScripts` but ran into some issues, so I thought I misunderstood its purpose.  I'll try again.

What I was trying to add the following options to the data source class:

* `data_alter`: Function to alter data.
* `extra_data`: Any extra data to bring into the data_alter function.  This was our GeoJSON
* `extra_scripts`: Any extra scripts to include for processing.  This was turf.

I am not sure if this is really a great idea.  The alter function is definitely helpful, but the rest of it may be too much and not sure if others need that sort of thing.

zzolo:
Not sure what I was doing wrong before, but I did get the import scripts to work.

Again, I don't think this should be the final way to do this, but just a proof of concept.

bcamper:
Nice. Yeah I was at least wondering if there might be a reasonable workaround like that while we figure out what the more general/permanent approach is.

zzolo:
Another, maybe better, solution is to allow for custom data source classes.  This would allow someone to do whatever they want.

bcamper:
Sorry to leave this hanging. We are preparing for the first official Tangram release in the next couple weeks, so it would be good to decide to do something here or not.

My original plan was to support custom data source classes as you suggest, but our use of ES6 transpiler + Web Workers does complicate that a bit (compared to, say, a Leaflet plugin). It's a good goal, but I think the method you've used here of extending individual data sources works pretty well and is consistent with many other design decisions. I would make some minor changes, but overall I'm comfortable with the approach.

The main thing I am struggling with now is where to fit it in the pipeline. I see why you only added it for GeoJSON tiled sources initially. Since different data source types have fairly divergent internal data structures (text vs. binary responses, some have to project to mercator and some don't, some have to scale coordinates to an internal implementation-specific tile scale and some don't, etc.), it's more difficult to provide a consistent hook for transforming data.

- One option would be to only support this (initially) in GeoJSON, as you have it now.
- Another option would be to provide the hook in each data source, with the caveat that the incoming source data can be in a different state depending on the data source type (so a user would have to be aware of that and deal with it accordingly).
- We could also re-org the data source behavior to try creating a more normalized pipeline (e.g. always provide lat/lng GeoJSON to a custom data transform function, regardless of the underlying data source type), but I'm not inclined to go this route right now because it could introduce a lot of excess processing (e.g. the Mapbox vector tile source would have to project to lat/lng, call the transform, then project back, adding multiple additional steps).

Do you have any further thoughts or feedback based on your experience so far?

zzolo:
I agree.  I think GeoJSON initially is fine, but if/when broadening it to other formats, it should just be the data that is coming in as its raw format.

Otherwise, no other thoughts.  Good luck with the release.  Amazing project and work.

bcamper:
@zzolo Thanks again for the PR and sorry for the delay. I've incorporated your changes into a separate branch, with a few small changes:

- Re-named the `alter_data` option to `transform` (a little more consistent with some other terms in the library): https://github.com/tangrams/tangram/commit/4a06e3f614efd8f70c179f01a799b897ed914ee9
- Fix the order of operations for the worker serialization so the custom data transform function is automatically de-serialized, obviating the need to handle it in the data source constructor: https://github.com/tangrams/tangram/commit/2599e3f04b60bd4cac262198618b9fcaaa299eb8
- Small TopoJSON refactor so that those sources can also take advantage of this (since TopoJSON is essentially just a compression transport mechanism for GeoJSON, it's easy to pass decoded GeoJSON through to the transform function. (https://github.com/tangrams/tangram/commit/7273b7df85bc9728c49973bfcbd1b791f3ff4658)

Still need to address further reusability and making this kind of functionality available across other data formats (like MVT), but I think this is a reasonable first step.

Thanks and if you have any other ideas, please let us know, we'd love to have you as a regular contributor :)

zzolo:
Excellent!  Things are pretty busy with other projects at the moment, but hopefully I will find some time to check this out soon.

PR #135, opened 2015-04-20
Merge `points` and `sprites` styles

bcamper:
These two styles had a lot of duplication, and Sprites had become a more sophisticated version of Points. Now they are just Points, and can be rendered as a plain color (by default), or with a sprite/texture, by creating a point-based style with a `texture` field specified:

```
styles:
  icons:
    base: points
    texture: icons.png
    sprites:
      # each sprite is defined as: [x origin, y origin, width, height]
      thing1: [0, 0, 32, 32]
      thing2: [0, 32, 32, 32]
```

Then when drawing point geometry, in the layer style rules:

```
style:
  name: icons
  sprite: thing2 # select the sprite to draw with
```


bcamper:
@meetar some demos might need updating

PR #142, opened 2015-05-08
Restore support for old `outline` style parameter (for lines)

bcamper:
After switching to a system where all outlines were drawn as additional draw groups alongside the object they were outlining, we realized that the old system, where we had an explicit `outline` syntax, was really useful. In particular, it's useful for things like drawing a consistent outline width (e.g. 2px outline) on a line that is scaling between zoom levels. This PR:

- Refactors tile mesh construction so that styles can delegate to other styles (e.g. a line style can build its inner line, then calculate the appropriate outline width and call the same or another line style to build the outline geometry).
- Adds support for the `outline` parameter in the line base style.
- Simplifies the dynamic line scaling: now that base tile zoom is determined by `floor` instead of `round`, we only need to scale *up* towards the next zoom level, not both up *and* down towards the previous zoom level.

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -308,7 +308,7 @@ Enjoy!
             },
             'halftone': {
                 setup: function (style) {
-                    scene.config.background.color = 'black';
+                    scene.config.scene.background.color = 'black';
 
                     var layers = scene.config.layers;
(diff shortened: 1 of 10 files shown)


bcamper:
TODO: would like to add back the same outline capabilities for polygons too, though this is lower priority.

PR #159, opened 2015-06-19
Label priority

bcamper:
Adds the ability to specify label priority in draw rules in the scene file. This is done through the `priority` field. For example, to set different priorities on `places` vs. `landuse_labels` layers:

```
    places:
        data: { source: osm }
        draw:
            text:
                priority: 1
                font:
                    typeface: italic 11pt Helvetica
                    fill: black

    landuse_labels:
        data: { source: osm }
        draw:
            text:
                priority: 2
                font:
                    typeface: Italic 10pt Lucida Grande
                    fill: darkgreen
```

@sensescape @nvkelso @meetar:
Currently, **lower** priority values take precedence over higher ones. So `priority: 1` labels will be drawn before `priority: 2` labels. In English, I usually think of something that is "priority one" as being more important than something that is "priority two". On the other hand, "high priority" could also imply a higher number :) 

However, the current behavior is opposite of how our `order` field works, which is more a painter's style approach, where higher `order` value geometries are drawn on top of those with lower `order` values. We could leave `priority` as is, or make them consistent (for what it's worth, it's easier to insert important labels later on by simply giving them a higher priority value, vs. having to go back and re-number everything, so there are arguments both ways).

Feedback on which makes most sense and/or will be easiest to work with?

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -461,3 +461,18 @@ layers:
                     typeface: Italic 10pt Lucida Grande
                     fill: darkgreen
                     stroke: { color: white, width: 3 }
+
+    # building_labels:
+    #     data: { source: osm, layer: buildings }
+    #     filter:
+    #         name: true
+    #         any:
+    #             - { $zoom: { min: 17 }, height: { min: 50 } }
+    #             - $zoom: { min: 18 }
+    #     draw:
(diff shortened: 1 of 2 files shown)


bcamper:
TODO when merging: update docs

PR #161, opened 2015-06-27
Inlay blending mode

bcamper:
In some cases, it may be desirable to let labels, icons, or other items drawn with full alpha blending be occluded by other objects, such as buildings.

This branch adds a blending mode for that, called `inlay` (as opposed to `overlay`). The inlay blend has depth test on, but depth write off. This means that the inlay geometry is appropriately depth-positioned with regard to any previously drawn geometry, but will not guarantee correct depth between other geometry drawn with the inlay mode (e.g., multiple inlay geometries will overlay each other).

### Labels with `blend: overlay`

tangram-1435027367572 1

### Labels with `blend: inlay`

tangram-1435027416750 1

### DIFF
--- a/src/scene.js
+++ b/src/scene.js
@@ -662,6 +662,11 @@ export default class Scene {
         this.setRenderState({ depth_test: true, depth_write: false, alpha_blend: (allow_alpha_blend && 'multiply') });
         count += this.renderStyles(styles, program_key);
 
+        // Inlay styles: depth test on, depth write off, blending on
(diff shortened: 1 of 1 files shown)


meetar:
:+1: :shipit:

bcamper:
@blair1618 can't remember if we have implemented the blend modes in ES yet -- either way, this is one more, but only a slight config change from the others.

blair1618:
@bcamper Haven't implemented blend modes yet, added a card https://trello.com/c/o15jXnMI/557-blend-modes-in-stylesheet

PR #163, opened 2015-07-01
Pack vertex attributes

bcamper:
Packs some vertex attributes into smaller data types:

- position: floats -> signed shorts
- normal: floats -> signed bytes
- texture UVs: floats -> unsigned shorts
- line extrusion: floats -> signed shorts
- layer order: floats -> signed shorts

This produces about a 44% decrease in memory usage for VBOs in a typical complex scene (midtown Manhattan zoom 15).

### DIFF
--- a/src/styles/builders.js
+++ b/src/styles/builders.js
@@ -558,7 +558,7 @@ Builders.buildQuadsForPoints = function (
     width, height, angle, scale,
     vertex_data, vertex_template,
     scaling_index,
-    { texcoord_index, texcoord_scale }) {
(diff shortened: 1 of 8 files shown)


hjanetzek:
Have you seen http://cesiumjs.org/2015/05/18/Vertex-Compression/ ? Two bytes may suffice for normals. (In VTM I used the +z hemisphere variant mentioned in the cited paper as one cannot usually look below buildings)

bcamper:
@hjanetzek have not seen that, will check it out!

PR #164, opened 2015-07-06
Styles can enable GL extensions

bcamper:
Custom styles can now enable GL extensions, with a new `extensions` property (within `shaders`). For example, a style that uses the `OES_standard_derivatives` extension:

```
    grid:
        base: polygons
        lighting: false
        shaders:
            extensions: OES_standard_derivatives
            blocks:
                color: |
                    // From: http://madebyevan.com/shaders/grid/
                    // Pick a coordinate to visualize in a grid
                    vec3 coord = v_world_position.xyz / 10.;
                    // Compute anti-aliased world-space grid lines
                    vec3 grid = abs(fract(coord - 0.5) - 0.5) / fwidth(coord);
                    float line = min(min(grid.x, grid.y), grid.z);
                    // Just visualize the grid lines directly
                    color = vec4(vec3(1.0 - min(line, 1.0)), 1.0);
```

`extensions` is either a single extension name, or a list of one or more requested extensions, e.g. for multiple extensions:

` extensions: [OES_standard_derivatives, EXT_frag_depth]`

For each available extension, a `#define` with the following form will be set: `TANGRAM_EXTENSION_${name}`, e.g. 

`#define TANGRAM_EXTENSION_OES_standard_derivatives`

(Some extensions already set a `#define`, but by setting our own we also ensure consistency across extensions.)

If any of the extensions are not available on the client a warning will be printed to console. The style will still be compiled. However, by testing for the presence of enabled extensions (via `#define`, `#ifdef`, etc.), the shader can be written to be compatible on machines without the requested extension. If fallback isn't implemented and the style fails to compile, then it will not render (as with any other invalid rendering style).

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -133,6 +133,25 @@ styles:
         base: points
         texture: pois
 
+    grid:
+        base: polygons
+        lighting: false
+        shaders:
+            extensions: [OES_standard_derivatives]
+            blocks:
+                color: |
+                    // From: http://madebyevan.com/shaders/grid/
+                    // Pick a coordinate to visualize in a grid
+                    vec3 coord = v_world_position.xyz / 10.;
+
+                    // Compute anti-aliased world-space grid lines
+                    vec3 grid = abs(fract(coord - 0.5) - 0.5) / fwidth(coord);
+                    float line = min(min(grid.x, grid.y), grid.z);
+
+                    // Just visualize the grid lines directly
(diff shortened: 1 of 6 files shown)


bcamper:
TODO:
- allow `extensions` to accept a single value in addition to a list
- rebase on master
- add documentation branch/PR

bcamper:
@blair1618 can you give a look (at least over the description) to confirm this will work OK on native?

blair1618:
This shouldn't be any trouble to implement in -es. :+1:

PR #166, opened 2015-07-23
Multiple datasource layers

bcamper:
Allow scene layers to specify more than one layer to pull from the data source. The data from all layers is combined into one layer on the Tangram side.

Example applying the same label styling to `pois` and `buildings` layers in the data source:
```
layers:
    labels:
        data: { source: osm, layers: [pois, buildings] }
        filter: ...
        draw: ...
```


bcamper:
@nvkelso can't remember the specific case but I know you wanted this, maybe for POI demo?

PR #167, opened 2015-07-31
Add a `sprite_default` draw parameter to points

bcamper:
It turns out a common use case for sprites is to dynamically match the sprite name to a feature property. For example, many of our `kind` property values for POIs directly map to names of sprites in our texture. We can write a rule to match these:

```
draw:
   icons:
      sprite: function() { return feature.kind }
```

But! While that can eliminate hundreds or thousands of lines of style rules with this one rule, there are still times when we won't have a sprite for every possible feature `kind` value. This PR addresses that by adding a `sprite_default` property. When set, this property is used as the sprite value in cases where the value of `sprite` isn't found in the texture.

For example, we can make all sprites default to a generic icon, and only apply that rule at higher zooms, like this:

```
# add generic icon at high zoom, if direct match fails
generic-icons:
   filter: { $zoom: { min: 18 } }
   draw:
      icons:
         sprite_default: generic
```

These changes are particularly to support POI rendering, as seen in: https://github.com/tangrams/poi-labels-demo. That repo will be updated to use this feature.

### DIFF
--- a/src/styles/points/points.js
+++ b/src/styles/points/points.js
@@ -63,19 +63,30 @@ Object.assign(Points, {
             return null;
         }
 
-        style.sprite = rule_style.sprite;
-        if (typeof style.sprite === 'function') {
-            style.sprite = style.sprite(context);
+        let sprite = style.sprite = rule_style.sprite;
+        if (typeof sprite === 'function') {
+            sprite = sprite(context);
         }
+        style.sprite_default = rule_style.sprite_default; // optional fallback if 'sprite' not found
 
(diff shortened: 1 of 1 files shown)


bcamper:
cc @nvkelso working on this to simplify POI rules

bcamper:
@blair1618 I think this is a better solution than the one that uses an array of fields with fallback logic.

bcamper:
docs covered in https://github.com/tangrams/tangram-docs/pull/37

PR #169, opened 2015-08-03
Optionally enforce polygon winding order

bcamper:
While many GeoJSON/TopoJSON data sources (including Mapzen's tiles) are encoded with typical CCW-ordered outer rings (and CW inner rings), it turns out not all data sources follow this, and GeoJSON does not enforce it. In particular, user-supplied GeoJSON/TopoJSON files may have inverse ring orders depending on how/where they were generated.

To fix this, we can enforce CCW winding order. This PR adds an option to do that per-data-source. By default, **tiled** data sources do **not** enforce winding order (i.e. they are assumed to already be correctly wound); **non-tiled** data sources **do** enforce winding order (i.e. they are assumed to not be safe). (Defaulting off saves a bit of needless computation.)

These defaults can be overridden if desired, for example to reverse these options:

```
sources:
   osm:
        type: GeoJSON
        url:  http://vector.mapzen.com/osm/all/{z}/{x}/{y}.json
        enforce_winding: true # default for tiled sources is false
    schools:
        type: GeoJSON
        url: demos/data/school-districts-polygon.geojson
        enforce_winding: false # default for non-tiled sources is true
```

### DIFF
--- a/src/geo.js
+++ b/src/geo.js
@@ -235,3 +235,31 @@ Geo.multiPolygonArea = function (polygons) {
 Geo.ringWinding = function (ring) {
     return Geo.signedPolygonAreaSum(ring) > 0 ? 'CW' : 'CCW';
 };
+
+// Enforce winding order on outer/inner rings
+// winding: 'CW' or 'CCW'
+Geo.enforceWinding = function (geom, winding) {
+    let polys;
+    if (geom.type === 'Polygon') {
+        polys = [geom.coordinates];
+    }
+    else if (geom.type === 'MultiPolygon') {
+        polys = geom.coordinates;
+    }
+    else {
+        return geom;
+    }
+
+    for (let p=0; p < polys.length; p++) {
+        let poly = polys[p];
+
+        // If first ring winding doesn't match, reverse all rings
+        // NOTE: assumes ring winding orders already alternate as expected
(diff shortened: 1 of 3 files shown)


meetar:
So is enforce_winding for situations where the automatic winding order detection fails? If so, would it be any less accurate or useful to make it "reverse_winding" instead? (And regarding the computation, would it not be safe to test a few polygons and then assume the rest follow suit?)

bcamper:
When `enforce_winding` is `false`, no winding detection or correction is done. This is just to save computation, since there's no point in looping over the rings of every polygon when we know the order is already correct (such as in Mapzen's tile service). When `enforce_winding` is `true`, this detection is done, and corrected (reversed) only if necessary. So `reverse_winding` wouldn't necessarily be more accurate IMO, because the reversal is conditional (only done if needed). We're making an educated guess that it's best to do this enforcement for data sources we don't have much info on (e.g. standalone GeoJSON/TopoJSON), but to assume/hope it's not needed on tiled source, which tend to be generated by more specialized software. But these are assumptions, these defaults can be changed or overriden.

bcamper:
cc @burritojustice

meetar:
@bcamper docs entry for this element added to a new PR, awaiting the next release https://github.com/tangrams/tangram-docs/pull/37

bcamper:
@meetar thanks!

PR #170, opened 2015-08-05
Oriented bounding box

karimnaaji:
Use of oriented bounding box for labels, the collision detection between labels was only a broad phase with AABBs, this adds an extra narrow phase step and test collision against the OBB only when the AABB collides.


src/styles/text/label.js
```diff
@@ -10,41 +12,81 @@ export default class Label {
...
+            aabb: [],
             move_in_tile,
             keep_in_tile
         });
 
         this.id = Label.id++;
-
         this.buffer = this.buffer || 2; // TODO: make configurable
         this.buffer *= Geo.units_per_pixel;
+        this.keepMinDistance = true;
```
bcamper: Style in this codebase is `keep_min_distance` for variables, `camelCase()` for function/methods, (it's not my favorite but should stick to it for consistency).

src/styles/text/label.js
```diff
@@ -10,41 +12,81 @@ export default class Label {
...
 
-    occluded (bboxes) {
+    occluded (aabbs) {
         let intersect = false;
-
-        if (bboxes.length > 0) {
-            boxIntersect([this.bbox], bboxes, (i, j) => {
-                intersect = true;
-                return true;
+        let aabb = this.aabb;
```
bcamper: This isn't needed, `this` is in scope below where `aabb.obb` is referenced.

src/styles/text/label.js
```diff
@@ -10,41 +12,81 @@ export default class Label {
...
+
+        // Broadphase
+        if (aabbs.length > 0) {
+            boxIntersect([this.aabb], aabbs, (i, j) => {
+                // Narrow phase
+                if (OBB.intersect(aabb.obb, aabbs[j].obb)) {
```
bcamper: Can eliminate `aabb` local var and use `this.aabb` here.

src/styles/text/label.js
```diff
@@ -10,41 +12,81 @@ export default class Label {
...
         if (!intersect) {
-            bboxes.push(this.bbox);
+            aabbs.push(this.aabb);
         }
 
         return intersect;
     }
 
+    // keep a minimal distance between the labels
+    minDistance (aabbs) {
```
bcamper: Maybe make this method name a little more descriptive, I would have assumed it returned a min distance. Like `checkMinDistance()` or something.

src/styles/text/label_line.js
```diff
@@ -89,28 +90,18 @@ export default class LabelLine extends Label {
         return [ p1, p2 ];
     }
 
-    computeBBox (size) {
+    computeAABB (size) {
         let upp = Geo.units_per_pixel;
 
         let merc_width = this.size.text_size[0] * upp;
```
bcamper: This is a pre-existing issues, but these values are not actually mercator, so these are misleading variables names. It is an arbitrary scale/coordinate system local to the tile (the `Geo.units_per_pixel` provides that conversion). So we can just use some more generic name here. We should also rename (or just eliminate) the `Utils.pixelToMercator` method, since it is similar misnomer: https://github.com/tangrams/tangram/blob/obb/src/utils/utils.js#L437

karimnaaji: Ok I see, I'll rename those.

PR #171, opened 2015-08-05
GL Render states

karimnaaji:
Lazy setup for render states to reduce driver state changes.


bcamper:
Nice, thanks @karimnaaji!

PR #173, opened 2015-08-08
Add error and warning events

bcamper:
Adds initial error and warning events triggered by the `scene` object, which can be subscribed to by a Tangram client. These can be used to surface better feedback about scene config issues (such as when a data source is improperly configured, or a style's shaders fail to compile) -- the first expected use here is Tangram Play. Events receive a single object as an argument, with event-specific properties.

`scene` will emit these event types:

- `load`: scene successfully loaded, arguments:
  - `config`: the scene config object
- `error`: scene failed to load, arguments:
  - `message`: human-readable description
  - `type`': for now, only `yaml` for YAML validation errors (TODO: add more error types)
  - `error`: the Error object with further details (for YAML validation issues, can extract info on line # and error message)
  - `url`: the URL that the scene config was loaded from
- `warning`: scene loaded, but encountered an issue (each warning triggers a separate event), arguments:
  - `message`: human-readable description
  - `type`: area of functionality issue affects, such as `styles`, `sources`, etc. Additional type-specific fields are also included:
    - for `styles`:
      - `style`: the style that failed to compiled
      - `shader_errors`: array of shader errors, line # & shader block (if any) where error occurred can be extracted, along with GL error message and the source code of the line with the issue
    - for `sources`:
      - `source`: the data source that failed to instantiate

Note: the `load` event is technically a compatibility break (it was previously `loadScene` and had different arguments), however that API was not yet publicly documented so I think we can consider it a safe change.

A client can subscribe to these events as follows, given a `layer` object created by `Tangram.leafletLayer`. The `subscribe` method takes a single object, mapping event types to handler functions:

```
    layer.scene.subscribe({
        load: function (args) {
            console.log('scene loaded!');
        },
        error: function (args) {
            console.log(args.type, args.message);
        },
        warning: function (args) {
            console.log(args.type, args.message);
        }
    });
```

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -96,7 +96,8 @@ Enjoy!
         });
 
     layer.scene.subscribe({
-        loadScene: function (config) {
+        load: function (msg) {
+            var config = msg.config;
             // If no source was set in scene definition, set one based on the URL
             if (!config.sources || !config.sources['osm']) {
                 config.sources = config.sources || {};
--- a/demos/main.js
+++ b/demos/main.js
@@ -103,6 +103,12 @@ Enjoy!
                 config.sources = config.sources || {};
                 config.sources['osm'] = tile_sources[default_tile_source];
             }
+        },
+        error: function (msg) {
+            // debugger;
+        },
+        warning: function (msg) {
+            // debugger;
         }
     });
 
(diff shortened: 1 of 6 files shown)


bcamper:
@patriciogonzalezvivo ready for you soon...

patriciogonzalezvivo:
Awesome!

PR #176, opened 2015-08-14
Add a 'sprite_source' property to tie sprites to a feature property

bcamper:
`sprite_source` is a new draw group property for points that sets the `sprite` value to that of the specified feature property. For example:

`sprite_source: kind`

is syntactic sugar for

`sprite: function() { return feature.kind }`

The reasoning to add a specific property for it is:

- It turns out it's a fairly common way to specify feature/sprite mappings, with a pre-defined sprite sheet texture that lines up with known feature property values.
- It pairs well with the recently added `sprite_default` property. By setting a `sprite_source` and a `sprite_default`, it's easy to setup data-driven sprites with a clear fallback.
- The behavior and name is similar to the existing `text_source` field for the text style.

### DIFF
--- a/src/styles/points/points.js
+++ b/src/styles/points/points.js
@@ -63,11 +63,20 @@ Object.assign(Points, {
             return null;
         }
 
-        let sprite = style.sprite = rule_style.sprite;
+        // specify sprite explicitly
+        let sprite = rule_style.sprite;
         if (typeof sprite === 'function') {
             sprite = sprite(context);
         }
-        style.sprite_default = rule_style.sprite_default; // optional fallback if 'sprite' not found
+
+        // specify sprite by field
+        if (!sprite) {
+            sprite = rule_style.sprite_source;
+            sprite = (typeof sprite === 'string') && feature.properties[sprite];
+        }
+
+        // fallback if 'sprite' not found
+        style.sprite_default = rule_style.sprite_default;
 
         // if point has texture and sprites, require a valid sprite to draw
         if (this.texture && Texture.textures[this.texture] && Texture.textures[this.texture].sprites) {


bcamper:
@blair1618 what do you think of this?

@nvkelso this is intended to further simplify our POI style rules

nvkelso:
Nice!

blair1618:
I wonder if there's a more general way we can provide shortcuts access feature properties, so that we don't have to add a specific style property every time. Can we allow something like `$kind` or `$feature.kind` in general?

bcamper:
@blair1618 yes I have been thinking about that too. Here's a quick test of that, it was easy to do and is more generic:

https://github.com/tangrams/tangram/compare/property-substitution

On the other hand, we should think about normalizing the different ways we are already referring to feature properties. Current syntax examples:

- Draw group filter object, implicit feature property as key:
   - `filter: { kind: park }`
- Draw group filter function, explicit w/`feature` scope (equivalent to above example):
   - `filter: function() { return feature.kind === 'park' }`
- Text style `text_source` property, implicit feature property as value: 
   - `draw: { text: { text_source: kind } }`
- Draw group filter object, referencing non-feature property w/$ syntax:
   - `filter: { $zoom: { min: 10 } }`

How should we think about rectifying these to use the same syntax, without making it unwieldy? For instance, the filter syntax assumption that a feature property was being access, `kind` instead of something like `$feature.kind`, was a trade-off recognizing that it was a cleaner, simpler syntax that is used for 90% of filters. On the other hand, one could argue it is less clear.

Minimally, we could change the filter function syntax to be `$feature.` instead of just `feature.`, and/or accept both forms in filter objects or functions for backwards compatibility.

Ideas?

blair1618:
@bcamper yeah you're right, I had forgotten about all the ways we reference feature properties. In terms of making the syntax consistent and intuitive, it seems like we should have some principle that the syntax follows here. For instance, make `$property` in yaml *always* refer to `feature.property`. The examples then become:

 - `filter: { $kind: park }`
 - `filter: function() { return feature.kind === 'park' }` (JS should probably keep JS syntax)
 - `draw: { text: { text_source: $kind } }`
 - `filter: { zoom: { min: 10 } }` (`$` isn't used because zoom isn't in the feature data)

This is just one option that made sense to me. You also make a good point that if we change this, we can accept both syntaxes for a time (maybe with a warning).

bcamper:
@blair1618 ah right, but don't forget we already have "special" syntax, like

`filter: { $zoom: { min: 10 } }`

That does not refer to a feature property :/

blair1618:
Yeah check the last bullet point in my reply - `zoom` is kind of a special word already so this would just change they way that it's "special"

bcamper:
Oh sorry. Regarding JS, `$feature` isn't inherently any less "JS" than `feature.` is. Right now a variable called `feature` is put into JS function scope, it would be a one-character change to make it `$feature` :)

bcamper:
To elaborate, our initial thinking was that since feature properties were special, they would get the simplest syntax, meaning they could just be referred to directly, `kind`. This only really started breaking down with a property like `sprite`, which **already** excepts single-word string values that refer to named sprites. We didn't have this with the existing `text_source` property because it's overloaded to be either a feature property (when a string), or a function (returning the text to use for the label).

blair1618:
Ah, I suppose you could provide a variable `$feature` instead of `feature`, but that might suggest that one is referencing `feature.feature` :P

I definitely like how we prefer concise syntax for the most common styling cases, that's one of the best things about the stylesheet. Referring to feature properties with `$property` instead of `property` does involve that one extra character, it trades some typing for (what I think is) a pretty significant improvement in comprehensibility. This seems like a good trade-off to me because I don't think anyone has complained that the stylesheet is too verbose, but we have definitely seen cases of people struggling with understanding.

bcamper:
So there are also a couple branches relating to adding access to "scene"-level (global) properties. We have a `scene` block already for defining background color, with the intention to expand it to support things like language preference, and other user-defined properties that can be accessed in filters or functions. Zoom can also be considered one of these (it's a view-dependent scene property).

So we should take into account a syntax that provides access to those as well. These are not often-accessed, so I am thinking some kind of prefix like `scene.language`.

meetar:
`$` for feature properties and `scene.` for scene properties is growing on me, and if we could make it consistent within function scopes, so much the better – though I don't know that there's a good way to support both styles in a transition period, in the case of conflicts between style and function property names…

bcamper:
yeah been leaving this open because I'm hesitant to make any changes that require massive re-writing of styles with unclear overall benefit (`kind` -> `$kind`, etc. -- really that much better?)

bcamper:
Actually @blair1618, as mentioned the other day, still interested in possibly merging this as-is, because even though it's another explicitly defined property tying to features, it eliminates the (possibly wasteful) need for a JS property, and Eraser Map will need this, so it could be worth it.

blair1618:
Supporting both syntax forms in a single version does seem tricky. We can always make the change part of a new release, right?

@bcamper yeah regardless of the JS solution we end up with, the fastest JS function is the one we never have to run :) I had anticipated treating the special-keyword-feature-properties as a sort of built-in function, so we could access the value without running JS at all.

bcamper:
Right, I'm just looking for some reasonable solution in the near-term. I'm hesitant to introduce new syntax yet because the benefits just don't seem that great to me (my main dissatisfaction is with the different between bare-word feature property `kind` in filters, vs. scoped as `feature.kind` in JS functions... but this does not seem worth rewriting all styles to me).

This solution might be obviated by a more generic one in the future (if it turns out that a feature property as value is a common enough use case), but it conforms to the existing `text_source` behavior, so we could replace both at the same time.

bcamper:
This is likely to be resolved with a more generic approach, such as `sprite: feature.kind`.

PR #177, opened 2015-08-17
Label/Sprite offsets

karimnaaji:
Give the ability to offset sprite icons and labels,
- add a vertex attribute to the point style (offset in pixel applied directly in screen space)
- the following options are now availble from the stylesheet:
```yaml
draw:
    icons:
        # applies an offset of 13 pixels up along y-axis
        offset: [0px, 13px] 
[...]
draw:
    text:
        # applies an offset of 13 pixels down along y-axis
        offset: [0px, -13px]
```
- quick refactor on how features are stored in the text style


bcamper:
Can we flip the Y axis so positive offset is down? :) Since it's the default coordinate system for the browser/CSS, etc.

https://twitter.com/ID_AA_Carmack/status/633426212304646144

karimnaaji:
Oh yes, and it would also match with the native version! Let's get closer 1.0

bcamper:
Seems like it's still behaving as y pointing up... maybe it got double-flipped?

karimnaaji:
I just checked and it's working for me. Can you double check?

Issue #179, opened 2015-08-20
Text with light stroke has black junk around edge


nvkelso:
I'm seeing a lot of black junk around text that has both a fill and stroke, when the stroke is a light color like white or yellow.

<img width="64" alt="screen shot 2015-08-19 at 4 59 45 pm" src="https://cloud.githubusercontent.com/assets/853051/9372493/081cb834-4694-11e5-94e4-ed718c3cb35b.png">

<img width="113" alt="screen shot 2015-08-19 at 4 59 40 pm" src="https://cloud.githubusercontent.com/assets/853051/9372494/0835a920-4694-11e5-86bb-5a5cebb945b6.png">

bcamper:
This looks like an issue with pre-multiplied alpha handling between WebGL and Canvas. It looks like the text rasterized into the canvas uses pre-multiplied alpha, which means that instead of the halo simply alpha-fading to 0 at is edges, it is also "fading" out the halo RGB itself... which means for example with a white halo, you get a dirty gray at the edge as the RG components interpolate from 1 to 0. 

Our WebGL blending pipeline is currently setup to expect un-premultiplied alpha. We've found two ways to rectify this with our text rendering:

1. For text rendering, divide the color RGB by its alpha in the shader. Leave the rest of the blending pipeline alone. This is the simplest, but probably more costly approach, since it adds a per-fragment operation (for text only).

2. Add a new blending mode, alongside `overlay`, that is intended to be used with image sources with pre-multiplied alpha (e.g. `overlay-premultiplied`). This means:
   - Setting the `UNPACK_PREMULTIPLY_ALPHA_WEBGL` flag to `true` for text canvas textures.
   - Setting the blend equation to `gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA);` for the text (`overlay-premultiplied`) mode.

While Option 2 is probably more performant because it's an extra global GL state change (rather than a per-fragment operation), it's more expedient to fix this via Option 1 for now, then benchmark and revisit while we are looking at other text rendering adjustments.

bcamper:
Here is a before & after when using either option described (the results between the two fixes were visually indistinguishable).

Before - dirty text halo:
![tangram-1440099505518](https://cloud.githubusercontent.com/assets/16733/9393859/a7e7d34e-4752-11e5-85d1-ef332fc12427.png)

After - text halo removed:
[...]

PR #180, opened 2015-08-20
Un-multiply alpha on text styles

bcamper:
Fixes the "dirty" halos on text seen in #179 (see for more detail).

This was an issue with pre-multiplied alpha handling between WebGL and Canvas. It looks like the text rasterized into the canvas uses pre-multiplied alpha, which means that instead of the halo simply alpha-fading to 0 at is edges, it is also "fading" out the halo RGB itself. For example, with a yellow halo, you get a dirty gray at the edge as the RG components interpolate from 1 to 0 (instead of remaining constant while just the alpha interpolates).

Our WebGL blending pipeline is currently setup to expect un-premultiplied alpha. To fix this for now, we are simply dividing the color RGB by its alpha in the shader to "un-multiply" the alpha

In the future, we may consider adding a new blending mode, alongside overlay, that is intended to be used with image sources with pre-multiplied alpha (e.g. something like `overlay-premultiplied`). This route would include:

- Setting the UNPACK_PREMULTIPLY_ALPHA_WEBGL flag to true for text canvas textures.
- Setting the blend equation to gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA); for the text (overlay-premultiplied) mode.
- Adding a blend mode that would group these styles into a distinct render state.

Adding a new blending mode may be a more performant future solution, but the shader approach is a quick fix for now and can be adjusted without otherwise affecting behavior.

### DIFF
--- a/src/styles/points/points_fragment.glsl
+++ b/src/styles/points/points_fragment.glsl
@@ -45,6 +45,11 @@ void main (void) {
         }
     #endif
 
+    // Manually un-multiply alpha, for cases where texture has pre-multiplied alpha
+    #ifdef TANGRAM_UNMULTIPLY_ALPHA
+        color.rgb /= max(color.a, 0.001);
+    #endif
+
     #pragma tangram: color
     #pragma tangram: filter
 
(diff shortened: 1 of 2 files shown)


bcamper:
cc @nvkelso @blair1618 @karimnaaji

blair1618:
Makes sense to me. Until we can measure a performance benefit there's little sense adding another state change.

bcamper:
Closes #179

PR #181, opened 2015-09-01
Add $layer for filtering features according to their source layer

bcamper:
We support pulling multiple data source layers into a single style layer, like this:

`data: { source: osm, layer: [places, pois] }` 

However, there has been no way to distinguish *which* layer each feature came from, when filtering or dynamically computing properties for these features.

This PR adds a `$layer` property that contains this information. For the example above, `$layer` would be `pois` for features from the `pois` layer.

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -469,25 +469,43 @@ layers:
             filter: { kind: [hamlet, village, town, neighbourhood, suburb, quarter], $zoom: { max: 13 } }
             visible: false
 
-    landuse_labels:
-        data: { source: osm }
+    point_labels:
+        data: { source: osm, layer: [pois, landuse_labels] }
         filter:
             name: true
-            kind: [park, forest, cemetery, graveyard]
-            any:
-                # show labels for smaller landuse areas at higher zooms
-                - { $zoom: { min: 9 }, area: { min: 100000000 } }
-                - { $zoom: { min: 10 }, area: { min: 10000000 } }
-                - { $zoom: { min: 12 }, area: { min: 1000000 } }
-                - { $zoom: { min: 15 }, area: { min: 10000 } }
-                - { $zoom: { min: 18 } }
-        draw:
-            text:
-                priority: 4
-                font:
-                    typeface: Italic 10pt Lucida Grande
-                    fill: darkgreen
-                    stroke: { color: white, width: 3 }
+
+        landuse:
+            filter:
+                $layer: landuse_labels
+                kind: [park, forest, cemetery, graveyard]
+                any:
+                    # show labels for smaller landuse areas at higher zooms
+                    - { $zoom: { min: 9 }, area: { min: 100000000 } }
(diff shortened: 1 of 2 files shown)


bcamper:
cc @blair1618 @nvkelso 

Not sure about the naming of `$layer`, but I kept it "out" of the feature like this because:

- If it were a feature property, it would make more sense to call it `layer`, but this has too high a likelihood of collision and it seems better to keep the feature properties "pure" to minimize confusion.
- It's more performant to add it as a property to the context directly, once per layer, rather than having to iterate over all the features and insert it.

meetar:
It's definitely the most metal property name to date

blair1618:
:metal:

mjcunningham:
GWAR would work too.

blair1618:
This feels weird to me 1. because it's a new (and seemingly unrelated?) paradigm of using `$` and 2. because filters have always conceptually operated on features and their properties, which (as you mentioned) this doesn't seem to follow. 

Could we instead allow `layer` as a property in sublayers? (maybe we should start calling this concept something less overloaded, like `collection`?)

bcamper:
Agree that there is some ambiguity with `$`, but more generally it is a catch-all for properties that aren't "native" to the feature data. That doesn't mean they can't be related to the feature though, there is already a similar `$geometry` field for filtering on geometry type, and this seems similar: https://github.com/tangrams/tangram/blob/master/src/styles/style_parser.js#L131

Regarding "filters have always conceptually operated on features and their properties", `$zoom` is an important filter property that is not feature-specific.

I'm totally open to suggestions if there is a better syntax for including these types of fields that are "computed" from the feature, but I was trying to avoid adding yet more syntax prefixes.

Also wasn't sure about what you meant by this? "Could we instead allow layer as a property in sublayers?"

blair1618:
Oh you're totally right, `$zoom` isn't a feature property at all. 

What I meant in the other comment was something like:
```yaml
layers:
    point_labels:
         data: { source: osm, layer: [pois, landuse_labels] }
         landuse:
             layer: landuse_labels
             filter:
                 kind: [park, forest, cemetery, graveyard]
                 any:
                     ...
```
but that's not really any better. Thinking about it along with `$zoom`, `$layer` seems less weird now.

bcamper:
We're kind of in "less weird" territory here :)  Personally I'm satisfied if we just keep a distinction between a) properties that "came with" the feature directly from the data source (no manipulation of the original data) and then b) additional properties deemed useful for filtering/functions, prefixed with $.

meetar:
I'm fine with the prefix – but from a developer perspective, I don't think the distinction is particularly meaningful… I suppose it is less likely to suggest that every feature has a .layers property accessible in styling functions?

bcamper:
I was mostly concerned with not mucking with the feature data, to avoid possible collisions, e.g. if we over-wrote a property someone had in their data called `layer` (which seems quite plausible).

PR #182, opened 2015-09-04
Add cardinal direction placements for point labels

bcamper:
Add a `placement` property for point labels, to place them in a cardinal direction anchored to their point position (default positioning is centered on anchor point).

Example:

```
draw:
   text:
      placement: E
      offset: [12px, 0px]
      font:
         ...
```

Valid `placement` values are: `N`, `S`, `W`, `E`. Note these are actually interpreted as relative up/down/left/right placements relative to the anchor, not actually moving the label in the specified direction on the map. If this is undesirable we can also use an alternative like: `up`, `down`, `left`, `right`.

Also still thinking about support for:

- Dynamically determining `placement` from a feature property and/or JS function
- Supporting multiple placement options, to try in order of priority, e.g. `placement: [E, S]`

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -560,8 +560,10 @@ layers:
                 $zoom: { min: 18 }
             draw:
                 text:
-                    offset: [12px, 0px]
-                    placement: E
+                    anchor: right
+                    # align: left
+                    offset: [13px, 0]
+                    move_into_tile: false
                     priority: 6
                     font:
                         family: Helvetica
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -427,7 +427,7 @@ layers:
         filter: { name: true, not: { kind: [peak, viewpoint, bicycle_rental, car_sharing] }, $zoom: { min: 15 } }
         draw:
             icons:
-                offset: [0px, -13px]
+                # offset: [0px, -13px]
                 size: [[13, 12px], [15, 18px]]
                 interactive: true
 
@@ -560,7 +560,8 @@ layers:
                 $zoom: { min: 18 }
             draw:
                 text:
-                    offset: [0px, 13px]
+                    offset: [12px, 0px]
+                    placement: E
                     priority: 6
                     font:
                         family: Helvetica
--- a/src/styles/text/feature_label.js
+++ b/src/styles/text/feature_label.js
@@ -1,6 +1,7 @@
 import Utils from '../../utils/utils';
 import Geo from '../../geo';
 import {StyleParser} from '../style_parser';
+import LabelPoint from './label_point';
 
 export default class FeatureLabel {
 
@@ -75,6 +76,16 @@ export default class FeatureLabel {
             text_wrap = default_font_style.text_wrap;
         }
         style.text_wrap = text_wrap;
+
+        // default alignment to match anchor
+        if (!rule.align && rule.anchor && rule.anchor !== 'center') {
+            if (LabelPoint.isLeftAnchor(rule.anchor)) {
+                rule.align = 'right';
+            }
+            else if (LabelPoint.isRightAnchor(rule.anchor)) {
+                rule.align = 'left';
+            }
+        }
 
         style.align = rule.align || default_font_style.align;
 
--- a/src/styles/text/feature_label.js
+++ b/src/styles/text/feature_label.js
@@ -49,7 +49,7 @@ export default class FeatureLabel {
         let size_kind = ft_size.replace(/([0-9]*\.)?[0-9]+/g, '');
 
         // TODO: improve pt/em conversion
-        style.px_logical_size = Utils.toPixelSize(ft_size.replace(/([a-z]|%)/g, ''), size_kind);
+        style.px_logical_size = parseInt(Utils.toPixelSize(ft_size.replace(/([a-z]|%)/g, ''), size_kind));
         style.px_size = style.px_logical_size * Utils.device_pixel_ratio;
         style.stroke_width *= Utils.device_pixel_ratio;
         style.size = size.replace(size_regex, style.px_size + "px");
--- a/src/styles/text/label.js
+++ b/src/styles/text/label.js
@@ -51,23 +51,21 @@ export default class Label {
         return true;
     }
 
-    // whether the label should be discarded
-    // 1. try to keep the label in tile if the label (to avoid collision over tile for now)
-    // 2. if 1. -> keep a minimal distance between the label
-    // 3. if 2. -> perfom occlusion
+    // Whether the label should be discarded
+    // Depends on whether label must fit in the tile bounds, and if so, can it be moved to fit there
     discard (aabbs) {
         let discard = false;
 
-        // perform specific styling rule, should we keep the label in tile bounds?
-        if (this.options.keep_in_tile) {
+        // Should the label be culled if it can't fit inside the tile bounds?
+        if (this.options.cull_from_tile) {
             let in_tile = this.inTileBounds();
 
-            if (!in_tile && this.options.move_in_tile) {
-                // can we move?
-                discard = this.moveInTile();
+            // If it doesn't fit, should we try to move it into the tile bounds?
+            if (!in_tile && this.options.move_into_tile) {
+                // Were we able to fit it in the tile?
+                discard = this.moveIntoTile();
             } else if (!in_tile) {
-                // we didn't want to move at all,
-                // just discard since we're out of tile bounds
+                // discard since we're out of tile bounds
                 return true;
             }
         }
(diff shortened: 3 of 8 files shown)


bcamper:
@nvkelso thoughts on style property names and values? e.g. `placement` and `NSWE` vs. others.

@karimnaaji @blair1618 we'll want to add to ES once we settle on syntax.

nvkelso:
I'm fine with `placement`. Is the more CSS-like property name `position`? 

I'm confused by the syntax, though. If I set the `placement: E`, I would expect the resulting text part to be offset several points right (in screen space reference frame) of the icon without also needing to specify an additional `offset: [12px, 0px]`. In your example, is the offset option to get my desired behavior, or required?

Labeling on the corner is super useful: (adding) NW, NE, SW, SE.

Fuller set:

cardinal, offset, up-down
C, center, none
N, top, up
S, bottom, down
W, left, left
E, right, right
NW, top-left, up-left
NE, top-right, up-right
SW, bottom-left, down-left
SE, bottom-right, down-right

I prefer `cardinal` over `offset` (but don't like `up-down` at all).

blair1618:
From the diff, it looks like `placement` specifies the edge of the label that will coincide with the point position. I didn't understand the connection there until I looked at the code - maybe something like `anchor` would be more clear? Corner placements are a good idea too!

bcamper:
I borrowed `placement` from CartoCSS (which has `text-placements`). `position` means something different internally but I suppose it doesn't cause an issue to use in the scene file. Anyway I don't have a strong preference. I also thought about `anchor` as @blair1618 suggested, maybe that's both a less overloaded and more specific term.

Below are some images to illustrate the current behavior. The reason I added the offset in the example is that I was drawing an icon and text separately, and since Tangram currently has no concept of these being linked (which it can in the future), it would not know how much to offset the label to the right to clear the icon. Currently the anchors are up against the edge of the text. I've added a red dot at the label point (around which the text is centered by default) to show this. Offset is then applied on top of that. We could apply some buffering factor if there was a logical choice? I was trying to leave it flexible but I want the default behavior to be good.

(Text baseline is not centered here, that's a separate issue to fix.)

What's your expectation compared to this?

Center (default):
screen shot 2015-09-04 at 10 28 37 pm

West:
screen shot 2015-09-04 at 10 28 57 pm

East:
screen shot 2015-09-04 at 10 29 09 pm

North:
screen shot 2015-09-04 at 10 29 19 pm

South:
screen shot 2015-09-04 at 10 29 32 pm

I like adding corners, will do that!

sensescape:
Yes, I like cardinal too. And definitely corners too! Can we control spacing between the anchor and position too? Or can we set it to be at a comfortable space, not too tight or too far from the anchor point.

bcamper:
Maybe what we need is some kind of `padding` setting, to buffer the label relative to its anchor point. An example could be:

`padding: [10px, 5px]`

This would be interpreted as:

- Centered anchor: no additional offset for label
- East anchor: label is offset 10 pixels to the right
- West anchor: label is offset 10 pixels to the left (e.g. -10px)
- South anchor: label is offset 10 pixels  down
- North anchor: label is offset 10 pixels up (e.g. -10 pixels)
- NW, SW, etc., extrapolated from behavior above

sensescape:
The amount of space at the top and the bottom should be different because of the ascenders and the descenders.

bcamper:
Typography is the worst.

bcamper:
So would it be best to specify separate offsets for every direction? We could do that too.

sensescape:
: ) Yes, being able to adjust the padding would be good, better to have full control, especially when we start using different fonts.

nvkelso:
@bcamper In your example for `padding: [10px, 5px]`, what does the `5px` do?

bcamper:
@nvkelso duh I mis-labeled it :/

Should be:

- South anchor: label is offset **5 pixels** down
- North anchor: label is offset **5 pixels** up

I see now that additional options are needed though.

nvkelso:
I dig the examples, they match what I expect. 

For padding it's really the bottom anchor (less than everyone else) and top anchor (sometimes more) that need the custom paddings – the other anchors can more or less all share the same padding. Is it better to specify this using a sublayer with a filter or with property/properties declaration?

nvkelso:
CSS has levels of details for properties like `padding` and `padding-top`, could the same be useful here?

bcamper:
I tend to feel that CSS has gone overboard with *both* the shorthand and individual properties (it's a lot to keep track of both as an author and implementer). 

I am more of a fan of accepting different value types for different scenarios though, like:

- `offset: 5px`: apply 5px offset in the anchor direction (e.g. none if centered, 5px left if W, etc.)
- `offset: [10px, 5px]`: apply 10px offset for horizontal (W or E) anchors and 5px for vertical ones (N or S)
- `offset: [10px, 8px, 5px, 15px]`: CSS-like pattern, right-top-left-bottom values, e.g. 10px right for E, 8px up for N, 5px left for W, 15px down for S

(Not sure if those are exactly right, but for example)

meetar:
I tend to agree – 

So in this scheme, why would you specify offsets for multiple directions if the placement is only ever in one direction?

bcamper:
@meetar that is a good question -- anticipating the need for these (mentioned at bottom of original PR description):

- Dynamically determining placement from a feature property and/or JS function
- Supporting multiple placement options, to try in order of priority, e.g. placement: [E, S]

nvkelso:
+1 for different value types for different scenarios. Since the corners generally need less padding the full set might look something like: `offset: [4px, 2px, 4px, 4px, 3px, 3px, 3px, 3px]` where `offset: [N, S, E, W, NE, SE, NW, SW]` or whatever order convention we adopt?

bcamper:
@nvkelso would like to revisit this and get it merged. To be resolved:

- Do we prefer `placement` or `anchor`?
- We seem to prefer cardinal labels, e.g. `N`, `SE`, etc. For situations where the view rotates/tilts, does `N` actually mean "up", and we are just using them because it's convention? (I'm OK with that but we need to clarify.) Or does a `N` anchor flip upside down when the view is rotated 180 (I wouldn't expect so).
- For either, I am thinking in the future (not in initial PR) we can support multiple acceptable anchors in order of preference, like `anchor: [E, S]`, and have an offset value corresponding order, e.g. `offset: [5px, 10px]` could apply a `5px` offset when `E` anchor is used, and a `10px` offset when `S` anchor is used.

nvkelso:
We use 'label_placement' in tiles to say the point where the label icon should be placed and from where the label text should be anchored. I wonder if  'orientation' is better to describe that we're talking about the orientation from that point (rather than confuse with similar 'position').

All the cardinals are with respect to screen space (even thru screen rotation), this is to reduce flicker. Allowing reference to be stable with respect to geography is a future feature enhancement, and should be paired with other label enhancements.

Agree on future multiples.

_n

bcamper:
`orientation` seems fine to me too. I can update the branch with that.

bcamper:
Hmm... forgot text labels also have an `angle`, it's mostly internal, but it's still potentially confusing with `orientation`. Not a deal breaker I guess.

bcamper:
Thinking about this more... having a `N` direction that actually means "up", and can be pointing South, does put me off a bit. The Mapbox GL style spec uses "anchor" and up/down/etc. terminology, which I think may be more appropriate:

> text-anchor 
Optional enum.  One of center, left, right, top, bottom, top-left, top-right, bottom-left, bottom-right. Defaults to center. Requires text-field.
Part of the text placed closest to the anchor.

https://www.mapbox.com/mapbox-gl-style-spec/#symbol

blair1618:
`orientation` is confusing to me, my first guess would be that `orientation` rotates a label. `anchor` seems the least ambiguous.

nvkelso:
There's no right way, let's just be consistent and get this merged.

blair1618:
Right on. For the record, I vote for `anchor` as the key and I'm indifferent between `N, S,W, E` and `left, right, top, bottom`

@karimnaaji Do labels in tangram-es have functionality for something like this yet?

karimnaaji:
No nothing yet, mainly waiting on this to be settled down.

bcamper:
Updating to use `anchor`, and `up`, `down`, `left`, `right` (and corners `top-left`, etc.) terminology.

karimnaaji:
:+1:

bcamper:
OK, so now we have:

- `anchor`: `center` (default), `top`, `bottom`, `left`, `right`, `top-left`, `top-right`, `bottom-left`, `bottom-right`
- `align` is defaulted as follows if it is not explicitly set:
  - `anchor: center|top|bottom` => default `align: center`
  - `anchor: left|top-left|bottom-left` => default `align: right`
  - `anchor: right|top-right|bottom-right` => default `align: left`
- JS-only option to move the label into the tile is now exposed as style rule `move_into_tile` (defaults true to match existing behavior, but should be set to false if using anchor/align functionality for text + icon combinations, otherwise, text can get out of sync with expected position)

nvkelso:
This is going to be awesome! :)

PR #183, opened 2015-09-04
Dynamic label priority

bcamper:
Adds the ability for the label `priority` parameter to be dynamic, either:

- Tied to a feature property, by setting `priority` to a string value with the property name, e.g.
  - `priority: label_priority`
- Computed from a JS function, e.g.
  - `priority: function() { return Math.floor(feature.area / 1000); }`

Requested by @nvkelso for upcoming tile data for label priorities.

### DIFF
--- a/src/styles/text/text.js
+++ b/src/styles/text/text.js
@@ -437,7 +437,19 @@ Object.assign(TextStyle, {
             }
 
             // label priority (lower is higher)
-            let priority = (rule.priority !== undefined) ? parseFloat(rule.priority) : -1 >>> 0;
+            let priority = rule.priority;
+            if (priority !== undefined) {
+                // if priority is a number, use it as-is, otherwise, check type
+                if (typeof priority === 'string') {
+                    priority = feature.properties[priority]; // get priority from feature property
+                }
+                else if (typeof priority === 'function') {
+                    priority = priority(context);
+                }
+            }
+            else {
+                priority = -1 >>> 0; // default to max priority value if none set
+            }
 
             // label offset in pixel (applied in screen space)
             let offset = rule.offset || [0, 0];
--- a/src/styles/text/text.js
+++ b/src/styles/text/text.js
@@ -440,10 +440,11 @@ Object.assign(TextStyle, {
             let priority = rule.priority;
             if (priority !== undefined) {
                 // if priority is a number, use it as-is, otherwise, check type
-                if (typeof priority === 'string') {
-                    priority = feature.properties[priority]; // get priority from feature property
-                }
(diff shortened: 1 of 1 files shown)


bcamper:
@blair1618 this follows behavior for label text, and similarly proposed for sprites. You OK with it?

bcamper:
Notes from discussion with @blair1618: also considering requiring a `feature.` prefix, e.g.:

`priority: feature.priority`

This would also apply to #176.

bcamper:
So with `feature.`, we'd be looking at these permutations:

- `sprite: cafe`: show the sprite named `cafe`
- `sprite: feature.kind`: tie the sprite to the value of a feature property
- `priority: 5`: assign an explicit priority
- `priority: feature.priority`: tie the priority to the value of a feature property
- `text_source: feature.kind`: tie the label text to the value of a feature property
- `text_source: function() { return feature.height > 100 ? 'high' : 'low' }`: dynamically compute the label text based on feature properties

`sprite` continues to be the only ambiguous case *without* adding the `feature.` syntax, because it's the only style property (so far) where a string value can have two meanings -- a string can be the name of a sprite defined in the scene file, or it can refer to a feature property. #176 solves this by adding a separate `sprite_source` property; the alternate proposal is to solve it by making the feature property explicit with `feature.`. The latter has grown on me because it does make the relationship more clear. 

Either way we are still left with the inconsistency of filter properties implicitly referring to features (`kind: park`), but I continue to believe that more concise syntax is preferable for readability (granted the inconsistency can be confusing).

bcamper:
I have disabled the original syntax in this PR, pending further syntax discussion. I may try the `priority: feature.priority` syntax (also for sprites) in another branch.

In the meantime, this PR enables JS functions for label priority, which serves our needs in the short-term.

nvkelso:
To recap, is this the supported syntax?

* priority: 5: assign an explicit priority
* priority: feature.priority: tie the priority to the value of a feature property

bcamper:
We'll refrain from adding any new syntax as part of this PR. Instead, we'll just use a JS function property for now to tie priority to a feature property:

`priority: function() { return feature.priority }`

As before, `priority: 5` for an explicit priority.

Proposed future syntax, which I will revisit in a PR applying to sprites and label text all together, is `priority: feature.priority` (or `sprite: feature.kind` for sprites).

meetar:
@bcamper could you verify the updated docs here? https://github.com/tangrams/tangram-docs/blob/1ed9fcf485bbf2b34a2402fc5b7cca1cffd52ea1/pages/draw.md#priority

bcamper:
Make some minor tweaks to this and offsets.

PR #184, opened 2015-09-08
Track shader block 'scope'

bcamper:
We recently added shader error events with detailed metadata, allowing Tangram Play to extract information about which styles shader blocks failed to compile, including specific line numbers mapped from GL shader compile errors. However, after using this in the editor, it became clear that the need was more complex than the original implementation. For example, Tangram itself injects `global` shader blocks internally, for common shared functions, throwing off the line numbers for user-defined shader code. Additionally, compile errors in a block that was mixed in from a 'parent' style wouldn't be found when the 'child' style failed to compile, preventing the error from being properly syntax highlighted.

This PR adds lower-level tracking of where each shader block came from. For example, when `styleA` mixes in `styleB`, we now track which blocks came from the 'parent `styleA`, and which were added as part of `styleB`'s own definition. This adds more detailed error metadata, which should allow Tangram Play or other clients to properly trace the error back to its original source.

### DIFF
--- a/src/gl/shader_program.js
+++ b/src/gl/shader_program.js
@@ -26,6 +26,7 @@ export default class ShaderProgram {
 
         // key/values for blocks that can be injected into shaders at compile-time
         this.blocks = Object.assign({}, options.blocks||{});
+        this.block_scopes = Object.assign({}, options.block_scopes||{});
 
         // list of extensions to activate
         this.extensions = options.extensions || [];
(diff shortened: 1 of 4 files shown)


bcamper:
Trying these new changes in Tangram Play has significantly improved the situation, but have also revealed further dependencies that make it tricky to properly trace errors. For example, a `color` block in `styleB` can have a syntax error, which can *also* cause a cascading error in `filter` block inherited from `styleA`. The second "error" gets highlighted in `styleA`'s `filter` block, even though there is nothing apparently wrong in that block and nothing that can be done there to fix the error.

So to properly highlight errors, we would need to "sandbox" each shader block in its own function, within well-defined in/out/inout parameters (which isn't a bad idea anyway, could add a small overhead to shader time due to function call, but my understanding is that the shader compilers are pretty aggressive about inlining).

Merging this now since it is an improvement, and will look at shader block sandboxing later.

PR #186, opened 2015-09-10
Meetar/line extrude

meetar:


bcamper:
Closing because wayyyy out of date, but this could be re-created on the current builder code base. Would need some discussion about how to do it without impacting the non-extruded case though.

PR #187, opened 2015-09-11
Meetar/labeldev

meetar:
adding comments


bcamper:
Hey @karimnaaji can you review these comments to make sure they are accurate and up to date with recent changes?

src/styles/text/text.js
```diff
@@ -240,11 +240,36 @@ Object.assign(TextStyle, {
     },
 
     createLabels (tile, texts) {
-        let labels_priorities = {};
-
+        let labels_priorities = {};  // this will store all labels, sorted into objects by priority
```
karimnaaji: Labels of the tile only.

src/styles/text/text.js
```diff
@@ -305,18 +335,33 @@ Object.assign(TextStyle, {
         // Process labels by priority
         let priorities = Object.keys(labels).sort((a, b) => a - b);
         for (let priority of priorities) {
-            if (!labels[priority]) {
+            if (!labels[priority]) { // no labels at this priority, skip to next
                 continue;
             }
 
+            // i is a label object at the current priority level, example:
```
karimnaaji: Not sure where this is coming from, but `i` is not the label object of the current priority, nor is `labels[priority][i]`. But the object description is ok.

src/styles/text/text.js
```diff
@@ -305,18 +335,33 @@ Object.assign(TextStyle, {
...
+            //  label: LabelLine,
+            //  style: "100 24px Helvetica/rgb(102,102,102)/rgb(255,255,255)/8"
+            // }
+
             for (let i = 0; i < labels[priority].length; i++) {
                 let { style, feature, label } = labels[priority][i];
 
+                // test the label for intersections with other labels in the tile
                 if (!label.discard(this.aabbs[tile])) {
+                    // if it didn't collide (or won a collision)
```
karimnaaji: There is not really a concept of winning a collision, not colliding would be sufficient.

karimnaaji:
Most of those are ok, :+1: @meetar!

src/styles/text/text.js
meetar: I think this was changed since I wrote this comment -- I'll update it

PR #197, opened 2015-10-07
Access to normalized model & world coordinates (ES compatibility)

bcamper:
Since Tangram JS and ES have different internal tile representations (for some reasons justifiable, others not), we are adding a `g_model_position` global variable, which normalizes the vertex position into a [0, 1] range representing the bounds of the tile. Note un-clipped geometry can still validly be outside of this range. This makes some style features, which previously relied on the `a_position` attribute, possible in a platform-independent way.


bcamper:
@blair1618 now that I've added this, I wonder if a function *is* the better approach, if it might have more chance of being optimized out or at least not triggering unnecessary work. The problem with this approach is the math (simple, but annoying especially for things like converting z space that will seldom be used) will *always* get executed at the start of the shader. Unless the compiler is smart enough to optimize out the variable, but I am unclear?

bcamper:
@blair1618 alternative:

```
vec3 modelPosition() {
    return a_position.xyz * 32767. / TANGRAM_TILE_SCALE; // position in model space in [0, 1] range
}
```

bcamper:
Still need to add equivalent to `worldPosition()` in ES (see https://github.com/tangrams/tangram-es/pull/333). But going to merge in the meantime since we need `modelPosition()` more immediately.

bcamper:
Updated to use some small GLSL utils for unpacking normalized integer attributes, removing some magic numbers from shaders.

PR #198, opened 2015-10-08
Only mix each ancestor style once

bcamper:
This prevents shader blocks that have already been mixed in from an ancestor style from being mixed a second time.

This supports a 'diamond'-like mixin hierarchy where:

```
      A
     / \
    /   \
   B     C
    \   /
     \ /
      D
```

In this case (or more complex ones like it where a style is mixed into multiple ancestors), style `D` should end up with only one copy of style `A`.

Note: one thing I am not entirely pleased with in this solution is that it only applies explicitly to shader blocks. Because mixing the same style multiple times does not cause breakage for any properties other than shader blocks, this may be an acceptable approach for now. It could raise some potential issues with order of mixin application, since shader blocks will be mixed in the order of the first parent found, while later mix occurrences of a style could affect other properties through overwrite or flag behavior.

### DIFF
--- a/src/styles/style_manager.js
+++ b/src/styles/style_manager.js
@@ -258,6 +258,7 @@ StyleManager.mix = function (style, styles) {
     }
 
     // Merge shader blocks, keeping track of which style each block originated from
+    let mixed = {}; // all scopes mixed so far
     for (let source of merge) {
         if (!source.blocks) {
             continue;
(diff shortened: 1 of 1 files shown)


bcamper:
@blair1618 @tallytalwar this is something you might want to take a look at. Basically, there is only special handling for preventing shader blocks from being mixed in multiple times. For all other style properties, we let them be mixed multiple times, which does not cause compile errors (and with current implementation would be much harder to prevent :). I'm a little unsure about what that means for order of application, e.g. shader blocks for style A are added the first time they are encountered, so will execute in that order relative to other shader blocks, but the if style A is later mixed again, its other properties may overwrite those from earlier styles.

Then I realized we probably have a similar situation with the ES implementation. Also, I don't think we can rely on filtering shader blocks by duplicates in ES. It is valid in many cases to have the same shader block source repeated (as long as they originate in different styles).

PR #200, opened 2015-10-12
Don't extrude polygon edges along tile boundaries by default

bcamper:
When constructing polylines, we exclude line segments that coincide with tile boundaries by default (they can be enabled if desired with the `tile_edges: true` style parameter, usually for debugging).

We have not been applying the same logic to polygon extrusion, most likely an oversight. We haven't noticed because Mapzen vector tile building polygons are unclipped, and these are the most common extrusion case. A recent issue reported by a user revealed this problem when using non-tiled GeoJSON.

This extends the same behavior (and `tile_edges` flag) to the `polygons` style.

### DIFF
--- a/src/styles/builders.js
+++ b/src/styles/builders.js
@@ -84,7 +84,13 @@ Builders.buildExtrudedPolygons = function (
     vertex_data, vertex_template,
     normal_index,
     normal_normalize,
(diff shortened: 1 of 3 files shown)


bcamper:
cc @blair1618 we will want to apply the same logic in ES.

PR #202, opened 2015-10-16
Alpha background

bcamper:
Adds alpha to the Tangram GL canvas, which allows for Leaflet layers (or other HTML content) to be placed *beneath* the map, in addition to on top of it. The GL canvas is now inserted in the correct location in the Leaflet DOM (previously it was pulled out of the DOM to make positioning simpler, and because without alpha it often got covered by other layers). 

This also necessitates a few changes to client interactivity handling, since we want listeners and pop-ups attached to the map container, not the Tangram scene container. (see https://github.com/tangrams/tangram/compare/alpha-background#diff-a2925ce738063617bef38726407de2d0R484, https://github.com/tangrams/tangram/compare/alpha-background#diff-a2925ce738063617bef38726407de2d0R511); cc @meetar and @nvkelso on this point for adjusting other demos (but will put this Tangram change in a 0.4 release for better compatibility transition).

Color values can now include an alpha component (including for "opaque" geometry that isn't otherwise doing GL blending with other Tangram features). For example:

```
scene:
    background:
        color: [0, 0, 0, .5] # allow background to show through with 50% opacity
```

or 

```
scene:
    background:
        color: transparent
```

or

```
        draw:
            polygons:
                color: [.25, 0, 0, .25] # draw red with 25% opacity in background
```

Here's an example of Tangram buildings and roads rendered over Stamen's terrain raster tiles:

screen shot 2015-10-15 at 10 16 47 pm

### DIFF
--- a/demos/css/main.css
+++ b/demos/css/main.css
@@ -15,7 +15,7 @@ body {
     width: 200px;
     margin-left: -100px;
     position: absolute;
-    z-index: 10;
+    z-index: 100001;
     text-align: center;
     font-family: Helvetica, sans-serif;
     font-size: 12px;
--- a/demos/main.js
+++ b/demos/main.js
@@ -189,7 +189,7 @@ Enjoy!
     // Update URL hash on move
     map.attributionControl.setPrefix('');
     map.setView(map_start_location.slice(0, 2), map_start_location[2]);
-    map.on('moveend', updateURL);
+    map.on('move', updateURL);
 
     // Take a screenshot and save file
     function screenshot() {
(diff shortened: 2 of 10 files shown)


bcamper:
Note: I haven't been able to test this on iOS yet, and given that we've seen different premultiplied canvas alpha behavior on iOS vs. desktop Safari -- causing the "dirty" text halos -- I need to make sure to do that before this is merged.

blair1618:
Is `background: color: transparent` equivalent to `background: color: [0, 0, 0, 0]`? If so, why add exceptional behavior? How much do we really like CSS? ;)

bcamper:
It was just an example that works for free because 'transparent' is a CSS
'color' picked up by the parser automatically :) I expect the same is true
in the C++ version.

blair1618:
Ooh I see! Very nice.

PR #204, opened 2015-10-21
Data sources can append custom URL params

bcamper:
This adds the ability for data source definitions to add custom URL parameters to the query string. This has little benefit currently, since you can just write out the whole URL explicitly. But it will allow us to more easily customize parameters like API keys:

- By having the Leaflet plugin set the API key param in the scene file when it's passed through as an option
- Once we incorporate file inclusion, this will allow users to easily import a Mapzen basemap and then set their own API key.

Example:
Instead of the full URL, `http://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson?api_key=vector-tiles-HqUVidw`, you can write:

```
sources:
   osm:
      type: TopoJSON
      url: http://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
      params:
         api_key: vector-tiles-HqUVidw
```

The future inclusion example mentioned above would then look like:

```
include: path/to/mapzen/style.yaml

sources:
    mapzen:
        params:
            api_key: vector-tiles-HqUVidw
```

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -159,7 +159,7 @@ sources:
     mapzen:
         type: TopoJSON
         url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
-        params:
+        url_params:
             api_key: vector-tiles-HqUVidw
         max_zoom: 16
         # # Data filtering demo with 'scripts' and 'transform' properties:
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -158,43 +158,43 @@ styles:
 sources:
     mapzen:
         type: TopoJSON
-        url:  http://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson?api_key=vector-tiles-HqUVidw
+        url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
+        params:
+            api_key: vector-tiles-HqUVidw
         max_zoom: 16
-    #     # Data filtering demo with 'scripts' and 'transform' properties:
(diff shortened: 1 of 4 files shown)


bcamper:
@patriciogonzalezvivo we will also be able to more effectively handle API keys in Play this way (rather than relying on regex)

patriciogonzalezvivo:
This seams great

bcamper:
@blair1618 I'd like to revive this in advance of scene "include" functionality -- it was partially created with the ability to set API keys in mind.

Another good example is the search demo I created the other day:

```
sources:
    search:
        type: GeoJSON
        url: https://search.mapzen.com/v1/search?api_key=search-pBUTKfQ
        params:
            text: los angeles
```

Will do a search for Los Angeles.

blair1618:
Ah, this looks convenient!

URL parameters don't need to be ordered, do they?

It would be nice if the syntax associated the 'params' with the URL, since that's the only part it applies to, though the simplicity of this syntax is also nice.

bcamper:
URL query string parameters can be in any order, yes. Hm, rather than nesting it under `url`, I'd probably prefer to clarify the naming if needed, like `url_params`. I just thought `params` was more convenient and is likely to be commonly understood from a web perspective. Plus we don't really have non-URL based data sources right now, though I imagine we may eventually!

blair1618:
Yeah I don't think nesting items under `url` would be useful. `source.url_params` would be clearer to me than `source.params` (again, not a web person here). 

The dynamic data sources used in tangram-es are technically non-url based, but also aren't declared in the scene's `sources`.

bcamper:
`url_params` works, I'll update.

PR #205, opened 2015-10-25
Multi-line labels

bcamper:
Re-implements multi-line labels (previous implementation had issues and was disabled). Labels with word wrap / line breaks are now drawn directly as multi-line labels in the Canvas texture (e.g. the word wrapping is implemented as part of the Canvas text drawing, vs. as part of the Tangram label/geometry building/rendering process).

The maximum number of characters per line before a line break occurs is configurable via the `max_line_width` parameter (should this be `max_line_characters`?). 

```
draw:
   text:
      max_line_width: 10
      ...
```

Word wrap is **enabled** by default for point labels (labels for point or polygon geometries) with a default `max_line_width` of **15 characters** (is this the best default?), and is **disabled** by default for line labels. Word wrap can be explicitly disabled by setting `max_line_width: false`.

Additionally, line breaks occurring *in the label text itself*, e.g. `\n` characters, will always trigger line breaks / multi-line label rendering. For example, the label text `One\nTwo\nThree` will render as three separate lines, regardless of the `max_line_width` setting. If these breaks are undesirable, they can be filtered out of the input with a `text_source: function() { ... }`.

It is also possible to control the text **alignment** when there are multiple lines of text. Possible alignment values are `center` (default), `left` justified, and `right` justified. Alignment has no effect if the label fits on a single line, as the alignment only affects the the position of the lines *relative to each other*; it does not affect the position of the entire label itself (use `offset` for that, and/or the upcoming `anchor` functionality, see #182).

```
draw:
   text:
      align: right
      ...
```
Examples:

Defaults of `max_line_width: 15` and `align: center`:
center

With `align: right`:
right

With `align: left`:
left

With `max_line_width: 0` (every word renders on its own line, because every word exceeds the max allowed length):
width 0

With `max_line_width: false`, e.g. word wrapping disabled:
word wrap off

### DIFF
--- a/src/styles/text/feature_label.js
+++ b/src/styles/text/feature_label.js
@@ -1,4 +1,5 @@
 import Utils from '../../utils/utils';
+import Geo from '../../geo';
 import {StyleParser} from '../style_parser';
 
 export default class FeatureLabel {
@@ -7,7 +8,7 @@ export default class FeatureLabel {
         this.text = text;
         this.feature = feature;
         this.tile_key = tile.key;
-        this.style = this.constructFontStyle(rule, context, default_font_style);
+        this.style = this.constructFontStyle(feature, rule, context, default_font_style);
         this.style_key = this.constructStyleKey(this.style);
     }
 
@@ -16,7 +17,7 @@ export default class FeatureLabel {
         return Utils.hashString(str);
     }
 
-    constructFontStyle (rule, context, default_font_style) {
+    constructFontStyle (feature, rule, context, default_font_style) {
         let style = {};
         rule.font = rule.font || default_font_style;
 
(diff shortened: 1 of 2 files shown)


bcamper:
cc @nvkelso @blair1618 for syntax/behavior feedback, and @karimnaaji for thinking about if/how this will match up well with ES.

nvkelso:
> property name for `max_line_width` and text wrap

It's confusing that we're talking about **text wrap**, but `wrap` isn't in the property name. Mapnik includes it in their property name directly with [wrap-width](https://github.com/mapnik/mapnik/wiki/TextSymbolizer). 

What about calling this: `text_wrap` like we call `text_source`? I think it makes more sense as top level instead of shoved into the `font` section.


> default max_line_width of 15 characters

Yes, 15 chars makes sense to me. Can be adjusted to taste by designer later.


> Possible alignment values are center (default)

Center alignment default makes sense to me.


> Word wrap is enabled by default for point labels (labels for point or polygon geometries) and disabled by default for line labels. 

Points and polygon text wrap (and not line geom label wrap) matches my expectation. 

**For documentation:** how would we enable text wrap on line labels?

nvkelso:
Nice work, btw!

karimnaaji:
This is great! It would definitely match on ES. I started to implement it and we'll have the same result.

blair1618:
Looks good to me; I also like the idea of "wrap" in the name to indicate the limiting behavior. I would guess that `true` for the value would turn wrapping on with the default limit?

bcamper:
Agree, `text_wrap` makes more sense, will update. (It already is outside of `font` actually, at the `text` level. `font` properties only affect the *appearance* of the text itself, not the *layout* of the text.)

Also makes sense to have `text_wrap: true` turn on with the default (currently only `false` turns it off, `true` wouldn't turn it on, but agree that is confusing). Similar to `lighting: true` sets default lighting, while `lighting: fragment|vertex` can set specific modes.

For lines: you can turn it on simply by setting a wrapping value (which will be `false`/`null` by default). Agree we should reflect that in documentation.

nvkelso:
Would `text_wrap: true` be the default value? Seems useful enough that I
vote this is the default.

bcamper:
Yep, that's right.

bcamper:
Updated re: discussion, merging (can adjust further as needed). Thanks!

PR #206, opened 2015-10-26
Uniform run-time values forward from ancestor to descendant

bcamper:
This fixes an issue where uniform values set at run-time would not inherit to "descendant" styles they were mixed into. This was because uniform values were cloned (along with the rest of the style properties) at mixing time. The "parent" and "child" uniforms ended up as separate JS properties, even if they were intended to have the same value (e.g. the child style did not override the uniform with a new default value). Usually this is not noticeable, but in situations where uniforms are dynamically altered at run-time (such as demos with GUIs), the values would not propagate and the parent and child would diverge. @patriciogonzalezvivo noticed this when working on small "modules" of common style functionality.

This PR fixes the behavior by creating JS getters and setters for uniforms. The getter returns an explicitly defined value for that style if one has been set, otherwise it will forward the request to the appropriate ancestor style that the uniform was mixed from. This necessitates keeping track of uniform "scope" (e.g. the source style for each uniform), in the same way that we now do for shader blocks (in the latter case to support error handling). The setter sets the uniform value on the style, which means it will be returned as a explicit value (overriding the original one) in subsequent getter calls. This effectively mimics the behavior of the JS prototype chain; actual prototype forwarding wasn't feasible in this case because styles can mixin uniforms from multiple "ancestors".

### DIFF
--- a/src/styles/style_manager.js
+++ b/src/styles/style_manager.js
@@ -197,6 +197,7 @@ StyleManager.mix = function (style, styles) {
     if (style.mixed) {
         return style;
     }
+    style.mixed = {};
 
     // Mixin sources, in order
     let sources = [];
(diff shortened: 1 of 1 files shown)


bcamper:
@blair1618 I don't think this will be a pressing issue on ES anytime soon, but it is an interesting thing to be thinking about. I am assuming we would have the same issue there with uniforms being "cloned" at mixing time.

PR #208, opened 2015-10-30
Tangram can be loaded outside of a bundled JS

louh:
### The bottom line

Now, you can bundle up all of your app's front-end dependencies via Browserify except for Tangram. Tangram will wait until `Tangram.leafletLayer()` is called before looking for Leaflet.

### Usage

In HTML:

```html
  <script src='./path/to/tangram.min.js'></script>
  <script src='/js/bundle.js'></script>
```

In your `bundle.js`:

```js
var L = require('leaflet')
// Your other required dependencies

// [Init Leaflet map...]

// Add Tangram scene layer
// Provide the required Leaflet variable to Tangram
var layer = Tangram.leafletLayer({
  leaflet: L,
  scene: 'https://raw.githubusercontent.com/tangrams/refill/gh-pages/refill.yaml',
}).addTo(map);
```

### Changelog

- An entry point for `dist/tangram.min.js` is provided in `package.json` in case the repository is downloaded as an NPM module.
- When building the module, the output is piped through `derequire` so that other requires won't break by looking for submodules of Tangram.
- `Tangram.leafletLayer()` is wrapped so that `leaflet_layer.js` does not look for Leaflet until this function is called
- `Tangram.LeafletLayer` is never defined so it's removed from `module.js`

### Notes re: browserify & browserify-shim

Tangram cannot currently be bundled with other JS because this prevents Tangram from discovering its own URL for web workers.

The [browserify-shim module](https://github.com/thlorenz/browserify-shim) is not required to use Tangram. However, with it you can do this:

In `package.json`:

```json
  "browserify": {
    "transform": [ "browserify-shim" ]
  },
  "browserify-shim": {
    "tangram": "global:Tangram"
  }
```

In your `bundle.js`:

```js
var L = require('leaflet')
var Tangram = require('tangram') // via browserify-shim
```

This allows the developer to maintain consistency about dependencies although Tangram is still a global object. This is also more future proof because if the issues within Tangram are resolved this allows a developer to throw away the shim and the code is already prepared to accept Tangram as a normal `require`. [More details here](https://github.com/thlorenz/browserify-shim#a-expose-global-variables-via-global).

### Next steps

Webpages that load all scripts in individual tags should behave normally.

Check to see how it works in Webpack (for web map, cc @hanbyul-here)

~~Argh, tests are broken.~~ fixed

### DIFF
--- a/makefile
+++ b/makefile
@@ -12,7 +12,7 @@ all: \
 debug: dist/tangram.debug.js
 
 dist/tangram.debug.js: .npm src/gl/shader_sources.js $(shell ./build_deps.sh)
-	node build.js --debug=true --require './src/module.js' --runtime | derequire > dist/tangram.debug.js
+	node build.js --debug=true --require './src/module.js' --runtime | $(DEREQUIRE) > dist/tangram.debug.js
 
 dist/tangram.min.js: dist/tangram.debug.js
 	$(UGLIFY) dist/tangram.debug.js -c warnings=false -m -o dist/tangram.min.js
--- a/makefile
+++ b/makefile
@@ -2,6 +2,7 @@
 UGLIFY = ./node_modules/.bin/uglifyjs
 KARMA = ./node_modules/karma/bin/karma
 JSHINT = ./node_modules/.bin/jshint
+DEREQUIRE = ./node_modules/.bin/derequire
 
 # Build debug and minified libraries
 all: \
@@ -11,7 +12,7 @@ all: \
 debug: dist/tangram.debug.js
 
 dist/tangram.debug.js: .npm src/gl/shader_sources.js $(shell ./build_deps.sh)
-	node build.js --debug=true --require './src/module.js' --runtime > dist/tangram.debug.js
+	node build.js --debug=true --require './src/module.js' --runtime | derequire > dist/tangram.debug.js
 
 dist/tangram.min.js: dist/tangram.debug.js
 	$(UGLIFY) dist/tangram.debug.js -c warnings=false -m -o dist/tangram.min.js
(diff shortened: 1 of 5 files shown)


bcamper:
Nice, will take a look! IIRC, the webpack problem was even weirder -- I think even when Tangram was loaded as a script tag, webpack was re-writing the document with its own content, causing Tangram to be unable to find it later...

bcamper:
I'm not sure what happens here if you call `leafletLayer` multiple times (which is valid). Can we add a check so it only initializes the first time it's called?

louh:
I guess it would try to reconstruct everything each time. I could do this:

```js
function extendLeaflet(options) {

    // If LeafletLayer is already defined when this is called (because it was defined previously), just return that immediately
    if (typeof LeafletLayer !== 'undefined') {
        return new LeafletLayer(options);
    }

    // Leaflet layer functionality is only defined in main thread
    if (Utils.isMainThread) {
       [...]
```

Is there an example where someone is actually doing multiple Tangram.leafetLayer() calls?

One caveat so far is that if someone does this:

```js
// One instance of Leaflet
var layer1 = Tangram.leafletLayer({
  leaflet: L1,
  scene: 'scene1.yaml',
}).addTo(map1); 

// I want to use a second instance of Leaflet on the same page, for whatever reason
var layer2 = Tangram.leafletLayer({
  leaflet: L2,
  scene: 'scene2.yaml',
}).addTo(map2); 
```

And there are multiple Leaflet instances on the same page only the first one gets defined as L.

This is probably not going to happen often, though?

bcamper:
Thanks, this looks good - agreed on not needing to support multiple Leaflets imported under different names in the same page (that seems pretty arcane).

PR #213, opened 2015-11-06
Support anchors for points/sprites

bcamper:
Generalizes the logic from point labels so that it also applies to points/sprites. The same `anchor` values apply (`left`, `bottom-right`, etc.). Unlike with labels, the `align` keyword is not needed.


bcamper:
@karimnaaji you also added this to ES as part of text anchors, yes? Any divergence here?

karimnaaji:
No divergence :+1: ! In ES it's here, just need an access from the stylesheet.

PR #214, opened 2015-11-09
Font property interpolation by zoom

bcamper:
Adds zoom-based interpolation ("stops") for the following font properties:

- Font size (specified in px, pt, or em)
- Fill color (partially worked previously, fixes bugs with using string colors)
- Stroke width (specified in px)
- Stroke color

```
draw:
  text:
    font:
      family: Helvetica
      fill: [[14, white], [18, gray]]
      size: [[14, 12px], [16, 16px], [20, 24px]]
      stroke:
        color: [[16, white], [18, red], [20, blue]] 
        width: [[14, 3px], [20, 8px]]
 ```

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -270,102 +270,89 @@ layers:
                 style: flat_lines
                 color: white
                 width: 12
-                # width: [[13, 0px], [14, 1px], [15, 8], [18, 8], [19, 0]]
                 order: function () { return feature.sort_key + 5; }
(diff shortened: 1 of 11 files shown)


bcamper:
@blair1618 @karimnaaji we will need to add this for ES too before we can use, but @nvkelso said this could reduce the style rules significantly, so I prioritized it.

bcamper:
also cc @sensescape :)

PR #217, opened 2015-11-13
Apply multipolygon ring decoding to MVT sources

bcamper:
The next rev of the MVT format (https://github.com/mapbox/vector-tile-spec/) clarifies how multipolygons should be encoded: as a single "flattened" series of rings, where individual polygons can be extracted by examining the winding order of each to determine inner/outer rings; a new polygon starts each time an outer ring is found.

We are already doing this for polys generated by geojson-vt. Now we are also applying it to MVT tile sources. This is backwards compatible with our current tiles, and forwards compatible with https://github.com/mapzen/mapbox-vector-tile/pull/21.

### DIFF
--- a/src/sources/geojson.js
+++ b/src/sources/geojson.js
@@ -1,4 +1,5 @@
 import DataSource, {NetworkSource, NetworkTileSource} from './data_source';
+import {MVTSource} from './mvt';
 import Geo from '../geo';
 
 // For tiling GeoJSON client-side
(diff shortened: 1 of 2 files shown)


bcamper:
Verified on current 0.4 tiles on prod, and on dev tiles w/new encoding.

tallytalwar:
WIll create a card for this for ES side.

PR #222, opened 2015-12-02
Label repetition

bcamper:
This introduces draw parameters for controlling how often labels are allowed to repeat. Two style parameters are added to control this behavior, `repeat_distance` and `repeat_group`.

output_lktevy

output_j36n16

- The `repeat_distance` parameter (for `text`-based `draw` groups) that specifies **minimum distance in pixels** between labels.
  - Currently, this is simply calculated as a center-to-center distance between two label bounding boxes, though it could be expanded to account for label size. For performance, possibly a simpler approximation and not a true distance between bounding boxes; the behavior here needs to adhere reasonably close to the desired style, but does not need to be pixel-perfect.
  - Note that because the Tangram JS architecture processes each tile in isolation (and manages labels and their accompanying texture atlases per tile), we currently can only enforce minimum label distance **within a tile**. This means that you may still see labels near each other when they are drawn close to the edges of adjacent tiles. This can be addressed in a future overhaul of the collision system for now, the overall behavior is still much improved.
  - The *default value* for `repeat_distance` is the logical pixel size of a tile (256 pixels), meaning that by default, a given label will only be placed once per tile.
  - Note that if `repeat_distance: 0px`, no restrictions are made on placement, and labels are allowed to repeat freely.
  - Examples:
```
draw:
   text:
      repeat_distance: 100px # label can repeat every 100 pixels
      ...
```
```
draw:
   text:
      repeat_distance: 0px # labels can repeat anywhere
      ...
```
- By default, labels are **grouped by layer** before repetition logic is applied. This means that a unique text string is only checked for repetition against other labels that matched the same draw rules; another label with the same text may be placed within the `repeat_distance` if it is the result of a different set of draw rules. This allows two features with the same name but different classifications (such as a highway and nearby minor road) to both be labelled.
  - However, the grouping can also be controlled by setting the `repeat_group` property. This property defaults to an internal representation of all the feature's matching rules, but if set to another (arbitrary) string value, allows for more sophisticated control over which features are allowed to repeat.
  - Examples:
```
roads:
   major_roads:
      filter: { kind: major_road }
      draw:
         text:            
            ...
   minor_roads:
      filter: { kind: minor_road }
      draw:
         text:
            # labels from the major_roads layer can repeat near these, because they
            # are in different repeat groups by default            
            ... 
```
```
roads:
   draw:
      text:
         # labels in the sub-layers below won't repeat near each other, 
         # because they have been placed in the same repeat group
         repeat_group: roads-fewer-labels
   major_roads:
      filter: { kind: major_road }
      draw:
         text:            
            ...
   minor_roads:
      filter: { kind: minor_road }
      draw:
         text:            
            ...
```

These different behaviors can be observed here:

Repeat checking turned off with `repeat_distance: 0px` (e.g. same as current behavior):
tangram-mon nov 30 2015 15-20-43 gmt-0500 est

With default repeat distance and grouping. Note that many road-related labels appear to repeat near each other because of classification differences (links, etc.):
tangram-mon nov 30 2015 15-20-56 gmt-0500 est

With all roads in the same repeat group (`repeat_group: roads-fewer-labels`), additional repetition is reduced:
tangram-mon nov 30 2015 15-21-04 gmt-0500 est

### DIFF
--- a/src/styles/rule.js
+++ b/src/styles/rule.js
@@ -8,6 +8,7 @@ export const whiteList = ['filter', 'draw', 'visible', 'data', 'properties'];
 export let ruleCache = {};
 
 function cacheKey (rules) {
+    rules = rules.sort((a, b) => a - b);
     var k = rules[0].id;
     for (var i=1; i < rules.length; i++) {
         k += '/' + rules[i].id;
--- a/src/styles/style_parser.js
+++ b/src/styles/style_parser.js
@@ -144,7 +144,7 @@ StyleParser.getFeatureParseContext = function (feature, tile) {
 
 // Build a style param cache object
 // `value` is raw value, cache methods will add other properties as needed
-// `transform` is optional transform function to run on values
+// `transform` is optional transform function to run on values (except function values)
 StyleParser.cacheObject = function (obj, transform = null) {
     if (obj == null) {
         return;
@@ -155,11 +155,11 @@ StyleParser.cacheObject = function (obj, transform = null) {
     }
 
     if (typeof transform === 'function') {
-        if (Array.isArray(obj) && Array.isArray(obj[0])) {
+        if (Array.isArray(obj) && Array.isArray(obj[0])) { // zoom stops
             obj = obj.map(v => [v[0], transform(v[1])]);
         }
-        else {
-            obj = transform(obj);
+        else if (typeof obj !== 'function') { // don't transform functions
+            obj = transform(obj); // single value
         }
     }
 
--- a/src/styles/text/label.js
+++ b/src/styles/text/label.js
@@ -4,6 +4,8 @@ import boxIntersect from 'box-intersect'; // https://github.com/mikolalysenko/bo
 import Utils from '../../utils/utils';
 import OBB from '../../utils/obb';
 
+import log from 'loglevel';
+
 export default class Label {
 
     constructor (text, size, options) {
@@ -23,20 +25,22 @@ export default class Label {
         // Broadphase
         if (aabbs.length > 0) {
             boxIntersect([this.aabb], aabbs, (i, j) => {
+                log.trace(`${this.text} broad phase collide`, this, this.aabb, aabbs[j]);
+
                 // Narrow phase
                 if (OBB.intersect(this.aabb.obb, aabbs[j].obb)) {
+                    log.trace(`${this.text} narrow phase collide`, this, this.aabb.obb, aabbs[j].obb);
                     intersect = true;
                     return true;
                 }
             });
         }
-
-        // No collision on aabb
-        if (!intersect) {
-            // it's clean, add it to the list of bboxes
-            aabbs.push(this.aabb);
-        }
         return intersect;
+    }
+
+    // Add this label's bounding box to the provided set
+    add (aabbs) {
+        aabbs.push(this.aabb);
     }
 
     // checks whether the label is within the tile boundaries
--- a/src/styles/text/layout_settings.js
+++ b/src/styles/text/layout_settings.js
@@ -1,3 +1,5 @@
+import Geo from '../../geo';
+
 var LayoutSettings;
 
 export default LayoutSettings = {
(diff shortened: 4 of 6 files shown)


bcamper:
Call for comments from @blair1618 @nvkelso @karimnaaji @meetar @sensescape:
- Thoughts on `repeat` and `repeat_group` parameter names? Immediately after writing this and seeing the examples, I see that `repeat: false` is really semantically confusing since it actually means all repeats are allowed :)  We could be more explicit and call this `repeat_dist`, `min_repeat_dist`, etc. (I think I was going for brevity originally).
- I'm unsure if the default `repeat_group` behavior (grouping at the most specific style level) is the best. This was my interpretation of what @nvkelso wanted, but we might tweak it, or if not, we can still set more specific behavior in the style rules.

meetar:
Agree re: `repeat: false` – `repeat_distance` or `repeat_limit` etc would be better (I tend to prefer un-abbreviated property names, for clarity). Also not totally clear on `repeat_group` – Does "This property defaults to an internal representation of all the feature's matching rules" just mean it's an inheritable property? Or that it's some kind of hash such that any feature with the same equivalent rules would have the same repeat_group by default, even if it was in a separate branch of the styling rules?

bcamper:
@meetar It's a bit of a mouthful ("all the feature's matching rules") because of parallel matching rules (besides inheritance, a feature can match other rules "alongside"). The only deterministic way I could see to take that into account was to make the group be the unique combination of rules (which is used for other internals like caching the merged properties already anyway). In many cases I think features are matching a single layer, but we do also use the technique for some common styling needs.

meetar:
So practically, does that mean any features which share any matching
styling rules will also share a repeat_group by default?

bcamper:
Only if they share all the same styling rules.

meetar:
:+1:

blair1618:
I also agree that `repeat: false` should change. Maybe `min_spacing` / `minimum_spacing` would more clearly reflect the behavior. `min_spacing: false` would make more sense, at least. Then a corresponding `spacing_group` parameter if we wanted to keep those lexically similar. 

Aside from naming, do we need to accept `false` as a value? Doesn't `0` cover that without introducing extra reasoning?

tallytalwar:
Should we also allow to exclude a specific sublayer from a `repeat_group`?

Also is there any significance of `repeat_group` name and uber layer's name same? `roads` in the above example?

karimnaaji:
I am confused that `repeat: false` means that the label can repeat anywhere, at a first read I thought that this means that the repeat was not allowed. What about:
```yaml
repeat: false # meaning you don't want labels to repeat
repeat_dist: 100px roads # apply to group roads
# or even
repeat: false 100px roads # repeat: true|false min_dist group
```
It's more clear while reading the stylesheet.

tallytalwar:
I like @karimnaaji suggestion of `repeat: false 100px roads # repeat: true|false min_dist group`.

bcamper:
I agree `repeat` was a poor name choice :)  I think `repeat_dist` would work, @sensescape also suggested `min_repeat_dist`.

To @karimnaaji's suggestion, I'm not crazy about having to parse out multiple values from a single string (like `repeat_dist: 100px roads`). I think it's more consistent with the rest of the scene to have separate properties and be explicit about it. I'm also hoping that the default grouping logic (which we may want to change from currently implemented) works well enough that it is rare to change this property (or maybe it is not needed in the stylesheet at all).

bcamper:
(@tallytalwar, there was no special significant of `repeat_group: roads` in the example above, it is an arbitrary string; I'm reluctant to do anything more advanced there, such as matching against layer names, for what I think is an advanced feature.)

bcamper:
@blair1618 in JS, `repeat: false` and `repeat: 0` doesn't really necessitate much extra, but I can see how it might in C++ :)  We can drop this, I just wasn't sure if it was clear to people that they allow labels to repeat freely by putting a "distance" of zero in.

karimnaaji:
@bcamper right, it's more consistent with the rest of the scene, and let the possibility to transform the repeat distance in stops. Then I also like @blair1618 mentioned, setting `min_repeat_distance: false ` makes you understand at first read that you don't want it to repeat, otherwise giving a value makes the minimal distance effective.

blair1618:
After a little thought, `min_spacing` doesn't quite work because there's no indication that it only applies to items with the same text. I like `min_repeat_distance` though!

@bcamper It's not that parsing `false` is especially difficult, it's more that it adds a piece of knowledge that you need to read the stylesheet and doesn't add any functionality. And what would `true` mean? It would be a little silly to accept one and not the other.

nvkelso:
This PR gives me warm and fuzzy feelings :+1: 

I prefer the two property names to start with the same word so it's more self-documenting that they're related (and not abbreviated):

* `repeat_distance`
* `repeat_group`

Are these properties zoom "stoppable" now? The `repeat_distance` property would be super useful to set like this now, I can see some uses for the `repeat_group` property, but more of an advanced feature.

I was similarly confused as @tallytalwar about the `repeat_group: roads` example. Glad to hear it's arbitrary and supported. Maybe make the example `repeat_group: road-labels-less-spacing` (spacing is the Mapnik term, might improve SEO) or something else besides one of the example layer names? Excluding a specific layer from a `repeat_group` seems like a 2.0 advanced feature (there are use cases, but none pressing today, and are blocked on a rewrite of how labels works as a unified symbolizer).

tallytalwar:
@nvkelso, with respect to excluding sublayers ... I think it will work with current implementation.

Example: (have minor roads exlcuded from the major road and all other roads group)

```
roads:
   draw:
      text:
         repeat_group: all_roads
   major_roads:
      filter: { kind: major_road }
      draw:
         # will pick up repeat_group value from parent layer
         text:
             ...
   minor_roads:
      filter: { kind: minor_road }
      draw:
         text:         
             # overwrites parent's repeat_group value
             repeat_group: not_in_all_roads
             ...
```

I do not know how its implemented (have not looked the code changes), but if `repeat_group` is acting like any other draw group parameter, then during feature matching all but `minor_roads` will have `not_in_all_roads` value for `repeat_group` parameter. All other road types will have `all_roads` value for `repeat_group` name. Hence `minor_roads` will be excluded.

@bcamper, correct me if I am wrong :D.

bcamper:
@tallytalwar you are correct, that works. I think this is pretty flexible using normal inheritance rules.

Thanks everyone, will incorporate changes and summarize here.

bcamper:
Changes:

- Updated syntax to `repeat_distance` (and keeping `repeat_group`)
- Remove `repeat_distance: false`, this is accomplished with `repeat_distance: 0px`
- Updated custom repeat group example to `repeat_group: roads-fewer-labels` for clarity
- Updated PR description to reflect syntax changes.

I think this covers it?

PR #223, opened 2015-12-07
Add `blend_order` to control render order of styles with non-opaque blending

bcamper:
`blend_order` is an optional parameter used when defining a render style, which controls the order in which styles *with non-opaque blending (`add`, `multiply`, `inlay`, `overlay`)* are rendered. `blend_order` is an integer greater than or equal to zero.

Previously, styles were bucketed by blending type, and rendered in the following order: `opaque`, `add`, `multiply`, `inlay`, `overlay`.

With this PR, the order is now:

- All `opaque` styles render first
- All non-`opaque` styles **without** a defined `blend_order` render next, sorted by `add`, `multiply`, `inlay`, `overlay`
- All non-opaque styles **with** a defined `blend_order` render last, sorted by `blend_order` (ascending), sub-sorted by `add`, `multiply`, `inlay`, `overlay`

Each group above also now has a final sub-sort by style `name`, to provide a consistent render order (to aid in debugging and/or reduce unforeseen visual differences caused by render order, though these should not occur if the intended behavior is correct).

The practical effect is that styles can be placed on top of others by assigning a `blend_order`. The following adds a `marker` style, which is rendered on top of the `icons` style, which is (implicitly) on top of the default `text` style (all use `overlay`):

```
styles:
   marker:
      base: points
      blend: overlay
      blend_order: 1   # marker goes on top of icons and text
      ...

   icons:
      base: points
      blend: overlay
      blend_order: 0   # icons go on top of text, underneath marker
      ...
```

### DIFF
--- a/src/scene.js
+++ b/src/scene.js
@@ -6,6 +6,7 @@ import subscribeMixin from './utils/subscribe';
 import Context from './gl/context';
 import Texture from './gl/texture';
 import VertexArrayObject from './gl/vao';
+import {Style} from './styles/style';
 import {StyleManager} from './styles/style_manager';
 import {StyleParser} from './styles/style_parser';
 import SceneLoader from './scene_loader';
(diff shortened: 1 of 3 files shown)


bcamper:
@blair1618 I think the most relevant item here is the function used for sorting the styles: https://github.com/tangrams/tangram/compare/blend-order?expand=1#diff-e90eeb4f0c5a50cd03b669245e7e0724R424

PR #225, opened 2015-12-07
Resolve ambiguous draw rules by lexical sort

bcamper:
Sometimes a feature matches multiple layers, at the same depth, that set the same property (or properties). This isn't desirable because it's ambiguous, and should be avoided when authoring layer rules (it *is* often useful to write multiple layers that match the same feature, they should just be used to control different properties, for example one set of rules to control color and another parallel set to control order). However, it's also a little too easy to do accidentally :)

Previously, the resolution of which layer would "win" was undefined, leading to divergent behavior between Tangram JS and ES. This PR resolves the conflict by **lexically sorting the layers by name**. As with deeper layers overwriting the properties of shallower layers, the **last layer sorted by name** will overwrite any earlier layers. This means that if layers named `layerA` and `layerB` both try to set a property, the value defined in `layerB` will be used:

```
layers:
   layerA:
      color: red

   layerB:
      color: blue   # feature will be blue because `layerB` sorts after `layerA`
```  

This is not intended as the final solution for this issue. But before we introduce additional syntax for reducing collisions, it provides a deterministic fallback behavior that can be used by both platforms. Note, a suggested and preferred resolution would be to use the rule that occurs *later* in the file, e.g. last definition wins. However, since plain JS objects don't guarantee sorting of keys by insertion order, it is not currently possible to implement this on the JS side; this could possibly be enabled by future additions to the js-yaml parser to either use an ES6 Map (which does support enumeration of keys by insertion order), or the addition of positional metadata for YAML mappings.

### DIFF
--- a/src/styles/rule.js
+++ b/src/styles/rule.js
@@ -15,17 +15,18 @@ function cacheKey (rules) {
     return k;
 }
 
-export function mergeTrees(matchingTrees, key, context) {
-    var draw = {},
-        draws,
-        treeDepth = 0,
-        x, t;
+// Merge matching layer rule trees into a final draw group
+export function mergeTrees(matchingTrees, group) {
+    let draws, treeDepth = 0;
 
-    // Visible by default
-    draw.visible = true;
+    let draw = {
+        visible: true // visible by default
+    };
+
+    let layers = new Set(); // layers that contributed to the draw group
 
     // Find deepest tree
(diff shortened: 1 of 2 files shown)


bcamper:
@blair1618 there are some other incidental changes and a test here, but it basically boils down to this sort: https://github.com/tangrams/tangram/pull/225/files#diff-deabab718e02af3782a643dcdbbde01aR48

PR #226, opened 2015-12-09
Set texture

patriciogonzalezvivo:
Dynamically change textures using an ```url```/```image```/```canvas```/```data``` by providing one of this on the ```scene.config.textures```.  Ex:

```
scene.config.textures.orbits.url = "http://..../image.png";
scene.updateConfig({ rebuild: true });

// or

var canvas = new document.createElement("canvas");
...
scene.config.textures.orbits.element = canvas;
scene.updateConfig({ rebuild: true });

// or

var image = new document.createElement("image");
...
scene.config.textures.orbits.element = image;
scene.updateConfig({ rebuild: true });

// or

var raw_data = { data: [], width: 500, height: 500 }
...
scene.config.textures.orbits.data = raw_data;
scene.updateConfig({ rebuild: true });
```

### DIFF
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -73,12 +73,18 @@ export default class Texture {
     }
 
     load(options = {}) {
+        this.loading = null;
+
         if (typeof options.url === 'string') {
-            return this.setUrl(options.url, options);
+            this.setUrl(options.url, options);
         } else if (options.element) {
-            return this.setElement(options.element, options);
+            this.setElement(options.element, options);
         } else if (options.data && options.width && options.height) {
-            return this.setData(options.width, options.height, options.data, options);
+            this.setData(options.width, options.height, options.data, options);
+        }
+
+        if (this.loading) {
+            return this.loading.then((tex) => { this.calculateSprites(); return tex; });
         }
     }
 
@@ -100,7 +106,6 @@ export default class Texture {
             image.onload = () => {
                 try {
                     this.setElement(image, options);
-                    this.calculateSprites();
                 }
                 catch (e) {
                     log.warn(`Texture '${this.name}': failed to load url: '${url}'`, e, options);
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -262,6 +262,10 @@ export default class Texture {
 
 // Static/class methods and state
 
+Texture.create = function constructor(gl, name, options = {}) {
+    return new Texture(gl, name, options);
+};
+
 // Destroy all texture instances for a given GL context
 Texture.destroy = function (gl) {
     var textures = Object.keys(Texture.textures);
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -198,7 +198,7 @@ export default class Texture {
             return;
         }
 
-        options.filtering = options.filtering || this.filtering || 'linear';
+        options.filtering = options.filtering || 'linear';
 
         var gl = this.gl;
         this.bind();
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -18,7 +18,7 @@ export default class Texture {
 
         this.name = name;
         this.source = null;
-        this.source_type = 'none';
+        this.source_type = null;
         this.loading = null;    // a Promise object to track the loading state of this texture
         this.filtering = options.filtering;
         this.sprites = options.sprites;
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -138,6 +138,15 @@ export default class Texture {
         this.data = null;
     }
 
+    setImage(image, options) {
+        this.image = image;
+        this.update(options);
+        this.setTextureFiltering(options);
+
+        this.canvas = null; // mutually exclusive with other types
+        this.data = null;
+    }
+
     // Uploads current image or buffer to the GPU (can be used to update animated textures on the fly)
     update(options = {}) {
         if (!this.valid) {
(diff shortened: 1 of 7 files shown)


src/gl/texture.js
```diff
@@ -16,19 +16,20 @@ export default class Texture {
             this.valid = true;
         }
         this.bind();
-        this.image = null;      // an Image object/element that is the source for this texture
-        this.canvas = null;     // a Canvas object/element that is the source for this texture
+
+        this.name = name;
+        this.source = null;
```
bcamper: Let's use `null` instead of strings like "none", that's more JS-like.

src/scene.js
```diff
@@ -1000,6 +1000,28 @@ export default class Scene {
         }
     }
 
+    // setTexture (style_name, uniform_name, element) {
```
bcamper: revert these changes

src/scene_loader.js
```diff
@@ -86,7 +86,10 @@ export default SceneLoader = {
 
                     // Texture by URL, expand relative to scene file
                     if (typeof tex === 'string' && !config.textures[tex]) {
-                        style.texture = Utils.addBaseURL(tex, path);
+                        let texture_name = '__' + style_name;
+                        config.textures[texture_name] = { url: tex };
+                        style.texture = texture_name; // point style to location of texture
+                        // style.texture = Utils.addBaseURL(tex, path);
```
bcamper: we can take out these commented lines

src/scene_loader.js
```diff
@@ -115,11 +122,14 @@ export default SceneLoader = {
...
-                                style.material[prop].texture = Utils.addBaseURL(tex, path);
+                                let texture_name = '__' + style_name + '_material_' + prop;
+                                config.textures[texture_name] = { url: tex };
+                                style.material[prop].texture = texture_name; // point style to location of texture
+                                // style.material[prop].texture = Utils.addBaseURL(tex, path);
                             }
                             // Texture by object, move it to the global scene texture set and give it a default name
                             else if (typeof tex === 'object') {
-                                let texture_name = '__' + style_name;
+                                let texture_name = '__' + style_name + '_material_' + prop;
```
bcamper: Note: this fixes an existing bug in master.

src/scene_loader.js
```diff
@@ -99,10 +102,14 @@ export default SceneLoader = {
                 // If style has texture uniforms, expand texture URLs relative to scene file
                 if (style.shaders && style.shaders.uniforms) {
                     for (let {type, value, key, uniforms} of GLSL.parseUniforms(style.shaders.uniforms)) {
-                        if (type === 'sampler2D' && !config.textures[value]) {
-                            uniforms[key] = Utils.addBaseURL(value, path);
+                        if (type === 'sampler2D' && typeof value === 'string' && !config.textures[value]) {
+                            let texture_name = '__' + style_name + '_uniform_' + key;
```
bcamper: Note: this fixes an existing bug in master.

src/gl/texture.js
```diff
@@ -299,6 +316,7 @@ Texture.createFromObject = function (gl, textures) {
 Texture.changed = function (name, config) {
     if (Texture.textures[name]) { // cached texture
         // compare definitions
+        // TODO: fix for canvas/image/data
```
bcamper: `canvas`/`image` elements won't be stringified, which means that differences in them won't be detected. We probably need to go ahead and write (or use an existing) deep comparison method.

src/gl/texture.js
patriciogonzalezvivo: :+1:

src/gl/texture.js
patriciogonzalezvivo: Is not easier to add boolean flag for changes?

bcamper:
Can you please add a description of changes, and usage examples?

bcamper:
Please also remember to not commit the `dist/` files in the future, we should remove them from this PR, but even adding them in a single commit adds them to history forever :/

patriciogonzalezvivo:
Should I delete and create again this branch?

bcamper:
I don't think so, we can rebase before merging.

patriciogonzalezvivo:
sorry, I don't remember when I include them

bcamper:
We don't commit them except when a new release is cut'm

bcamper:
Now that we have this working, can we combine `canvas` and `image` into a single `element` parameter for textures? We can detect based on type, and other aspects are similar (both can be provided directly as WebGL textures).

patriciogonzalezvivo:
Done! Also, that lint error on text.js is annoying. Any idea how to solve it?

bcamper:
Re: lint, just remove `let texture = ` and let the `new Texture` call be on its own. I don't really like that syntax wise, but it's a result of how the rest of the system works. Maybe I can replace with a static instantiation method that would be clearer (`Texture.create`).

patriciogonzalezvivo:
:+1:

patriciogonzalezvivo:
should I merge master on it? or you want to do it before merging this brunch into master?

bcamper:
No please wait, I will review more later, thanks.

bcamper:
@patriciogonzalezvivo with https://github.com/tangrams/tangram/commit/4172dc5534f89740ee3ba09d106536c54c3ee59d, the `element` can be set to a CSS selector string. 

So you can setup HTML like:

`<canvas class='awesome-texture'>`

And then reference it as a texture with:

```
textures:
   awesome-texture:
      element: .awesome-texture
```

Or you could reference an element by id, but the use of `#` in CSS selector means you need to be careful to quote the YAML so it's not interpreted as a comment :) For example:

`<canvas id='texture-id'>`

```
textures:
   awesome-texture:
      element: '#texture-id'  # without quote, the # will be interpreted as a YAML comment!
```

patriciogonzalezvivo:
:+1: nice!

patriciogonzalezvivo:
yay!

PR #229, opened 2016-01-04
Add collision to points / text and sprites don't overlap

bcamper:
This PR extends the collision detection previously applied only to text labels so that it is also applied to features drawn with the point style (often sprites). `priority` is respected between these features to allow determination of which will be most likely to draw. (NB: this PR also fixes an issue where labels drawn with multiple styles derived from the `text` style would not collide with each other.) No new syntax is required, the behavior is automatically applied to points, though the `collide: false` parameter can be used to disable collision for points or text labels where desired.

output_9zbdfe
output_yyhhk0
output_5pr5ad

The primary architectural changes here are:

- Collision management is extracted into its own `Collision` class.
- `Point` style is refactored to use the same feature-queuing approach already used by the `Text` style: instead of immediately building features as they are submitted, they are added to a queue, which is then processed once all features in the tile have been parsed. This is because the collision phase must have access to all potentially occluding features before it can be performed.
- `Collision` class maintains a set of features to be collided, per tile (previously, the `Text` style maintained this set per tile, which has now been generalized).
- Each style with collideable features [notifies](https://github.com/tangrams/tangram/compare/point-collision?expand=1#diff-5865521f4fdb3091818c71745ae6a4edR163) the Collision manager that it has features to consider for collision.
- When the queue is complete, the style [submits](https://github.com/tangrams/tangram/compare/point-collision?expand=1#diff-5865521f4fdb3091818c71745ae6a4edR193) its collideable feature set to the Collision manager. That method returns a promise which is fulfilled once [all styles have submitted their features](https://github.com/tangrams/tangram/compare/point-collision?expand=1#diff-db559c158cbd05df9aacd09a95d7cb02R59), and the [collision between all of them](https://github.com/tangrams/tangram/compare/point-collision?expand=1#diff-db559c158cbd05df9aacd09a95d7cb02R115) has been performed. At that point, each style can continue to build only the features that survived the collision phase.
- This pattern allows for the addition of steps such as a collision phase which operate across styles, while keeping the rest of the feature/geometry building pipeline otherwise intact. The feature queuing pattern could be further generalized between `Point` and `Text` styles as a follow-up, however (even though `Text` derives from `Point`, there are still Canvas text-rendering-specific steps involving worker/main thread communication that require special handling).

Note: a remaining limitation of this architecture is that as collision is performed per tile, it is still possible for overlaps to occur where features cross tile boundaries. Such features must either be "pulled" into the tile boundaries (only acceptable for certain features and style cases), or be allowed to cross outside the tile bounds. While these artifacts are noticeable, the overall visual effect is much improved and eliminates the vast majority of overlaps. Speculatively, a more significant future overhaul could address this either by a) pooling all collideable features in a separate worker thread (challenging without a lot of overhead because [SharedWorkers](http://caniuse.com/#search=sharedworker) aren't supported beyond Chrome and FF), or b) a hybrid collision phase, where secondary collision only of features that cross tile boundaries is performed on the main thread.


nvkelso:
Wow, huge step forward! 

The Tokyo before/after reminds me there is a reference for "text" wrap for Japanese and Chinese character based languages, I'll dig that up.

bcamper:
@nvkelso @sensescape good call on the Asian language text wrapping, wouldn't be surprised if there was odd behavior there now (that we can't read :)  

I also noticed that some highway/major road labels seem to be getting displaced, they must have a lower priority than some of the POIs, but I wasn't sure if that was desirable at these zooms? But the behavior should match ES rendering, so we can compare with that too.

PR #230, opened 2016-01-11
Add screenshot support

bcamper:
Incorporates screenshot-related support from the demo into Tangram itself. This will be useful for automated render testing, for apps like Tangram Play that may want to capture the view periodically, and for other users who would like to incorporate screenshot support.

API changes:

- A new scene method, `Scene.screenshot()`, which queues a screenshot request, and returns a Promise that fulfills when the screenshot is available. (The screenshot must be "queued" because it cannot be captured immediately: we must ensure that the GL buffer is finished drawing, and then must capture the buffer contents just after rendering, before it is cleared by other operations.)
   - The promise resolves with an object containing two properties:
      - `url`: a data URL of the Canvas contents, suitable for loading into an `` or opening in a new tab/window
      - `blob`: a `Blob` of type `image/png`, suitable for saving to a file. Note that actually saving the blob to disk is left to the user or 3rd party library such as [FileSaver.js](https://github.com/eligrey/FileSaver.js/), as this goes a step beyond what can be considered generic Tangram functionality.
   - Example usage of requesting and saving a screenshot:
```
scene.screenshot().then(function(screenshot) {
    // uses FileSaver.js: https://github.com/eligrey/FileSaver.js/
    saveAs(screenshot.blob, 'tangram-' + (+new Date()) + '.png');
});
```
- A new scene event called `view_complete`, which fires when the view is in a "resting state", meaning new geometry is rendered, and no further tiles are loading/building. For example, when a scene is loaded, a `view_complete` event will fire when all tiles have loaded and the map renders. If the view is then zoomed in a level, another `view_complete` event will fire when the next zoom finishes rendering. `view_complete` can be subscribed to like other scene events:
```
scene.subscribe({
    view_complete: function () {
        console.log('scene view complete, rendered ' + 
            scene.render_count + ' primitives at: ' + 
            [scene.center.lng, scene.center.lat, scene.zoom].join('/'));
    }
});
```
- `Scene.screenshot()` can be paired with `view_complete` to automate screenshots, e.g. take a screenshot on initial scene view, take a screenshot for each new view, or a list of views for automated test generation, etc.

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -184,19 +184,6 @@ Enjoy!
     map.setView(map_start_location.slice(0, 2), map_start_location[2]);
     map.on('move', updateURL);
 
-    // Take a screenshot and save file
-    function screenshot() {
-        // Adapted from: https://gist.github.com/unconed/4370822
-        var image = scene.canvas.toDataURL('image/png').slice(22); // slice strips host/mimetype/etc.
-        var data = atob(image); // convert base64 to binary without UTF-8 mangling
-        var buf = new Uint8Array(data.length);
-        for (var i = 0; i < data.length; ++i) {
-            buf[i] = data.charCodeAt(i);
-        }
-        var blob = new Blob([buf], { type: 'image/png' });
-        saveAs(blob, 'tangram-' + (+new Date()) + '.png'); // uses FileSaver.js: https://github.com/eligrey/FileSaver.js/
-    }
-
     // Render/GL stats: http://spite.github.io/rstats/
     // Activate with 'rstats' anywhere in options list in URL
     if (url_ui && url_ui.indexOf('rstats') >= 0) {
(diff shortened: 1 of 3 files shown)


bcamper:
@louh @patriciogonzalezvivo Have wanted to move screenshot helpers into Tangram for awhile, doing it now to support some render automation @meetar is working on, and it's also relevant to Tangram Play (can replace some boilerplate screenshot code there currently, and be used for future features such as auto-capturing screenshot/thumbnail when saving a scene).

patriciogonzalezvivo:
:+1:

nvkelso:
+ @thisisaaronland and @zerebubuth.

thisisaaronland:
[this is good]

louh:
Sweet!

meetar:
:camera: :ok_hand:

PR #231, opened 2016-01-12
Add miter limit for line builder

bcamper:
Adds back the previously lost miter limit functionality for polyline builders. The miter limit converts miter joins to bevels when the join angle is very sharp, specifically when the miter join length is longer than the ratio of the `miter_limit` to the width of the line.  See [SVG miter limit](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit) for reference.

Exposes as `miter_limit` in `lines` style, default miter limit is 3., e.g.:

```
draw:
   lines:
      color: red
      width: 5px
      miter_limit: 2
```

Examples from [Hong Kong paths](http:tangrams.github.io/eraser-map/#20/22.27658/114.21159):

**Miter limit 3 (default)**
miter-3

**Miter limit 2**
miter-2

**Miter limit 1.5**
miter-1 5

### DIFF
--- a/src/styles/builders.js
+++ b/src/styles/builders.js
@@ -190,11 +190,18 @@ Builders.buildPolylines = function (
         texcoord_normalize,
         scaling_index,
         scaling_normalize,
-        join, cap
+        join, cap,
+        miter_limit
     }) {
 
     var cornersOnCap = cornersForCap[cap] || 0;         // default 'butt'
     var trianglesOnJoin = trianglesForJoin[join] || 0;  // default 'miter'
+
+    // Configure miter limit
+    if (trianglesOnJoin === 0) {
+        miter_limit = miter_limit || 3; // default miter limit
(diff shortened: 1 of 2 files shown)


bcamper:
@patriciogonzalezvivo I pulled this out from the more involved polyline builder branch. Can you check to make sure it looks OK on the current code base?

patriciogonzalezvivo:
:+1: Looks great. Way clear and elegant : )

nvkelso:
Nice! Setting the limit on some road classes at mid zooms could improve their rendering quite a bit, would adjusting this in the scene file cause a perf hit?

bcamper:
@nvkelso I wouldn't worry about that, technically the bevel is an extra triangle but it's not like a full rounded triangle fan and only happens on long miters anyway (also it looks like we're doing more work than necessary for bevels but @patriciogonzalezvivo is going to address that).

PR #232, opened 2016-01-12
Polylines caps joins simplify

patriciogonzalezvivo:
Splitting functionalities of ```addFan``` into individual functions ```addBevel``` and ```addSquare``` for efficiency. No chances on the API, nor on how it looks. Is mostly a performance improvement that need some benchmarking.


patriciogonzalezvivo:
Looking good!

PR #234, opened 2016-01-19
Add `density` parameter for native pixel density of textures

bcamper:
Adds a `density` parameter to texture definitions, to indicate the intended native pixel density that the texture was created for (e.g. mostly this means retina 1x or non-retina 2x, though other densities are possible). 

When style rules explicitly define the `size` of a sprite, this info isn't needed, because `size` is specified in CSS pixels, and the underlying texture is scaled to fit as necessary. But we found that in order to let sprites _automatically_ set their `size` appropriately, we needed this additional information about the sprite texture -- if the sprite was created for 1x resolution, it should be scaled up for retina display; if created for 2x resolution, scaled down on non-retina, etc.

`density` captures this information, and defaults to `1`. Example:

```
textures:
   pois:
      url: pois.png
      density: 2 # authored at 2x resolution
      sprites:
         # define sprites: [x origin, y origin, width, height]
         airport: [631, 105, 36, 38]
         ...
```

"Density" terminology was borrowed from HTML's relatively new [`srcset` attribute](http://www.w3.org/html/wg/drafts/html/master/semantics.html#attr-img-srcset) for supporting multiple resolution/density images (best image for current display is selected by browser). Alternatives could include `pixel_ratio`, from the browser's `window.devicePixelRatio`.

### DIFF
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -27,6 +27,7 @@ export default class Texture {
 
         this.name = name;
         this.filtering = options.filtering;
+        this.pixel_density = options.pixel_density || 1;
 
         // Destroy previous texture if present
         if (Texture.textures[this.name]) {
@@ -40,6 +41,7 @@ export default class Texture {
         this.sprites = options.sprites;
         this.texcoords = {};    // sprite UVs ([0, 1] range)
         this.sizes = {};        // sprite sizes (pixel size)
+        this.css_sizes = {};    // sprite sizes, adjusted for native texture pixel density
         log.trace(`creating Texture ${this.name}`);
     }
 
@@ -244,7 +246,9 @@ export default class Texture {
                 );
 
                 // Pixel size of sprite
+                // Divide by native texture density to get correct CSS pixels
                 this.sizes[s] = [sprite[2], sprite[3]];
+                this.css_sizes[s] = [sprite[2] / this.pixel_density, sprite[3] / this.pixel_density];
             }
         }
     }
@@ -268,7 +272,11 @@ Texture.destroy = function (gl) {
 // Get sprite pixel size and UVs
 Texture.getSpriteInfo = function (texname, sprite) {
     let texture = Texture.textures[texname];
-    return texture && { size: texture.sizes[sprite], texcoords: texture.texcoords[sprite] };
+    return texture && {
+        size: texture.sizes[sprite],
+        css_size: texture.css_sizes[sprite],
+        texcoords: texture.texcoords[sprite]
+    };
 };
 
 // Create a set of textures keyed in an object
@@ -332,9 +340,11 @@ Texture.getInfo = function (name) {
                 name: tex.name,
                 width: tex.width,
                 height: tex.height,
+                pixel_density: tex.pixel_density,
                 sprites: tex.sprites,
                 texcoords: tex.texcoords,
                 sizes: tex.sizes,
+                css_sizes: tex.css_sizes,
                 filtering: tex.filtering,
                 power_of_2: tex.power_of_2,
                 valid: tex.valid
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -27,7 +27,7 @@ export default class Texture {
 
         this.name = name;
         this.filtering = options.filtering;
-        this.pixel_density = options.pixel_density || 1;
+        this.density = options.density || 1;
 
         // Destroy previous texture if present
         if (Texture.textures[this.name]) {
@@ -248,7 +248,7 @@ export default class Texture {
                 // Pixel size of sprite
                 // Divide by native texture density to get correct CSS pixels
                 this.sizes[s] = [sprite[2], sprite[3]];
-                this.css_sizes[s] = [sprite[2] / this.pixel_density, sprite[3] / this.pixel_density];
+                this.css_sizes[s] = [sprite[2] / this.density, sprite[3] / this.density];
             }
         }
     }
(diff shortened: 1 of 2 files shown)


bcamper:
@blair1618 @nvkelso thoughts on terminology?

blair1618:
The terminology sounds fine to me, in fact I think just `density` could be sufficient in this context. 

However, this makes me wonder about supporting sets of multiple textures to support different densities (similarly to the HTML `srcset` or asset catalogs on iOS). Do you have an approach in mind for this? I could imagine specifying a texture "url" as a list of url/density pairs that tangram would then be able to select from.

bcamper:
Simplifying to just `density` works, will update.

Hmm... I agree about the benefit of being able to specify different images with different densities under the umbrella of the same logical texture. But I don't know that doing this under `url` would work, because the sprite definitions themselves would change at different densities. Essentially I think you would need to replicate the entire texture definition for each density level? I need to think more on possible syntax for that, does it make you think of anything?

bcamper:
Possible options (not sure I love any of them):

When `densities` is provided, it serves as a map of texture definitions (but then if there's a single definition is the `density` parameter still available, and that gets weird with multiple ways to specify):
```
textures:
   pois:
      densities:
         1:
            url: pois.png
            ...
         2:
            url: pois@2x.png
            ...
```

Or keep using the `density` param, but allow multiple to be specified as an array:
```
textures:
   pois:
      - url: pois.png
        density: 1
        ...
      - url: pois@2x.png
        density: 2
        ...
```

I wish we could find a way to leverage a convention like "@2x.png", but I'm unsure how that would work with sprite definitions.

blair1618:
I had imagined that sprite definitions wouldn't change for images provided at multiple densities, the coordinates will just be interpreted as if they are meant for a 1x image and then scaled accordingly. If we also adopt a naming convention for images to encode their respective densities, it could look something like:
```yaml
textures:
   pois:
      url:
         - pois.png
         - pois@2x.png
      sprites:
         # define sprites: [x origin, y origin, width, height]
         airport: [631, 105, 36, 38] # for @2x, this acts like [1262, 210, 72, 76]
         ...
```
If someone wanted to arrange their sprite sheet differently at different densities then they would be out of luck here, but that doesn't seem terribly onerous to me.

bcamper:
That would mean the sprite coordinates are specified in 1x pixel sizes and would be doubled for 2x? Maybe that makes sense, just seems a little odd to me since people are used to specifying sprite regions with absolute pixel coordinates (I think).

bcamper:
(Separately I would prefer not to enforce a file-naming scheme, but we could easily have a mapping between densities and URLs)

nvkelso:
I'm down with `density`.

It would be lovely to support the `@2x.png` and `@3x.png` conventions on iOS and the web here by default (and someday add overrides). 

I'm more concerned about the region definition for each @size file, though. I've found the interpolation between 2x and 3x is straightforward (linear), but from 1x to 2x it is not. 1x is just too small to be purely scale down, it ends up being slightly different artwork, which can mean slightly different regions.

Not to complicate things, but there's also the related concept of "icon size". Right now we do that be specifying size overrides and have Tangram interpolate the asset up/down (e.g.: large, medium, small), but that is producing noticeable visual artifacts (especially in Tangram JS) and it would be better to have an option for pre-composited assets. At least on Tangram JS, this is the largest problem.

Riffing on Matt's example:

```
      url:
         - pois.png
         - pois@2x.png
      sprites:
         # define sprites: [x origin, y origin, width, height]
         airport: [631, 105, 36, 38]
         airport@2x: [1262, 210, 72, 76]
         ...
```

If needed you could specify a different region for the other resolution, or if not provided it would auto scale.

bcamper:
Removing from 0.5 because it seems this is quickly becoming more complicated than expected...

blair1618:
Naturally, artwork will need to be tweaked at some scales to look good, otherwise we would just do the scaling automatically :)

I do however think we should enforce that sprite regions span the same range of density-corrected pixels in each image (i.e. define sprite regions once). iOS and OS X already enforce this for icon sets and Android is moving in that direction as well. Enforcing this will ensure consistent sprite size and collision behavior across all screen densities. The alternative, specifying sprite regions once for each density, seems wasteful and error-prone. 

For example, consider that we have an airport icon region that spans 36x38 pixels at 1x and 70x74 pixels at 2x. If icon size _is not_ specified in a draw rule, the density-corrected sprite size will be used, which is 36x38 at 1x and 35x37 at 2x. The difference in the two sizes means that the icons may collide with another label on a 1x display, but not on a 2x display. If icon size _is_ specified in a draw rule, the icon will be scaled to a non-integer ratio (and look fuzzy) at one or both of the densities. 

So yes, using only one set of sprite region definitions _does_ put a constraint on the user, but it's a constraint that keeps everyone sane and ends up producing a better map.

blair1618:
Plus, if we tried this and found that for some reason we really did need to specify regions `@2x`, we could do that later as a purely additive change.

nvkelso:
> Plus, if we tried this and found that for some reason we really did need to specify regions @2x, we could do that later as a purely additive change.

I'm fine with that supporting this later (we don't have any art for it today). 

I do know from personal experience in production systems this will be ideal to support longer term. Overall we should  aim for legibility over "correctness". Once the text size becomes variable (for universal access / accessibility reasons), it's more of a fluid dynamics simulation anyhow ;)

bcamper:
I pulled in the original piece of this -- the texture `density` parameter and auto-sizing of sprite display size to account for it -- into master (21 months later!). There is some good discussion here which can be used to extend support for separate images per density, later on.

In the immediate, this functionality helps with the addition of new 2x shield images, which would otherwise require tedious YAML to set the correct 1x size for each sprite.

bcamper:
Forgot to say, changes were applied in aca43dffad7ddaec1631cc5e0b01c47d7dd13739

PR #235, opened 2016-01-20
Label text can be defined as a series of fallbacks

bcamper:
This adds the ability for `text_source` to be an array of possible sources of label text, in priority order. Each array element is evaluated as if it was a `text_source` value (meaning each element can be either a *string* value that specifies a feature property by name, or a *function* that returns displayable label text). The first non-null evaluated source in the array is used as the label text.

The primary use case here is to support **preferred language** for text labels. For example: 

`text_source: [name:en, name]`

This will display an English label (`name:en`) when available, and will fall back to the default local `name` when not available.

For example, here is Moscow with default `name` only, and with the English preference as specified above:

output_gtv4rw

Looking to the future, once we have support for *scene properties* (see https://github.com/tangrams/tangram/compare/scene-properties), we will be able to control a preferred language with a single global scene variable, and even modify that preference at run-time (modify config and rebuild scene). For example (pending final scene property syntax):

```
scene:
   language: 'name:en' # prefer English

layers:
   labels:
      data: ...
      draw:
         text:
            text_source: [scene.language, name] # use scene preference, or fall back to local name
            ...
```

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -397,6 +397,7 @@ layers:
             filter: { kind: highway, $zoom: { min: 7 } }
             draw:
                 text:
+                    text_source: [name:en, name]
                     priority: 2
                     font:
                         fill: '#666'
@@ -408,6 +409,7 @@ layers:
             filter: { not: { kind: highway }, $zoom: { min: 13 } }
             draw:
                 text:
+                    text_source: [name:en, name]
                     priority: 5
                     font:
                         fill: '#666'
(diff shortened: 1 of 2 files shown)


bcamper:
@blair1618 I think this is relatively straightforward for ES as well, but can you please take a look?

nvkelso:
:+1:

nvkelso:
Confirming the tiles behavior on the server is to drop localized names if they are the same as the default name:

* https://github.com/mapzen/TileStache/blob/integration-1/TileStache/Goodies/VecTiles/transform.py#L518-L534

blair1618:
So the first "truthy" result is used? Yeah we should be able to manage that.

tallytalwar:
This should be fine on ES also.

PR #237, opened 2016-01-21
Filter lines on all tile edges

bcamper:
Adjusts our tile edge filtering (which removes line segments that fall on tile edges, as they are generally not desirable to draw) so that it will filter out lines that fall on **any** tile edge, not just the edges of the "local" tile. This is done by simply modding the coordinates against the tile size.

This is useful for cases where geometry is clipped to some other tile multiple, e.g. 3-tile bbox range. See https://github.com/mapzen/vector-datasource/issues/197 for background.

Before:
tangram-1453403245861

After:
tangram-1453403090822


bcamper:
alert @blair1618

PR #241, opened 2016-02-04
Support arbitrary winding order

bcamper:
Currently, Mapzen's own vector tiles have a CCW winding on polygon rings, which is opposite what the MVT format [currently specifies](https://github.com/mapbox/vector-tile-spec/tree/master/2.1) (Geo/TopoJSON do not have an officially specified winding order). We plan to rectify this in an upcoming tile release (see https://github.com/mapzen/mapbox-vector-tile/pull/21), but it raised the issue that Tangram ideally ought to support data sources with either winding order, for maximum compatibility.

This PR enables that by using the order of the first polygon ring processed in each data source, to indicate the expected order of outer rings. This allows for automatic switching between CW or CCW-wound polygons on a per-data-source level, even mixing sources with different winding orders. (Note this assumes that the data source is internally consistent in its winding between polygons, but there is no rational way to support a source that is not.)

See related issue to add support in ES: https://github.com/tangrams/tangram-es/issues/519

### DIFF
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -10,6 +10,7 @@ export default class DataSource {
         this.name = source.name;
         this.url = source.url;
         this.pad_scale = source.pad_scale || 0.0005; // scale tile up by small factor to cover seams
+        this.default_winding = null;
         this.enforce_winding = source.enforce_winding || false; // whether to enforce winding order
 
         // Optional function to transform source data
@@ -109,15 +110,34 @@ export default class DataSource {
                         if (this.enforce_winding) {
                             Geo.enforceWinding(feature.geometry, 'CCW');
                         }
+                        // Use first encountered polygon winding order as default for data source
+                        else {
+                            this.updateDefaultWinding(feature.geometry);
+                        }
                     });
                 }
             }
+
+            dest.default_winding = this.default_winding || 'CCW';
         });
     }
 
     // Sub-classes must implement
     _load(dest) {
         throw new MethodNotImplemented('_load');
+    }
+
+    // Infer winding for data source from first ring of provided geometry
+    updateDefaultWinding (geom) {
+        if (this.default_winding == null) {
+            if (geom.type === 'Polygon') {
+                this.default_winding = Geo.ringWinding(geom.coordinates[0]);
+            }
+            else if (geom.type === 'MultiPolygon') {
+                this.default_winding = Geo.ringWinding(geom.coordinates[0][0]);
+            }
+        }
+        return this.default_winding;
     }
 
     // Register a new data source type, under a type name
--- a/src/styles/builders.js
+++ b/src/styles/builders.js
@@ -87,7 +87,8 @@ Builders.buildExtrudedPolygons = function (
         tile_edge_tolerance,
         texcoord_index,
         texcoord_scale,
-        texcoord_normalize
+        texcoord_normalize,
+        winding
     }) {
 
     // Top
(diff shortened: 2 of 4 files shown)


bcamper:
I'm realizing this also may mean that we can/should deprecate the `enforce_winding` parameter, since it should be moot with automatic winding order adaptation. Will verify.

bcamper:
cc @blair1618 does this make sense

tallytalwar:
This makes sense to me too, also agree with the `enforce_winding parameter. We also do not have `enforce_winding` supported in es!

blair1618:
I'm curious why you've made winding a property at the data source level; is it not sufficient to check the winding of the first polygon in each feature and take that to be the winding of all exterior polygons?

bcamper:
@blair1618 winding order could vary between data sources (but should be consistent within any one), for example mixing (currently backwards) Mapzen tiles with Mapbox in the same scene (not necessarily very useful given the data overlap, but valid).

blair1618:
Right, we clearly don't want the whole scene to save one winding order for everything, but I'm wondering why the data source has to save the winding order at all when we should be able to determine it for each feature as they are processed.

bcamper:
Oh yes, sorry. Two reasons:

- Since it's expected (reasonably required) to be consistent within the data source, it's wasteful to check once it's been determined the first time.
- Given the current JS data flow, it was simply easier to to record it at the data source level (where it's inferred, then pass that to the tile, which is then available to the polygon builder). The data source tile loading code *could* have just set the property on the tile and not saved it, but it's saved for efficiency as mentioned above.

I think the first point is valid for ES, the second is mostly implementation-specific.

blair1618:
Ok, I see. In tangram-es it will probably be easier to implement this as a per-feature check - we have to evaluate the winding of every incoming polygon anyway so I don't think it's super wasteful. 

The reasoning here sounds good to me!

bcamper:
Yep. We need to evaluate the winding order for each polygon from an *MVT* source (to split out multi-polygons encoded as a single ring list), but not (at least in JS) for each features coming from a Geo/TopoJSON source. And even for the MVT case, in the JS code, the extruded polygon builder (which needs the winding order) is an isolated piece of code that wouldn't otherwise have access to that data, and I didn't want to check winding order twice. Am I missing something here?

blair1618:
Yep, all correct. In ES we decode the geometry from all data sources into a data structure called `TileData` in which winding is normalized to a known direction. This means that we only need to care about winding order at one step of the pipeline: the MVT -> `TileData` decoding. So when we are parsing the geometry of an MVT feature, we evaluate the winding of each ring *once*. If this is the first ring in a feature, it's automatically an exterior ring and the winding direction is made the "reference" winding for that feature. We then process every other ring relative to that reference, making them interior rings or new exterior rings appropriately. At the end we get our `TileData` with known polygon winding that we can pass to builders without the need to specify a winding direction. 

Does that help explain how I'm thinking about this?

hjanetzek:
@blair1618 currently we check for winding order after the first ring in the mvt source to determine when the next polygon starts (assuming that polygons start with ccw rings).
Earcut checks it again and enforces winding where outer rings are in clockwise order. Looks like there is some room for optimization to skip this check in earcut when we require TileData::polygons to have a fixed order.

bcamper:
@blair1618 yep thanks for the additional detail. I was explicitly avoiding always normalizing the winding order because of the added cost of manipulating the data, but that cost will differ across platforms. Anyway not for this PR really, sounds like we are OK with the logic!

blair1618:
@hjanetzek Good call. I think I slightly muddled the point in the pipeline where that normalization happens :\ But you're right that we might be able to make earcut's job easier there.

PR #242, opened 2016-02-04
Exclude line segments that fall outside the tile bounds

bcamper:
Our previous tile edge removal logic only removed line segments that were *on* (or nearly on, within a tolerance) the tile bounds. This works for unbuffered tiles like Mapzen's, which are clipped to tile boundaries, but does not render properly for buffered tiles, such as Mapbox's.

This approach simply throws out line segments that lie outside the tile bounds (retaining the tolerance). Along with https://github.com/tangrams/tangram/pull/241, ensures proper support for Mapbox-served MVT tiles.

Before/after examples, where the "double lines" visible with the previous logic are lines at the edge of the tile buffer area (so one buffer for each adjacent tile, two lines total), which were not removed, but now are no longer rendered:

output_g5tx0t

output_9x4pzh

### DIFF
--- a/src/styles/builders.js
+++ b/src/styles/builders.js
@@ -694,10 +694,10 @@ Builders.outsideTile = function (_a, _b, tolerance) {
     let tile_max = Builders.tile_bounds[1];
 
     // TODO: fix flipped Y coords here, confusing with 'max' reference
-    if ((_a[0] < tile_min.x + tolerance && _b[0] < tile_min.x + tolerance) ||
-        (_a[0] > tile_max.x - tolerance && _b[0] > tile_max.x - tolerance) ||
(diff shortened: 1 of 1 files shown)


bcamper:
Note: investigating some small grey building polygons that are unexpectedly popping in/out in those screens. I believe that these are from an older build, pre-https://github.com/tangrams/tangram/pull/241, and thus were not drawn properly due to winding order issues, rather than code in this PR. Will confirm.

bcamper:
Confirmed building polygon artifacts are not from this PR. Example area from above, with correct building rendering, before the current PR:

screen shot 2016-02-04 at 3 07 31 pm

bcamper:
@blair1618 I think this is working and is actually simpler than before? I have yet to find any issues with Mapzen tiles, though I admit I've only really be able to spot-check several different areas (automated visual testing soon FTW).

PR #247, opened 2016-02-16
Raster tile support

bcamper:
Adds support for raster tile rendering, with addition of new data source and built-in styles:

- `Raster` data source type
  - Requires a `url` tile template, in the same manner as other tiled sources.
  - Optionally allows the GL texture `filtering` mode to be set, with the same allowed values (`mipmap`, `linear`, `nearest`) as for other texture definitions in the scene file. Default is `mipmap` for power-of-2 images, and `linear` for non-power-of-2 images (raster tiles are generally power-of-2, though other sizes would be scaled to fit the tile square). Setting `filtering: nearest` allows for the raster tiles to be pixelated when scaled past their `max_zoom`.
  - Optionally defines a `max_zoom` beyond which tiles are over-zoomed. As with other data sources, defaults to `18`.
  - Implementation-wise, the raster source simply synthesizes a single quad geometry covering the entire tile area, and accompanying texture UVs from [0, 1].
  - Example:
   ```
   sources:
       stamen-terrain:
           type: Raster
   ```
- `raster` built-in rendering style
  - Raster data sources must be added to a `layer` and `draw` group before they are rendered. This is done with the `raster` style.
  - `order` determines the layer order as with other styles.
  - `color` defaults to white and any other value set by the draw group will *tint* the raster texture. For example, setting `color: [0.5, 0.5, 0.5]` will darken the underlying tiles by 50%.
  - Implementation-wise, the raster style is simply a textured polygon, that automatically samples from the provided raster source for that tile.
  - Example:
   ```
   layers:
       terrain:
           data: { source: stamen-terrain }
           draw:
               raster:
                   order: 0 # draw on bottom
                   color: [0.5, 0.5, 0.5] # darken by 50%
   ```
tangram-tue feb 16 2016 18-28-13 gmt-0500 est
- Custom raster styles can be defined in `styles` with `base: raster`.
  - The `raster` style is derived from the `polygons` rendering style, and provides the same shader blocks. 
  - Raster-specific uniforms are defined for:
    - `u_raster_texture`: texture sampler (`sampler2D`) for that tile's texture
    - `u_raster_texture_size`: `vec2` of width/height pixel dimensions of the raster texture
    - `u_raster_texture_pixel_size`: `vec2` of UV size of a single pixel in the raster texture, e.g. inverse of width/height or 1/256 for a 256x256 image, for sampling nearby pixels
  - Example style that converts raster tiles to greyscale:
   ```
    greyscale:
        base: raster
        shaders:
            blocks:
                filter: |
                    float luma = dot(color.rgb, vec3(0.299, 0.587, 0.114));
                    color.rgb = vec3(luma);
   ```
tangram-tue feb 16 2016 18-27-50 gmt-0500 est
  - That example could itself be extended to apply a colorize effect, in this case yellow:
   ```
    colorize:
        mix: greyscale
        shaders:
            uniforms:
                colorize: [0.9, 0.9, 0.2] # yellow
            blocks:
                filter: |
                    color.rgb *= colorize;
   ```
tangram-tue feb 16 2016 18-28-00 gmt-0500 est

### DIFF
--- a/src/gl/shader_program.js
+++ b/src/gl/shader_program.js
@@ -375,7 +375,8 @@ export default class ShaderProgram {
     setTextureUniform(uniform_name, texture_name) {
         var texture = Texture.textures[texture_name];
         if (texture == null) {
-            texture = Texture.create(this.gl, texture_name, { url: texture_name });
+            log.warn(`Can't find texture '${texture_name}'`);
+            return;
         }
 
         texture.bind(this.texture_unit);
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -17,6 +17,7 @@ export default class Texture {
         this.bind();
 
         this.name = name;
+        this.retain_count = 0;
         this.source = null;
         this.source_type = null;
         this.config_type = null;
@@ -32,6 +33,9 @@ export default class Texture {
 
         // Destroy previous texture if present
         if (Texture.textures[this.name]) {
+            // Preserve previous retain count
+            this.retain_count = Texture.textures[this.name].retain_count;
+            Texture.textures[this.name].retain_count = 0; // allow to be freed
             Texture.textures[this.name].destroy();
         }
 
@@ -45,6 +49,10 @@ export default class Texture {
 
     // Destroy a single texture instance
     destroy() {
+        if (this.retain_count > 0) {
+            log.error(`Texture '${this.name}': destroying texture with retain count of '${this.retain_count}'`);
+        }
+
         if (!this.valid) {
             return;
         }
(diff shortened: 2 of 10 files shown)


bcamper:
@blair1618 can you look this over, especially as regards uniform naming?

mjcunningham:
:raised_hands:

blair1618:
So `u_raster_texture_pixel_size` is just `vec2(1.0) / u_raster_texture_size`? Is there a lot of value in having a dedicated uniform value for that? If we keep it we may want to name it something else, it could be mistaken for the size of the raster image in pixels.

bcamper:
I figured it was worth putting the single-pixel UV increment in as a uniform because you need it if you are going to do anything interesting with sampling, and why have to compute it for every pixel. But maybe I am overly cautious on fragment shader performance.

I agree the name could be confusing though, suggested alternatives?

bcamper:
See #282 instead.

PR #250, opened 2016-02-23
Proxy tiles

bcamper:
Adds "proxy tile" support: while tiles are loading, previously loaded tiles from higher or lower zooms will be shown in their place, providing more visual continuity and reducing "breaks" or "flashes" between tile zooms.

- Proxy retrieval: To speed up retrieval of proxy tiles, a new `TilePyramid` object tracks ancestor/descendant relationships across zoom levels.
  - `getAncestor()` and `getDescendants()` methods recursively find available proxy tiles above or below a given tile (for descendants, the max level to descend can be set, since there is up to 4x increase in number of potential proxy tiles for each level; current max depth default is 3 zooms).
 - There is special handling for over-zooming in `TilePyramid` that could be simplified, and in the future should be generalized to support over-zooming *between* intermediate zooms, in addition to past a max zoom (allow data sources to specify partial zoom range coverage, e.g. load tiles for z6, over-zoom them and skip z7, then load new tiles at z8). There is also some redundancy in tile indexing between `TilePyramid` and `TileManager` that could be improved.
- Labels:
  - Because labels are placed/collided once at tile load time, assuming a 1x view size, labels will overlap in undesirable ways when zoomed out for display as descendant proxy tiles. The current solution to this is to [fade out](https://github.com/tangrams/tangram/compare/proxy-view?expand=1#diff-60e93cbad892a92b22ab3f8baaf17594R66) labels on zoom out; currently, labels will become fully transparent when zoomed out 50%, as this appears to be a decent compromise.
  - Zoom-fade could also be applied to sprites/POIs, though this has not been done yet, as icon overlap (subjectively) appears both less problematic and can help retain visual consistency when icons would otherwise fade out and then back into the same place (Tangram ES addresses this with custom logic to prevent immediate fade-out-in, but this is not easily implementable in JS due to the current one-time label build pattern).
- Depth:
  - Similar to Tangram ES, proxy tiles are drawn behind the tiles they are proxying for, e.g. as the new tiles load, they are drawn in front of and appear to replace the proxy tiles. Because proxy tiles can be drawn at multiple mixed zooms (if zooming across several levels), each proxy tile is offset one additional `order` value for each zoom level away from the tile it is proxying.
  - For example, a z15 tile proxying a z16 one would be drawn 1 additional order behind, while a z14 tile proxying the same z16 tile would be drawn 2 orders behind.
- Scene/View refactor:
  - This PR also includes a significant reorganization, pulling much of the view logic into a new `View` object (which also holds the `Camera` and GL model/view matrices).
  - Related model matrix calculation is moved into `Tile`.
  - There is much remaining room for improvement of logic / data flow / separation of concerns between `Scene`, `TileManager`, and `View`, but this refactor provides a better basis for future work, and is closer to the Tangram ES architecture.

greenlake-zoom

uw-zoom

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,20 +45,21 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "babel": "5.8.29",
     "box-intersect": "1.0.0",
     "csscolorparser": "1.0.2",
     "earcut": "1.2.0",
-    "match-feature": "tangrams/match-feature#v1.0.0",
-    "gl-matrix": "2.2.1",
+    "geojson-vt": "2.1.6",
+    "gl-mat3": "1.0.0",
+    "gl-mat4": "1.1.4",
+    "gl-shader-errors": "1.0.3",
     "js-yaml": "3.4.6",
     "loglevel": "1.2.0",
-    "strip-comments": "0.3.2",
-    "gl-shader-errors": "1.0.3",
+    "match-feature": "tangrams/match-feature#v1.0.0",
     "pbf": "1.3.2",
-    "vector-tile": "1.1.2",
+    "strip-comments": "0.3.2",
     "topojson": "1.6.19",
-    "geojson-vt": "2.1.6",
-    "babel": "5.8.29"
+    "vector-tile": "1.1.2"
   },
   "devDependencies": {
     "babel-runtime": "5.8.29",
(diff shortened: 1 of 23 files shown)


bcamper:
Currently investigating why tests are failing with timeouts on Circle for this branch (but not others) - they run locally.

nvkelso:
:100:

bcamper:
Tests fixed!

PR #263, opened 2016-03-07
Global property substitution

bcamper:
This PR adds a concept of **global properties**, which are user-defined properties in the scene file that can be substituted for values elsewhere in the file. This is useful for values that are referenced in multiple places in the scene file, and would benefit from a centralized definition for easier editing. Examples include common colors, language preferences, or visibility flags used to tweak styles.

Global properties are scoped as everything under the top-level `global` block of the scene file, including nested properties:

```
global:
   labels: true # label visibility flag
   language: name:en # preferred language for label display
   colors: # group common colors together
      color1: red
      color2: green
      color3: blue
```

These properties can then be substituted elsewhere in the scene file with `global.` syntax (included nested dot notation, `global.group.property`), and are also available in JS function filters and properties under the same syntax:

```
layers:
   road-labels:
      data: { ... }
      visible: global.labels
      draw:
         text:
            text_source: [global.language, name] # display labels in preferred language, fallback to default name
            ...

   point-labels:
      data: { ... }
      draw:
         text:
            text_source: |
               function() {
                  // Make a compound label with "Preferred Name (Local Name)"
                  var preferred = feature[global.language];
                  if (preferred && preferred !== feature.name) {
                     return preferred + '\n(' + feature.name + ')';
                  }
                  return feature.name;
               }

   landuse:
      data: { ... }
      draw:
         polygons:
            color: global.colors.color2 # nested property
            ...
```

Global properties are similar to YAML's native anchor/reference capability, with the most significant difference being that they are applied *after* YAML parsing but *before* other Tangram parsing. This means we can:

- Alter them programmatically and rebuild geometry. For example, this enables a dynamic language preference setting. In the example above, the `global.language` property could be hook up to a language selector dropdown or app preference.
- Import and optionally override scene properties once we have scene `import` functionality (e.g. a base scene `A` could define properties which are then referenced by a scene `B` that has an `import: A`). This is outside the scope of current Tangram functionality, but coming soon.

### DIFF
--- a/src/scene_loader.js
+++ b/src/scene_loader.js
@@ -152,7 +152,7 @@ export default SceneLoader = {
     // Normalize some scene-wide settings that apply to the final, merged scene
     finalize(config) {
         // Replace scene properties
-        config = StyleParser.applySceneProperties(config);
+        config = StyleParser.applyGlobalProperties(config);
 
         // Assign ids to data sources
         let source_id = 0;
--- a/src/scene_loader.js
+++ b/src/scene_loader.js
@@ -151,7 +151,7 @@ export default SceneLoader = {
 
     // Normalize some scene-wide settings that apply to the final, merged scene
     finalize(config) {
-        // Replace scene properties
+        // Replace global scene properties
         config = StyleParser.applyGlobalProperties(config);
 
         // Assign ids to data sources
--- a/src/scene_loader.js
+++ b/src/scene_loader.js
@@ -2,6 +2,7 @@ import Utils from './utils/utils';
 import GLSL from './gl/glsl';
 // import mergeObjects from './utils/merge';
 import {StyleManager} from './styles/style_manager';
+import {StyleParser} from './styles/style_parser';
 
 var SceneLoader;
 
@@ -150,6 +151,9 @@ export default SceneLoader = {
 
     // Normalize some scene-wide settings that apply to the final, merged scene
     finalize(config) {
+        // Replace scene properties
+        config = StyleParser.applySceneProperties(config);
+
         // Assign ids to data sources
         let source_id = 0;
         for (let source in config.sources) {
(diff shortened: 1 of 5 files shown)


blair1618:
Yeah, regarding the issue of name collision between built-in and user-provided values: It makes some sense to me to create a separate top-level element just for user-provided values. This prevents collision with any of the scene properties that have built-in semantics (`background`, `animated`, and any future properties we add). And looking at the number of anchored values in eraser-map (which this mechanism would presumably replace) the collection of custom properties could be quite large, probably large enough to merit its own top-level element simply for organizational purposes. 

If we were to do this, the challenge would be giving this element a name that is as succinct and intuitive as 'scene'. Possibilities: `global`, `prefs`, `my`, `meta`

bcamper:
I would probably lean towards `global`. We're using it for shader blocks, so the term is in use (in a different context, which could be confusing, though we have `color` for both draw params and shader blocks so that's... consistent anyway :)

- `vars`: maybe some benefit in sounding like the new CSS Variables (but they use `var` keyword and can cause confusion with JS `var` inside functions anyway).
- `props`: I don't generally like abbreviations, and we already have layer-level `properties`
- `my`: perl flashbacks
- `meta`: these don't actually seem like metadata
- `_`: could put them in a YAML-safe character, like jquery or underscore, succinct (`_.language`) but probably arcane syntax wise (and we can't use `$`, which is probably easiest to understand, because it's already used for `$zoom` and `$layer` special properties)
- `settings`: too long?

bcamper:
@blair1618 switching to `global`

bcamper:
(updated PR description to reflect `global` syntax)

bcamper:
@blair1618 @tallytalwar also moved ahead with deprecating layer-specific `properties`, as we concluded (we think anyway :) that anything their scope of behavior can be accomplished with the new `global` properties.

@meetar we will need to convert some styles that used layer `properties`, like the Explorer, to use this new syntax. Should be a straightforward change though (and less duplicated code!).

Issue #259, opened 2016-03-03
overzooming tiles distorts some shader patterns


nvkelso:
In the [Eraser Map scene file](https://github.com/tangrams/eraser-map/blob/nvkelso/v0.8-tiles-sort_key-orders/eraser-map.yaml), when I update the sources block's max_zoom from 18 to 16, the landuse dot patterns in the shader seem to freeze size at zoom 16 and then get larger and larger, like they're treated in meter space instead of pixel space.

/cc @sensescape 

**observed:**
<img width="547" alt="screen shot 2016-03-02 at 17 03 16" src="https://cloud.githubusercontent.com/assets/853051/13481557/7eaf70a8-e09b-11e5-8368-21d100efafe8.png">

**expected:**
<img width="612" alt="screen shot 2016-03-02 at 17 03 25" src="https://cloud.githubusercontent.com/assets/853051/13481556/7e97fe8c-e09b-11e5-8030-cd66773f5c89.png">

blair1618:
I wonder - does the `modelPosition` accessor in our shaders account for over-zooming? I think right now it may be returning 0 and 1 at the limits of the larger tile that is being over-zoomed, but for re-styling it may be more useful to return 0 and 1 at the limits of the simulated tile at the higher zoom level.

bcamper:
This makes sense, I'll give it a try.

bcamper:
Will be fixed by https://github.com/tangrams/tangram/pull/266. Verified in desktop and mobile browsers.

nvkelso:
Will the fix be part of the v0.5 Tangram series? All the new v2.0 scene file slippy map demos are pegged to that, and all suffer from this problem (especially building outlines at high zooms).

bcamper:
It's in the 0.6 release I'm prepping now... but we'll want to update our
demos to use that for proxy tiles anyway ;)

On Wed, Mar 9, 2016 at 8:10 PM, Nathaniel V. KELSO <notifications@github.com

bcamper:
Released with 0.6.0.

PR #266, opened 2016-03-08
Scale modelPosition() to account for tile overzooming

bcamper:
We concluded in https://github.com/tangrams/tangram/issues/259 that we should scale the `modelPosition()` accessor to account for tile overzooming, providing a consistent scale for shader-based patterns. 

This is done by adding an additional scaling factor of `2^(tile style zoom - tile coord zoom)` to the `modelPosition()`.

Implementation notes:
- `u_tile_origin.z` was already set to the tile's *style zoom* (the zoom at which the tile is being displayed) in JS, but to the *coordinate zoom* (e.g. the zoom at which the tile was loaded) in ES. We concluded that it should be set to the **style zoom** on both platforms.
- However we also need access to the tile coordinate zoom, which is now provided as `u_tile_origin.w`.
- Previously, JS was using `u_tile_origin.w` as the depth offset for proxy tiles. This was moved to a new uniform `u_tile_proxy_depth`, however this will not be documented/exposed as a public "API" uniform (unlike `u_tile_origin`, which is available to users).
- `modelPosition()` has never really been in a true `[0, 1]` range, because geometry can go outside the tile bounds. This means it cannot safely have a modulus or `fract` applied. Instead we've opted to simply scale it, meaning the scale of its values is consistent with what the user would expect for a given zoom level, but the actual values cover a greater range.

### DIFF
--- a/src/gl/shaders/accessors.glsl
+++ b/src/gl/shaders/accessors.glsl
@@ -1,9 +1,17 @@
-// Vertex position in model space: [0, 1] range over the local tile
-// Note positions can be outside that range due to unclipped geometry, geometry higher than a unit cube, etc.
 #ifdef TANGRAM_VERTEX_SHADER
 
+// Vertex position in model space: [0, 1] range over the local tile
+// Note positions can be outside that range due to unclipped geometry, geometry higher than a unit cube, etc.
 vec4 modelPosition() {
-    return vec4(SHORT(a_position.xyz) / TANGRAM_TILE_SCALE, 1.) + vec4(0., 1., 0., 0.);
+    return
+        vec4(
+            SHORT(a_position.xyz)                       // scale normalized short to full range
+            / TANGRAM_TILE_SCALE                        // scale coords to ~0-1 range
+            * exp2(u_tile_origin.z - u_tile_origin.w),  // adjust for tile overzooming
+        1.)
+        + vec4(0., 1., 0., 0.);
+        // NB: additional offset to account for unusual Tangram JS y coords,
+        // should be refactored to remove
 }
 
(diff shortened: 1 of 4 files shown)


bcamper:
cc @blair1618

blair1618:
In the corresponding change set for ES, I noticed a small difference in the way line extrusion is being calculated [in ES](https://github.com/tangrams/tangram-es/blob/overzoom-modelposition/core/resources/shaders/polyline.vs#L73) and [in JS](https://github.com/tangrams/tangram/blob/master/src/styles/polygons/polygons_vertex.glsl#L65)

In ES I found that line extrusion was only correct when `dz` was calculated using the "coordinate zoom". The vertex coordinates and extrusion vectors were produced at the scale of the "coordinate zoom", so it seems to me that this should be the value compared to the map zoom. Does the current code in JS result in correct extrusion scaling for over-zoomed tiles?

bcamper:
Lines are scaling properly in JS for overzooming. That suggests to me that there's a difference in the meters-per-pixel scale being used when the extrusion vectors are calculated. In JS, the feature's style zoom is used for all styling-related calculations, of which line width / extrusion is included. If it were based on coordinate zoom instead, then it would be expected to see the line width scaling up (incorrectly) with overzooming?

blair1618:
Ah, yes! The line-width evaluation in ES currently uses the "coordinate zoom" of the tile, not the "styling zoom" - so both implementations are internally self-consistent. Once this corresponding PR in tangram-es is merged, we could switch to evaluating width using the "styling zoom" for consistency (before we only had the "coordinate zoom" available in shaders).

PR #268, opened 2016-03-09
update order

meetar:


bcamper:
I'm actually not sure we need/want to update this one, since it's just a test fixture (I already updated the bundled demo in `demos/`, which did need it).

meetar:
Fine by me – I was just on a roll and noticed it :)

bcamper:
We can keep the tests simpler without this one.

meetar:
:+1:

PR #271, opened 2016-03-11
Revise lighting blocks: always include `color` and `normal`

bcamper:
This is yet another revision of shader block application, after the v0.6.0 changes revealed compatibility issues with existing styles and limitations on procedurally generated textures and custom lighting setups.

With this new logic, the `color` and `normal` blocks are always injected, regardless of the `lighting` mode:

- `lighting: fragment`:
  - `normal` and `color` blocks executed in the fragment shader.
- `lighting: vertex`:
  - `normal` block is executed in the vertex shader
  - Lighting intensity (e.g. lighting if applied to a pure white surface) is calculated in the vertex shader, and interpolated in the fragment shader.
  - `color` block is executed in the fragment shader, and the result is multiplied by the interpolated lighting intensity. This allows for `color` to create procedural textures that are still properly lit when `lighting: vertex` is applied.
- `lighting: false`:
  - `normal` and `color` blocks executed in the fragment shader. No Tangram lighting is applied, but the user can create alternate shading models in these blocks if desired.

In all cases, the `filter` block is executed last in the fragment shader, as a final color pre-processing step.

### DIFF
--- a/src/styles/polygons/polygons_fragment.glsl
+++ b/src/styles/polygons/polygons_fragment.glsl
@@ -19,6 +19,10 @@ varying vec4 v_world_position;
     varying vec2 v_texcoord;
 #endif
 
+#if defined(TANGRAM_LIGHTING_VERTEX)
+    varying vec4 v_lighting;
+#endif
+
 #pragma tangram: camera
 #pragma tangram: material
 #pragma tangram: lighting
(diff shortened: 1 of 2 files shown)


bcamper:
@patriciogonzalezvivo this is a revised (and I think the most flexible) approach, after discussing further with @blair1618. Can you please take a careful look and think through your existing styles? I loaded several of them (from before the modifications we made last night to "fix" them), and it all looked good, to me, but I'd like to settle on this once and for all :)

patriciogonzalezvivo:
:+1:  seams like a solid approach. I'm excited to use vertex lighting more.

bcamper:
@patriciogonzalezvivo can you verify your styles work as expected on this build? Including Line of Sight etc.

blair1618:
Tried this out in tangram-es and it doesn't seem to cause major problems. I had to make one change to our default scene where a `color` block referenced the `position` vector, which I had to replace with `worldPosition()` (`position` isn't provided in the fragment shader).

patriciogonzalezvivo:
:+1: :+1:  checked.... this fix works on LineOfSight

bcamper:
Cool, moving ahead with this for

bcamper:
Cool, moving ahead with this for Cool, moving ahead with this for v0.6.1 release.

bcamper:
@blair1618 good to go on this? Would prefer to not have to do yet another revision :)

blair1618:
Yep! We can roll out the same change in tangram-es today

PR #275, opened 2016-03-15
Uv y meters

patriciogonzalezvivo:
screen shot 2016-03-15 at 9 05 52 am


blair1618:
Can I suggest that we start by not exposing the scaling factor? It would be easy and non-breaking to add later, it might make sense to first see what the actual design needs are. 

It looks like distances here are effectively normalized to 1 unit per tile-width, is that right? Using tile size as a baseline for distance makes sense, it's the natural scale-independent metric. What about scaling distance to 256 units per tile-width? It won't exactly match "logical pixels" at all views, but it might be easier or a designer to reason about than tile size.

bcamper:
@blair1618 agreed on not exposing the texcoord ratio initially, done in https://github.com/tangrams/tangram/pull/275/commits/482724a400daafaa12e0d850726fc388b3ccb5d3. 

I'm seeing this PR as a way to soft-launch this feature with v0.7.0. It will change default behavior for polyline UVs to have a ratio of 1 (square) (this is a change in visual behavior, though not a public API syntax change, but these UVs are seldom used anyway currently, generally only in styles like Tron).

bcamper:
Quick Bubble Wrap test update (will need to file issue to fix more generally) -- @nvkelso @sensescape 

output_3llxqv

meetar:
:straight_ruler: :100:

bcamper:
This is good to go for initial release anyway, future integration work will likely happen around tying this into repeated sprites, as well as possibly smarter line patterns across tile boundaries (if/when we have hints in the tiles to help us join up lines that have been clipped).

nvkelso:
@zerebubuth re tile comment.

sensescape:
nice!

patriciogonzalezvivo:
:+1:

Issue #273, opened 2016-03-14
Line labels are too sparse on overzoomed tiles


bcamper:
When tiles are overzoomed, the line labels can become too sparse. This is because we place one label per line segment. When tiles represent a constant screenspace area at every zoom (as unzoomed tiles do), this has provided a decent level of label density. But when those line segments are scaled up to cover several times the original screen area (16x for z16 tiles overzoomed to z20!), the labels get spread out over a much larger area.

To solve this, we can introduce some optional factor for placing labels at subdivided points along the line geometry, which can be adjusted according to the tile's overzoom factor (if any).

z19 with z19 tiles:
![tangram-mon mar 14 2016 16-03-00 gmt-0400 edt](https://cloud.githubusercontent.com/assets/16733/13758136/471bd5ac-e9fe-11e5-9d67-1cda9f303581.png)

z19 with z16 overzoomed tiles:
![tangram-mon mar 14 2016 16-03-07 gmt-0400 edt](https://cloud.githubusercontent.com/assets/16733/13758141/51ede9ca-e9fe-11e5-849d-2d142976d0f3.png)

bcamper:
Closed via #277 and released in [v0.6.2](https://github.com/tangrams/tangram/releases/tag/v0.6.2).

PR #277, opened 2016-03-18
Subdivide line labels by segment on overzoomed tiles

bcamper:
We've found that our line labelling logic is insufficient for overzoomed tiles, because it only places one label per line, which often provides enough density at each tile's "native" zoom, but is too sparse (sometimes dramatically so) when tiles are overzoomed multiple levels. See #273.

This PR increases line label density by attempting to place multiple labels per line. For each line, up to `2^dz` tiles are placed, where `dz` is the number of levels of overzooming being applied. For example, a z16 tile would still place 1 (2^0) label at z16, but when displayed at z18 the same line would place up to 4 (2^2) labels. This is done with a "coarse subdivide", splitting the line by segment. Each label placement is assigned a range of segments in which it can place.

Examples with and without subdivide:

output_cm5gxy

output_hrxole

This could likely be improved further, either with more precise (fractional segment) subdivide, and/or by inverting the pattern so that instead of creating multiple `LineLabel` instances per feature, we have one `LineLabel` instance that can emit multiple labels. But this approach was a straightforward and significant near-term improvement, and fits well with the current architecture (we already generate multiple label instances for `Multi` geometries). We could also adjust the heuristic to place more labels, but in my testing I found that label repeat distance was usually the limiting factor (we could adjust that default repeat value to increase density if desired).

@karimnaaji and I discussed this, some more investigation should be done on the ES side. He thought the same issues might apply, but from a quick glance I am not seeing the same magnitude of label sparsity on device.

### DIFF
--- a/src/styles/text/text.js
+++ b/src/styles/text/text.js
@@ -418,7 +418,7 @@ Object.assign(TextStyle, {
 
     // Build one or more labels for a line geometry
     buildLineLabels (size, line, options, labels) {
-        let subdiv = Math.min(options.subdiv, line.length);
+        let subdiv = Math.min(options.subdiv, line.length - 1);
         if (subdiv > 1) {
             // Create multiple labels for line, with each allotted a range of segments
             // in which it will attempt to place
--- a/src/styles/text/text.js
+++ b/src/styles/text/text.js
@@ -380,6 +380,9 @@ Object.assign(TextStyle, {
         }
         layout.repeat_group += '/' + text;
 
+        // Max number of subdivisions to try
+        layout.subdiv = tile.overzoom2;
+
         return layout;
     },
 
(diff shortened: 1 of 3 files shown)


bcamper:
Targeting this for 0.6.2 release.

bcamper:
@karimnaaji can you take a look please?

karimnaaji:
Yep, looking at this in details.

karimnaaji:
Logic looks good to me! It's a nice way to handle this. I will add an issue to track this on -es. 
A thing I noticed is that when `options.subdiv == line.length` the first label would try to iterate on the entire line (with `segment_index = 0`, `segment_max = lines.length`), and next labels only on segments of length 1; shouldn't we let all of the try the entire line in that case?

bcamper:
@karimnaaji good catch! I'll check it out.

src/styles/text/text.js
```diff
@@ -416,6 +414,26 @@ Object.assign(TextStyle, {
         }
 
         return labels;
+    },
+
+    // Build one or more labels for a line geometry
+    buildLineLabels (size, line, options, labels) {
+        let subdiv = Math.min(options.subdiv, line.length);
```
bcamper: @karimnaaji I think the issue you found is corrected just by changing this to `Math.min(options.subdiv, line.length - 1);`. That avoids a segment with `0` for both start and end, and keeps them evenly spaced. (I keep forgetting that we need to limit to line *segments*, not *points*, hence the `- 1`.) I think it's more straightforward to keep the different label placements to their own separate segment ranges (rather than letting multiple labels try the whole line range, which can only be less efficient since multiple labels would try the same segment, and only one can win).

karimnaaji: Yep that fixes the issue. I haven't run it but the code path seems ok for that case now.

>I think it's more straightforward to keep the different label placements to their own separate segment ranges

That's true it makes sense, and I feel like this would also help to keep the segment spacing more homogeneous.

Issue #253, opened 2016-02-26
Bundle multi-file scenes and resources


bcamper:
Bundle multiple scene files and related assets into a single file for easier distribution and less network overhead. Becomes more important as we add scene "include" functionality, increasing the number of network requests to load a scene.
- Multi-document YAML
  - Meta info about "file system" in first document
  - Custom logic to resolve paths, etc.
  - Doesn't support images/binary assets (without inefficient methods like base64 images)
- JSON wrapper with meta info
  - Individual scenes stored as YAML or JSON (just text, either works, and technically YAML is a superset of JSON anyway)
  - Stringified JSON scenes are smaller than YAML due to whitespace removal
  - Doesn't support images/binary assets as above
- Binary format
  - Protobuf, flatbuffers, etc.
  - Supports any asset type, and could reduce payload through custom data structures
  - Significant downside of needing to design format, encoder/decoder, less accessible (not human openable, debuggable, etc.)
- Zip archive
  - Natively preserves file paths, supports images/binary assets, easily readable by machines and openable by humans
  - Reasonable size improvements (compressed text), though we're already gzipping on server
  - Possibly less custom logic to resolve paths, etc. (still necessary, but can go through zip reader API vs. designing our own metata/format)
  - https://github.com/Stuk/jszip - large new depedency (75k minified), may increase value of conditionally loaded modules, e.g. only load TopoJSON, MVT, Zip, etc. libraries/modules as needed (currently the first two are always loaded with the library, regardless of whether they are needed for the scene's data sources. See #255.

bcamper:
cc @blair1618 initial notes for now

bcamper:
also cc @louh for future Play ramifications

blair1618:
Have you tried to emit the scene file in YAML using "flow" notation? I imagine it would be the same or smaller than JSON since it has the same basic structural syntax, but doesn't need quotation marks.

bcamper:
@blair1618 I have not, I'll see if js-yaml supports that option.

nvkelso:
I think this got solved by @patriciogonzalezvivo's Bundler work?

- https://github.com/tangrams/bundler

bcamper:
Yep - should probably have been closed with addition of zip bundles back in v0.9.0. The ES team has considered a pre-parsed binary version of the scene too, but with their recent huge improvements to YAML/scene parsing speed, that may no longer be necessary (or might be limited to an internal caching format). Closing this one. cc @matteblair

[...]

PR #278, opened 2016-03-18
Scene imports

bcamper:
This PR introduces the ability to compose multiple scene files, with a new top-level `import` block. Each scene file can import one or more additional scene files (which can then in turn recursively import others), e.g. to import a file with a POI texture defined:

`import: pois.yaml`

Or more than one:

`import: [pois.yaml, roads.yaml, landuse.yaml]`

The `import` statement works by **deep-merging** each imported scene into the current one. The "child" (imported) file is merged first, with the "parent" file merged after, overwriting any properties shared with the child. 

Use Cases
===
This is intended to be a generic approach to composition that supports several patterns:

- **Base Map + Overlay**: for users who want to display an existing basemap (provided by Mapzen or others), and then add their own data sources + styles on top of that. Probably the most common use case.
- **Themes / Customization**: several designs share large portions of their underlying filter and style rules (see Cinnabar, Refill, Zinc, etc.). With the upcoming ability to create [global properties](https://github.com/tangrams/tangram/pull/263) that can stand in for common values, these styles can be re-factored to derive from a common "skeleton" file (e.g. Skin & Bones), with a smaller number of higher-level properties configured for different color palettes, label visibility presets, etc. (by overriding the imported properties with new values). Provides for moderate style customization (by hand or via an UI) without commitment to writing or altering complex/low-level scene syntax.
- **File Management**: for users designing complex styles such as complete basemaps, splitting a large scene into several files will ease the authoring and maintenance burden. Files could be grouped by styling theme (landuse, roads, POIs, etc.), by function (all custom `styles` in one file, all `sources` in another, etc.), or by any other combination. Here is an [incomplete test branch](https://github.com/tangrams/eraser-map/compare/imports) splitting the Eraser Map scene into separate files.
- **Style/Shader/Sprite Library**: we can make several components of our existing styles reusable by extracting them into separate files, either on their own or packaged into "libraries" of functionality to be used as building blocks. This could include complete rendering styles (which can be used as-is or `mix`'ed with others), pre-defined POI sprite textures + accompanying style rules), or individual shader functions for patterns, procedural textures, etc. (a collection of importable shader components has already been started [here](https://github.com/tangrams/tangram-sandbox/tree/gh-pages/modules)).

Caveats
======
This is an intentionally simplistic approach, to enable several known use cases with minimal implementation logic, without attempting to anticipate pitfalls and prematurely optimize or design around them. The generic merge behavior enables a wide range of uses, while no doubt also making it easy to shoot oneself in the foot with deeply-nested or unexpected overwrites. We will need to watch for pitfalls as they emerge, and could consider more nuanced behavior if deemed more intuitive.

**Objects vs. Arrays**
One important aspect of the merge behavior is that it applies to all *YAML maps/JS objects*, but **not** to *YAML sequences/JS arrays*. While key/value objects merge (with new keys inserted, and existing keys overwriting the previous value), arrays are treated as scalar values that entirely overwrite the previous value (rather than merging the array contents).

For typical cases such as colors, this is the natural behavior:

**`child.yaml`**:
```
layers:
   roads:
      draw:
         lines:
            color: [0.7, 0.3, 0.8]
```

**`parent.yaml`**:
```
import: child.yaml
layers:
   roads:
      draw:
         lines:
            color: [0.3, 0.3, 0.3] # overwrites previous color array with new color array
```

This doesn't currently pose an issue in Tangram scenes because we only use lists as *property values*, not as part of the scene structure itself. For example, if `layers` were listed as a sequence, this behavior might not be generally applicable (one might expect the layers to be concatenated instead). This is an example that we might need to revisit with more context-specific logic as the scene syntax evolves. The goal will be to do the most intuitive thing, with the fewest number of special cases.

**Import Syntax**
To start, the `import` syntax only allows entire scene files to be imported wholesale. One could also imagine extending the syntax to provide further control, such as specifying top-level sections to merge (e.g. merge `styles` but not `layers`), or optionally namespacing imported property names to avoid name collisions.

**Network Requests**
Because all imports currently happen on the client, additional network requests are introduced, and the initial scene render must block on the completion of these. One possible way around this is an offline/command-line "scene compiler" that pre-merges scene files into one for more streamlined distribution. Another option is the adoption of a scene bundle format (such as simple zip files) that allows for all scene files and associated images/resources to be distributed as a single file while preserving their original structure (see #253).

Merge Example
===

A 3-level example:

**`parent.yaml`**:
```
import: child.yaml

layers:
   roads:
      visible: false # turn off road visibility
```

**`child.yaml`**:
```
import: grandchild.yaml

layers:
   roads:
      draw:
         lines:
            color: red # override color to red (overrides color from grandchild.yaml)
         points: # add points to the map (adds new draw group not in grandchild.yaml)
            color: blue
            size: 8px
            order: 10
```

**`grandchild.yaml`**:
```
layers:
   roads:
      draw:
         lines: # default road style
            color: grey
            width: 4px
            order: function() { return feature.sort_key }
```

Final merged scene:
```
layers:
   roads:
      visible: false # from parent.yaml
      draw:
         lines:
            color: red # from child.yaml
            width: 4px
            order: function() { return feature.sort_key }
         points: # from child.yaml
            color: blue
            size: 8px
            order: 10
```

Deprecations
===
The more generic import functionality also allows us to deprecate two earlier attempts at modularity:

- Remote `styles` imported via [`url` keyword](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/styles.md#url) (see [example](https://github.com/tangrams/tangram/pull/278/files#diff-7f549883b996e161107e531e5a5db1cbL110)).
- Remote shader `blocks`, referenced by URL (see [example](https://github.com/tangrams/tangram/pull/278/files#diff-7f549883b996e161107e531e5a5db1cbL60)).

This streamlines the syntax and means we can drop these from the list of parity features for ES.


bcamper:
cc @blair1618

blair1618:
This looks nice!

The only potential issue I can think of is the way that filters will merge. 

Let's merge `child.yaml`:
```yaml
layers:
  roads:
    data: { source: osm }
    filter: { not: { kind: 'rail' } }
```
into `parent.yaml`:
```yaml
import: 'child.yaml'
layers:
  roads:
    filter: { kind: 'rail' }
```
As I understand it, this will result in `result.yaml`:
```yaml
layers:
  roads:
    data: { source: osm }
    filter:
      not: { kind: 'rail' }
      kind: 'rail'
```
Clearly, this isn't going to give us much data to work with. It's a contrived example, but the problem I see here is that there's no good way to make a filter *less restrictive* with this kind of override. Maybe you could put `not: {}` in the parent? Oddly, I think you could put `filter: [ { kind: 'rail' } ]` in the parent and get the 'rails' features to pass. 

This just seems to have a lot of confusing consequences. Would it make more sense to treat an entire filter as an atomic unit for this kind of merging?

nvkelso:
Or add language so one layer can override another?

bcamper:
Discussed more IRL with @blair1618. It's definitely good to be documenting and understanding these cases. That said I'm still hesitant to implement any context-sensitive merge behavior until we have real-world examples of common problems. I think it's likely we will end up adding more specific rules eventually, but I'd feel more comfortable when I know it's a pain point, and not a premature optimization (or worse, has undesirable unintentional consequences).

Something I failed to document is that setting a property to `null` will overwrite the previous value, effectively allowing you to "delete" (or clear) a property. In the case above, you could use this to do `filter: { kind: 'rail', not: null }`. It certainly requires a precise understanding of the structures you are merging, but I think that is more generally true of *any* merge operation that modifies existing data. The safest merges are those that are additive: including a new data source, top-level layer, etc. (even these can be subject to namespace collisions).

My preference for now is to keep generic behavior, and to document and provide guidance. We should probably be encouraging people to use this for additive purposes, and to consider merging of filters, draw groups, etc. to be an advanced use case (supported but proceed with caution).

bcamper:
I'm going to go ahead and merge this to start lining up the 0.7 release. Continued comments are welcome though!

louh:
>  One possible way around this is an offline/command-line "scene compiler" that pre-merges scene files into one for more streamlined distribution.

I quite like this idea. Those of us who use front-end asset built tools will find this pretty useful, and we can make a Node.js module with a CLI pretty easily. 

In Tangram Play world, we would also be able to import this same module to allow someone to export a single "compiled" file from their multi-file "workspace."

bcamper:
@louh I do think there is some use for a scene bundler, but have also started to lean towards something like ZIP files, because:

- Compiling a scene is necessarily a "destructive" operation. When I tried outputting YAML for merged scene objects, the machine-outputted YAML is a lot more verbose and not as easily readable (this may be controllable through options), but more importantly, you generally can't break the single file back down into its original component files, which limits its usefulness for an authoring environment like Play (without something like source maps for Tangram YAML ;)  Zips can be decoded and the original folder and file structure remains intact.
- Zip files are a super old/"dumb" technology that can be created by almost any computer, and doesn't require nearly as much in the way of custom Tangram logic (though if we wanted an option to request remote hosted resources and bundle them, to remove all external dependencies, we would still need to do that).

PR #279, opened 2016-03-22
Selection event handlers

bcamper:
This PR simplifies the process of setting up Tangram feature selection callbacks (such as on hover or click), removing the need for boilerplate code for DOM events by tying into Leaflet's existing event handlers.

A new concept of *selection event handlers* has been added to Tangram, currently with two events supported:
- `hover`
- `click`

Selection event handlers will be passed the same `selection` object returned by direct calls to `scene.getFeatureAt()` (which is still supported and available for custom input/event needs).

Two additional properties have also been added to the returned `selection` object:
- `pixel`: the XY location within the map container where the event occurred, in the form `{ x, y }` (this wasn't previously necessary since the user already needed to pass it explicitly to `getFeatureAt(pixel)`)
- `leaflet_event`: the Leaflet event that triggered the selection is also passed through (it contains information that may be useful for deeper or existing Leaflet integrations)

Handlers for these events can be configured in two ways:

**When creating the Leaflet layer:**

An `events` object can be passed with other leaflet layer options. `hover` and/or `click` properties can be set to a callback function:

```
var layer = Tangram.leafletLayer({
   scene: 'scene.yaml',
   events: {
      hover: function(selection) { console.log('Hover!', selection); },
      click: function(selection) { console.log('Click!', selection); }
   }
};
```

**Updated after Leaflet layer creation:**

Selection events can be added, changed, or removed after layer creation with a call to `layer.setSelectionEvents(events)`. It takes the same `events` object as above:

```
layer.setSelectionEvents({
   hover: onTangramHover,
   click: onTangramClick
});
```

Or, to remove an event, `layer.setSelectionEvents({ click: null });`.

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -408,14 +408,6 @@ Enjoy!
     selection_info.style.display = 'block';
 
     function onFeatureHover (selection) {
-        // Don't show labels while panning
-        if (scene.panning == true) {
-            if (selection_info.parentNode != null) {
-                selection_info.parentNode.removeChild(selection_info);
-            }
-            return;
-        }
-
         // Show selection info
         var feature = selection.feature;
         if (feature != null) {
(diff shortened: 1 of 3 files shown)


bcamper:
cc @nvkelso @meetar @louh

PR #282, opened 2016-03-31
Raster tile sources w/"attachments"

bcamper:
This replaces and expands upon the initial raster tile support proposed in #247. There's a lot going on here...

Summary
===

- `Raster` data sources to load raster tiles from a URL template
- Built-in `raster` rendering style to render raster tiles, either as a basemap or optionally masked against other polygon geometry
- Style functionality for accessing one or more `Raster` data sources as samplers within shaders
- Built-in support for applying raster samplers as normal maps

Raster Data Sources
===
A new `Raster` data source type has been added, which:
- Loads raster tiles from a URL template and provides them as texture samplers within rendering `styles` (when configured to do so).
- Can be used directly as a geometry data source (e.g. as the `source` in the `data` block of a layer), in which case it synthesizes quads that fill the entire tile area. This is suitable for traditional raster tile rendering, where the rasters cover the whole screen area (usually as a base layer).
- Can also be "attached" to other data sources, for more advanced cases where the raster tiles are mixed with other geometry or even other raster tiles. Examples:
  - Drawing raster tiles that are "masked"/clipped to arbitrary polygons, such as only drawing terrain rasters over landuse polygons.
  - Blending two raster tiles together in a fragment shader.
  - Modifying geometry based on a raster tile, such as extruding heights according to a heightmap / elevation tile.
- Parameters:
  - Requires a `url` tile template, in the same manner as other tiled sources.
  - Optionally allows the GL texture `filtering` mode to be set, with the same allowed values (`mipmap`, `linear`, `nearest`) as for other texture definitions in the scene file. Default is `mipmap` for power-of-2 images, and `linear` for non-power-of-2 images (raster tiles are generally power-of-2, though other sizes would be scaled to fit the tile square). Setting `filtering: nearest` allows for the raster tiles to be pixelated when scaled past their `max_zoom`.
  - Optionally defines a `max_zoom` beyond which tiles are over-zoomed. As with other data sources, defaults to `18`.
- Basic raster source example:
```
sources:
  stamen-terrain:
    type: Raster
```

Raster Samplers
===
Each data source defines a numbered list of zero or more raster samplers that (when enabled by the style) are provided for use in shader blocks, along with GLSL convenience functions for accessing them. 
- By default, a `Raster` source provides a single raster sampler (itself), while a non-`Raster` source does not provide any raster samplers.
- Raster sources can be attached to other data sources with the `rasters` property, which takes an array of raster source names.
- In the following example, both sources provide a single raster sampler (Mapzen elevation tiles):
```
  sources:
    # Define Mapzen elevation tiles as a raster data source
    mapzen-elevation:
      type: Raster
      url: https://terrain-preview.mapzen.com/terrarium/{z}/{x}/{y}.png
    # Define Mapzen vector tiles as a data source, and attach the `mapzen-elevation` raster source
    mapzen-osm:
      type: TopoJSON
      url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
      rasters: [mapzen-elevation] # provide elevation tiles as a raster sampler when rendering 
```
- When a `Raster` source itself has additional raster sources set in the `rasters` property, the "parent" source will be the first raster sampler, and those from `rasters` will be added after (essentially it is as if the parent source was inserted as the first item in the `rasters` array).
```
  sources:
    # Provides a single raster source (rasterA)
    rasterA:
      type: Raster
      ...
    # Provides two raster source, rasterB as the first (index 0), rasterA as the second (index 1)
    rasterB:
      type: Raster
      rasters: [rasterA]
```

Built-In Raster Style
===
A built-in `raster` style is provided for typical rendering cases. The style applies the raster tile pixel color value to its geometry.

**Raster Basemap**

When used with a `Raster` source, this results in traditional raster tile rendering. An example using the Stamen terrain tiles as a basemap:
```
   layers:
       terrain:
           data: { source: stamen-terrain }
           draw:
               raster:
                   order: 0 # draw on bottom
```

tangram-wed mar 30 2016 16-51-48 gmt-0400 edt

In this example, `order` determines the layer order as with other styles. The `color` defaults to white and any other value set by the draw group will *tint* the raster texture.

**Filter Effects**

For example, setting `color: [0.5, 0.5, 0.5]` will darken the underlying tiles by 50%:

tangram-wed mar 30 2016 16-56-31 gmt-0400 edt

The `raster` style is derived from the `polygons` rendering style, and provides the same shader blocks. This allows for custom raster styles to be defined in `styles` with `base: raster`. For example, a style that converts raster tiles to greyscale:
```
styles:
    greyscale:
        base: raster
        shaders:
            blocks:
                filter: |
                    float luma = dot(color.rgb, vec3(0.299, 0.587, 0.114));
                    color.rgb = vec3(luma);
```

tangram-wed mar 30 2016 16-59-03 gmt-0400 edt

**Raster/Geometry Masking**

The `raster` style can also be applied to a non-`Raster` source that has a raster sampler attached. In this case, the Stamen terrain is attached to and masked against the OSM landuse polygons:

```
    sources:
        stamen-terrain:
            type: Raster
        mapzen-osm:
            type: TopoJSON
            url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
            rasters: [stamen-terrain] # attach stamen terrain

   layers:
       terrain:
           data: { source: mapzen-osm, layer: landuse } # render landuse layer from vector data source
           draw:
               raster:
                   order: 0 # draw on bottom
```

tangram-wed mar 30 2016 17-07-48 gmt-0400 edt

This technique can be used for combining shaded relief with landcover classifications, for example here is a base terrain layer in grey, with landuse polygons colorized with green:

tangram-fri apr 01 2016 12-35-12 gmt-0400 edt

Advanced Raster Styles
===
The above cases apply a single raster source to geometry (with optional color filter effects). More advanced effects can be achieved with custom styles derived from the `polygons` style.

By default, raster samplers are not included in `polygons`-derived styles, even if the data source provides them. This is because styles *with* raster samplers will not render until the attached raster textures load. To avoid unnecessary blocking / delay, they are disabled by default.

**Enabling Raster Samplers**

Styles can enable raster samplers with a new `raster` parameter, which can have the following values:
- `color`: Applies the value of the raster texture as the `color` in the fragment shader. This is the most common case, and is set as the default by the `raster` rendering style.
- `normal`: To support terrain shading, there is also built-in support for applying a raster source as a normal map.
- `custom`: This value is for cases where you want access to raster samplers, but the data is not formatted for direct use as a color or normal. Mapzen's RGB-packed elevation tiles are an example; the raw data must be decoded and re-interpreted for usable results, and is not intended for display.

**Normal Maps**

For example, a normal map lit by standard Tangram lights can be created as:

```
sources:
    terrain-normals:
        type: Raster
        url: https://terrain-preview.mapzen.com/normal/{z}/{x}/{y}.png
    mapzen-osm:
        type: TopoJSON
        url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
        rasters: [terrain-normals]

styles:
    normals:
        base: polygons
        raster: normal

lights: ...

layers:
    earth:
        data: { source: mapzen-osm }
        draw:
            normals:
                order: 0
```

tangram-wed mar 30 2016 21-39-44 gmt-0400 edt

**Direct Sampler Access**

When the style has `raster: custom`, the shaders can directly sample the raster for custom effects. The following (public) GLSL uniforms and helper functions are provided:

- `sampleRaster(int N)`: Samples the Nth raster source at the current pixel position (similar to `texture2D()` but adjusts UV offset and scale to account for any raster tile overzooming).
- `sampleRasterAtPixel(int N, vec2 pixel)`: Samples the Nth raster source at the pixel (not UV) position in `pixel`, e.g. `sampleRasterAtPixel(0, vec2(100, 50))` to sample at pixel `(100, 50)`.
- `vec2 currentRasterPixel(int N)`: Returns the current pixel position for the Nth raster source. Useful along with the above for sampling nearby pixels (e.g. to derive a normal map, perform a convolution kernel, etc).
- `vec2 rasterPixelSize(int N)`: A uniform array providing the pixel dimensions of the Nth raster source.

The following example uses the `sampleRaster()` method to show how Mapzen elevation tiles can be unpacked for color display:

```
styles:
    elevation:
        base: polygons
        raster: custom
        shaders: 
            blocks:
                global: |
                    // Unpack RGB elevation
                    float unpack(vec4 h) {
                        return (h.r * 1. + h.g / 256. + h.b / 65536.);
                    }
                color: |
                    color.rgb = vec3(unpack(sampleRaster(0)));
                    color.rgb = (color.rgb - .5) * 20.; // re-scale to a visible range for contrast
```

tangram-wed mar 30 2016 21-59-05 gmt-0400 edt

**Multiple Raster Samplers**

Here is another example that shows the use of two raster samplers in the same rendering style. First, sources are defined for Stamen's Watercolor and Toner tiles. To make them both available in the same rendering `style`, the watercolor source is attached as the second sampler for the toner source.

A simple `style` that does a 50/50 blend of the two sampler is then defined. Finally, the toner source is drawn with the blend style.

```
sources:
    stamen-watercolor:
        type: Raster
        url: http://tile.stamen.com/watercolor/{z}/{x}/{y}.jpg
    stamen-toner:
        type: Raster
        url: http://tile.stamen.com/toner/{z}/{x}/{y}.png
        rasters: [stamen-watercolor] # add watercolor sampler alongside toner

styles:
    # blend two raster samplers together
    blend-rasters:
        base: polygons
        raster: color
        shaders:
            blocks:
                filter: |
                    color = (sampleRaster(0) + sampleRaster(1)) * 0.5;

layers:
    earth:
        data: { source: stamen-toner }
        draw:
            blend-rasters:
                order: 0
```

tangram-wed mar 30 2016 17-20-15 gmt-0400 edt

Implementation Notes
===
- **Raster texture lifecycle:** Any given raster tile texture may be referenced by multiple tiles at once, either as a proxy tile, or when the same raster source is provided as a sampler to more than one rendering style.
  - This complicated the previous "single owner" relationship, in which the tile was responsible for destroying its textures when it was itself destroyed.
  - To support this, a referenced counting ["retain/release" pattern](https://github.com/tangrams/tangram/pull/282/files#diff-fc1e874176caee3f22f13cf9d225e067) was adopted. This is still in transition; the tile textures use retain/release, but some other textures still directly manage their lifecycle. It may be advisable to unify these further.
- **Raster tile overzooming:** A raster source may have a lower `max_zoom` than a data source it is attached to. In these cases, the lower-level raster tile itself needs to be "overzoomed" onto the higher level source being drawn.
  - This complicated texture sampling, because each tile may need to apply additional scaling and offsetting to UVs, to properly position itself within the correct subset of the larger overzoomed raster tile. This additional info is provided as the `u_raster_sizes[]`and `u_raster_offsets[]` uniforms, which are [calculated by the style](https://github.com/tangrams/tangram/pull/282/files#diff-e90eeb4f0c5a50cd03b669245e7e0724R501) at tile build time.
  - A set of [GLSL helpers](https://github.com/tangrams/tangram/pull/282/files#diff-e899aa78bc1880eadf7507349dace26c) provides macros for applying these adjustments in the shader. Note that these were written as *macros* instead of *functions*, because of the requirement that array indices be constant expressions; in this case, the behavior is simple enough that implementing as a macro did not pose a challenge.
  - Additionally, in cases where the geometry tile is itself overzoomed (e.g. raster being applied to a geometry source with `max_zoom: 16`, rendered at z18), the raster texture needs to be sampled with UV coordinates corresponding to the tile's *coordinate zoom* (not its *style zoom*). This necessitated an additional (currently private) accessor for the [model position within tile coordinate space](https://github.com/tangrams/tangram/pull/282/files#diff-d5e16a8f498be346a782a77255f01b60R17). When raster samplers are enabled by the style, this value is [added as as varying](https://github.com/tangrams/tangram/pull/282/files#diff-cf729d8391e34e162c4862e537f0a6b3R70) for UV calculation.
  - In the converse case, where the *geometry* data source has a lower `max_zoom` than a raster source, the raster will get capped at the geometry's `max_zoom`, meaning it will be prematurely scaled up. This is a known limitation, because there is currently no feasible way to provide *multiple* raster tile textures at the same time within the shader (it would likely require a much more costly approach, such as pre-rendering multiple raster tiles into a larger "virtual" raster tile texture on the CPU).
- **Number of raster samplers:** Styles can have zero or more raster samplers available at render-time, but this is dependent on the *data source*, not the *style*. The same rendering style can in fact be used with different data sources, each with varying number of raster samplers, for different layers within the same scene.
  - This complicates the shader source construction at compile-time, because we don't know how many raster samplers will be bound at render-time. This is solved by defining any raster sampler-related uniforms to be sized for the [maximum number of samplers possible](https://github.com/tangrams/tangram/pull/282/files#diff-e90eeb4f0c5a50cd03b669245e7e0724R430) (we know the max, if a data source included attachments for all possible raster sources in the scene).
  - While the raster sampler uniforms *can* hold up to the maximum possible raster samplers, in practice, fewer may be bound at render-time, as determined by the data source being used.
- **Geometry vs. Raster sources:** We don't want to create tile objects for sources that are raster-only. We [check[(https://github.com/tangrams/tangram/pull/282/files#diff-1e8bec308a3762b66b1c71be2d0a398eR831) which sources are referenced in the `source.data` field for all `layers`, and only generate tile objects for those.


bcamper:
Updated with some implementation notes that are likely applicable for ES as well.

@blair1618 please review!!

bcamper:
also cc @meetar @patriciogonzalezvivo

meetar:
:+1::+1:

I'd drop this straight into the docs (maybe with a few minor revisions)

bcamper:
@patriciogonzalezvivo I just realized that right now we're only injecting raster samplers [in the fragment shader](https://github.com/tangrams/tangram/pull/282/files#diff-e899aa78bc1880eadf7507349dace26cR3). I realize you'll need them in the vertex shader to do things like terrain geometry extrusion. I think this should probably be an optional flag on the style, like `raster_vertex: true`, since texture samplers in vertex shaders are actually not guaranteed by the spec, have lower limits, and also contribute to a max sampler limit across both vertex + fragment shaders. @blair1618 what do you think about that?

patriciogonzalezvivo:
:+1:

blair1618:
After a few minutes browsing a GL hardware database, I found that GPUs as new as the Mali-400 (in the Galaxy S4) have a GL_MAX_VERTEX_TEXTURE_IMAGE_UNITS of 0. I don't feel comfortable adding a built-in feature that depends on vertex texture samplers. 

GLSL extensions are a similar feature that we *do* allow, even though they aren't universally supported. But:
 1. A user can check for extension availability and code a fallback
 2. Extensions are unlikely to be used by a non-expert, you need very specific GLSL knowledge to even activate them 

I believe the terrain extrusion demo would still be possible without a built-in feature for this. It may require some knowledge of tangram's inner workings, but I don't think we can claim that this demo is achievable by someone who hasn't read some of tangram's source.

bcamper:
Summary of IRL discussion:

It might an OK compromise to require heavy custom shader code to add the raster samplers to the vertex shader, but, I would like to better understand why we think this is different from other hardware limitations, e.g. it is also possible to create styles that won't work on a subset of devices by exceeding other per-device limits. For example, since all custom shader uniforms are injected in both vertex and fragment shaders, it's easy to create this same situation just by trying to access a texture sampler in a vertex shader block on a device that doesn't support it.

In WebGL, > 98% of devices support at least 4 vertex shader texture samplers (according to webglstats.com). It would be nice to have a better idea of support % for ES devices, but I realize that info may be harder to come by.

It seems like a reasonable policy to not have any **default** behavior that isn't guaranteed by the spec (hence not defaulting raster samplers on for vertex shaders), but I am not totally convinced that doesn't mean we shouldn't have options that only work on a subset of devices (especially since we already do!). In the latter case, I think the more important thing (which we're currently probably not doing so well at) is to have more robust error checking and reporting so that the application can at least *know* that the requested style won't work, and inform the user or take other action (rather than just blowing up, which is probably what happens now).

blair1618:
At the very end of the "Raster Samplers" section there's an example that reads `rasters: [A]` where it seems like it should read `rasters: [rasterA]`. Is that a typo?

bcamper:
@blair1618 yes thank you!

nvkelso:
Super cool! 

1. You showed how we can mask a raster with vector, is the reverse possible? Like only show vector data when above 1,000 meters elevation.
2. Is it possible to take the color from a vector layer styling (like `landuse`), and merge that with the hillshading from the raster (to get green hillshaded parks, for instance).

blair1618:
Lookin' good. 

One tiny detail: our built-in styles are now `points`, `lines`, `polygons`, and `raster` (not `rasters`, as I would've guessed from the pattern).

bcamper:
@blair1618 you forgot `text` ;)  I see the point, but (and this may just be subjective) the word "raster" feels more like a single raster layer to me, you don't see the plural used much and it makes me think of multiple raster layers stacked. Can't say I have a really strong opinion though...

blair1618:
Oh I did forget about `text`. I guess the plural is not obviously more clear, so I'm fine with this.

bcamper:
@nvkelso 

>You showed how we can mask a raster with vector, is the reverse possible? Like only show vector data when above 1,000 meters elevation.

I had never thought of that, so I tried :)  Turns out it is, though it may not be advisable performance-wise since it involved a `discard` operation in the fragment shader (I did not benchmark though). Depending on the goals, it could make more sense to precompute those decisions in the tile data. Here's what it looks like though:

terrain-height-filter

>Is it possible to take the color from a vector layer styling (like landuse), and merge that with the hillshading from the raster (to get green hillshaded parks, for instance).

Yes! This was a core intended use case. The color assigned in the layer is available in the shader along with the raster styling, so it should be easy for us to mix these. Related, @meetar and I settled on the image-based environment maps as probably the best mix of ease of design + flexibility, vs. using built-in Tangram lights, in which case we could either have different environment maps for different polygon types, or, a single environment map which is re-colored based on polygon type. The latter is what's going on here (earth polygons drawn as grey terrain, landuse polys drawn with same environment map colored green):

tangram-fri apr 01 2016 12-35-12 gmt-0400 edt

nvkelso:
:+1:  @sensescape for park magic above :)

bcamper:
@nvkelso @sensescape to be clear, I don't think that is going to be workable for a production style.

nvkelso:
The generic + park based environment maps?

bcamper:
@nvkelso sorry, tried to clarify after posting. We **should** do the bottom image, with differently colored landuse polygons. That was a big driving use case for this PR (and the main reason it ended up more complex to implement than a traditional raster tile layer that covered everything).

The one I was steering us away from was the animated height filter thing, which should be considered more of a prototype.

nvkelso:
:)

bcamper:
@blair1618: per discussion, I added a `rasterPixelSize(raster_index)` macro/function, to provide a more consistent interface as an alternative to accessing the `u_raster_sizes` uniform directly. Make sense? https://github.com/tangrams/tangram/pull/282/commits/0bcb36919e2e553c47a179fe5c07bcd9e27e4491

blair1618:
Sounds good!

bcamper:
Merging as-is, will revisit raster samplers in vertex shaders later. I confirmed (I think, just tried a quick demo :) that it is technically possible, though pretty messy, to expose rasters in vertex shaders by copying the sampler uniform and macros definitions into a custom style's `global` shader block. This may be enough for us to port some of @patriciogonzalezvivo pre-existing terrain demos. However, I still think it is worth revisiting if more direct support should be included (and how this case differs from other hardware limitations).

tallytalwar:
Very well explained PR description here @bcamper.

src/styles/polygons/polygons_fragment.glsl
```diff
@@ -35,9 +40,14 @@ void main (void) {
     vec4 color = v_color;
     vec3 normal = TANGRAM_NORMAL;
 
-    // Normal material texture (fragment lighting only)
-    #if defined(TANGRAM_LIGHTING_FRAGMENT) && defined(TANGRAM_MATERIAL_NORMAL_TEXTURE)
-        calculateNormal(normal);
```
karimnaaji: What is expected to happen now when a material is applied as a normal map to regular geometry? 

According to this it seems that regular normal maps (not raster normal maps but user defined uv/planar/triplanar normal maps) are not supported anymore.

What about:
```c
// First, get raster normal if any
#ifdef TANGRAM_RASTER_TEXTURE_NORMAL
    normal = normalize(sampleRaster(0).rgb * 2. - 1.);
#endif

// Second, alter normal with normal map texture if any
#if defined(TANGRAM_LIGHTING_FRAGMENT) && defined(TANGRAM_MATERIAL_NORMAL_TEXTURE)
    calculateNormal(normal);
#endif
```

Or, raster normals/material normal maps can be mutually exclusive:
```c
#ifdef TANGRAM_RASTER_TEXTURE_NORMAL
    normal = normalize(sampleRaster(0).rgb * 2. - 1.);
#else
    #if defined(TANGRAM_LIGHTING_FRAGMENT) && defined(TANGRAM_MATERIAL_NORMAL_TEXTURE)
        calculateNormal(normal);
    #endif
#endif
```

bcamper: Ack, that is a mistake if we lost regular material normal maps! Let me look through history... but yes I think that allowing them to be compound makes sense.

meetar: Yep - just checked
http://tangrams.github.io/tangram-docs/?material/ex-normals.yaml#19/40.70533/-74.00975
locally using the 0.7 release, and the normal map isn't being applied.

bcamper: @karimnaaji I followed your suggestion and this should be fixed here (verified in Tangram Play): https://github.com/tangrams/tangram/commit/9113f3aab36a63af7ea9f6b6a18efc1424a17b0d

Maybe someone can double-check this and then we can cut a new release. Thanks for finding!

meetar: Confirmed with a local build!

Issue #281, opened 2016-03-28
Support custom fonts


bcamper:
We currently draw labels with Canvas text, which means we get web/CSS font styling "for free". However, the Canvas text API is not aware of and does not wait for custom web fonts (non-system fonts) to load before drawing. This means web fonts can't be used reliably (often the first tile or two to render without the font but cause the browser to begin the font loading process; later tiles draw with the font once it has loaded).

The [CSS Font Loading Module Level 3](https://drafts.csswg.org/css-font-loading/) provides an API for detecting font loads, but is only [natively supported](http://caniuse.com/#search=FontFaceSet) by Chrome and FF.

[`FontFaceObserver`](https://github.com/bramstein/fontfaceobserver) provides a partial polyfill that can be used to detect font loads across all browsers, and is a promising solution (with one issue noted below).

For full custom font support, we should consider adding a top-level `fonts` section to the scene file, allowing for named font faces to be specified, along with their URLs for loading. If used with TrueType or OpenType fonts, this could hopefully be done in a compatible way across Tangram JS and ES (this could even probably be done with web-specific formats like WOFF(2), which are apparently mostly containers for TT/OT font data). In a web context, additional fonts may also already be loaded via CSS (for example, using [Google Fonts](https://www.google.com/fonts)), which should be considered for any syntax to avoid attempts to load a font face more than once.

A first step could be to use `FontFaceObserver` to detect when *any* font face is specified for label rendering. This would allow for CSS-loaded fonts to start working properly, without any additional scene file syntax (just like locally installed fonts do on both web and native). However, `FontFaceObserver` currently times out (rather than immediately detecting) system fonts (e.g. Helvetica etc.) when using native `FontFaceSet` support on Chrome and FF. This adds a 3-second delay to all text rendering (clearly unworkable). Currently investigating solutions for this, which could involve either minor modifications to the behavior (either via fork or PR, possibly using the [`document.fonts.check()`](https://drafts.csswg.org/css-font-loading/#font-face-set-check) method), or running a build with native `FontFaceSet` support disabled (`FontFaceObserver`'s fallback behavior actually does detect system fonts, but is slower and using the native API is always preferable).

bcamper:
Got a WIP test branch working [here](https://github.com/tangrams/tangram/compare/fonts), but also had to make some temporary modifications to make `FontFaceObserver` load in a web worker (since Tangram currently its entire self into a worker, even for code that is only needed on the main thread).

bcamper:
[...]

Issue #252, opened 2016-02-26
More flexible hosting/bundling of library


bcamper:
There are significant limitations in the way Tangram currently loads its web workers. These mainly restrict others' desire to host or bundle the library, rather than loading it from the "official" Mapzen URLs.

Current library design:
- The same code is executed on both main and worker threads. This is because there is so much overlap of core code that needs to be executed by both. Many objects such as `Style` get instantiated in both thread types (albeit with some differences in behavior). There are a handful of switches (using `Utils.isMainThread` and `Utils.isWorkerThread`) to control behavior that differs across threads, usually where some code is applicable to one thread but not the other. Having one library file also simplifies distribution/hosting (rather than having to host and load a separate file built for worker execution).
- To instantiate its workers, Tangram needs a full copy of its own code. It currently accesses its code by looking for the `<script>` tag that it was loaded from. It restricts this search to known/expected library names: either `tangram.min.js` or `tangram.debug.js`.

Limitations of this approach:
- Tangram **must** be loaded from a file called either `tangram.min.js` or `tangram.debug.js`. For instance, copying it to a file called `tangram.js` will cause it to fail (because it cannot find the expected `<script>` tag.
- Tangram must be able to access its code in isolation from other code, which means it cannot be bundled with other JS dependencies (e.g. via a minifier or even simple `cat` bundle).
- It's possible for frameworks to modify the DOM _before_ Tangram searches for its `<script>` tag. For example, there was an apparent case where Webpack was removing `<script>` tags before Tangram executed, causing it to fail (but note this is speculative and would need to be verified with current versions, nonetheless the possibility remains).

Possible solutions:
1. Without changing current functionality, Tangram could have an option to provide an alternate URL from which its worker code should be loaded (instead of trying to infer that URL). Currently this could mean simply pointing back to the same URL that Tangram itself is being loaded from (since the main and worker thread bundles are the same); in option 4 below, the worker URL could point to a different bundle. There is actually an existing [internal option](https://github.com/tangrams/tangram/blob/master/src/scene.js#L53) for this which only needs to be exposed in the Leaflet plugin. However, this solution is incomplete, as it allows Tangram to be hosted under different filenames, but _not_ easily bundled with other code.
[...]

PR #284, opened 2016-04-04
Wait for web fonts to load before rendering

bcamper:
Since our text is rendered with Canvas, any CSS font can be specified in our `text.font` block in the scene file. This opens the door to use custom fonts in our styles. But for web fonts (e.g. loaded via CSS, not built-in system fonts), the Canvas draw text API is unaware of a given font face's current load/validity state. This means web fonts can't be used reliably (often the first tile or two to render without the font but cause the browser to begin the font loading process; later tiles draw with the font once it has loaded). See #281.

The [CSS Font Loading Module Level 3](https://drafts.csswg.org/css-font-loading/) provides an API for detecting font loads through the `FontFaceSet` object, but is only [natively supported](http://caniuse.com/#search=FontFaceSet) by Chrome and FF. [`FontFaceObserver`](https://github.com/bramstein/fontfaceobserver) provides a partial polyfill that can be used to detect font loads across all browsers.

**Solution**
This PR provides reliable web font support with the following flow:
- `text`-based styles collect unique font face (family, style, weight) combinations per tile
- When text sizes are requested (`CanvasText.textSizes()`), a new method `CanvasText.loadFonts()` is called to ensure that all fonts have been checked for availability
- `CanvasText.loadFonts()` does the following:
  - For each new font face it hasn't seen yet, it creates a promise via `FontFaceObserver` that will determine the font face's availability (it will resolve if it is available, and reject if it is not)
  - A `Promise.all()` collection is returned for all currently pending font face checks for the tile.
- Actual text rendering is unchanged whether the font face is available or not (it will fallback to a default font face if the requested one is not available). The purpose of checking for font face load status is to provide *consistent* rendering across tiles (e.g. all tiles have the same font face).

An example of POIs rendered with the [Montserrat](https://www.google.com/fonts/specimen/Montserrat) font:

screen shot 2016-04-04 at 1 26 08 am

**Implementation Issues**
The current version of `FontFaceObserver` only recognizes true web fonts in native `FontFaceSet` mode; common system fonts such as Helvetica actually timeout, creating an undesirable 3 second load penalty. This PR currently solves this via a [fork](https://github.com/tangrams/fontfaceobserver/tree/check-before-load) that calls `document.fonts.check()` (which includes system font checks and returns synchronously) before following the polyfill's existing `document.fonts.load()` path: https://github.com/tangrams/fontfaceobserver/commit/5c260094d198b8128c9ff52de0d39bebd0d6d6ac. A question is out to the maintainer as to this method.

Additionally, `FontFaceObserver` would not load within a web worker as-is. The module itself is not needed within the worker (it requires DOM access that is not possible on worker threads), but due to the way that Tangram currently loads its workers (see #252), almost all code is shared the module is still loaded anyway (and was throwing an error). The least invasive way to solve this was to shim `window` and `window.document` in worker threads, to be empty but defined objects (which allows the module to load). This is however a brittle solution that may point to the advantage of smarter main/worker thread package delineation.


louh:
:heart: Montserrat

bcamper:
Superseded by #326.

PR #291, opened 2016-04-12
Watchify + BRFS

dmvaldman:
Not ready to merge yet, but wanted to create a place for discussion. This current PR includes the (rebased onto master) `brfs` branch, in addition to a `watchify` workflow.

I've changed `module.js` to avoid explicitly providing `Tangram` on the `window` object, favoring using browserify's standalone bundling process which creates a UMD build (which implicitly creates the window object). We can also decide whether we want `Tangram` set on the worker instances as well.

To be decided on how this should work in tandem with the makefile. As mentioned in an email, I would suggest moving away from the makefile and towards simple npm scripts, so a developer can run

```
npm run dev // for bundling and watchify
npm run bundle // for bundling, minifying
npm run test // calls your Karma test suite
```

Currently I've avoided this issue by putting the build in a separate script, `build_dev.js`


bcamper:
For the worker, it's useful to have the `debug` items available somewhere, is that still happening but on `self` now ([here](https://github.com/tangrams/tangram/pull/291/files#diff-47426efabbdf4a14eefdc36d3a774fc0R75))?

I used `make` originally because it seemed the most "timeless" / agnostic to the rest of the pipeline, but I'm not tied to it if it's hindering the workflow. Running a watchify process from a make command doesn't inherently bother me, but if you think it would be weird to a typical developer (vs. expected by an npm command), moving towards npm scripts is fine with me (and I suppose we are not likely to move away from npm anytime soon).

For the watch process, what is a good way to handle switching branches? I noticed this was an issue when I was testing it out, I frequently switch branches and it seemed to be getting confused. Likewise, I don't recall seeing any error output, the build was silently failing. Those both seem issues that interfere with everyday use, otherwise the experience was much improved.

Small note on tests, I usually use the combination of `make karma-start` to start the Karma browser, then `make run-tests` several times (so the browser only needs to be setup once, not each time the tests are run). We can account for this with appropriate parallel npm scripts if we're moving to that.

src/module.js
```diff
@@ -59,33 +59,21 @@ var debug = {
...
-
-    window.Tangram = module.exports = {
-        leafletLayer,
-        debug,
-        version: version.string
-    };
-
-}
-
-if (Utils.isWorkerThread) {
```
dmvaldman: We'll need to keep this block for worker to have access to `self.Tangram`

bcamper: I'd like access to the `debug` object, but it doesn't need to be under a `Tangram` property per se.

dmvaldman:
I'm fine with either a makefile or npm scripts. Having a make command like `make bundle-watch` or something works. I have a preference for npm scripts as I think it's more friendly to the JS dev community and they don't come with the dependency burden of gulp/grunt. If we decide on npm scripts we can have both those and make commands and gracefully deprecate one if that's a concern.

Do you mean switching git branches? I'm not sure why there would be a problem. Maybe some things aren't npm installed in other branches?

For the build silently failing. It seems to add syntax error support, you need more middleware, go figure. There is a generic syntax-error module https://www.npmjs.com/package/syntax-error, or we can go with a some middleware choices (which may be a good option anyway because it can refresh the browser automatically on bundle creation):

- https://www.npmjs.com/package/watchify-server
- https://www.npmjs.com/package/watchify-middleware
- https://www.npmjs.com/package/errorify

bcamper:
npm scripts is cool.

Middleware to automate reload sounds interesting (I do think syntax error
feedback is needed or you can end up pretty confused as to why it's not
updating). Have you checked out the budo dev server?

Not sure what I was seeing with git branches, I'll try again when the
rest of this is ready.

Did you want to also remove the shaderSources module?

src/module.js
dmvaldman: So actually, as you originally mentioned, simply exporting Tangram will attach it to the worker, and so code block can be safely deleted.

It's a little confusing to me but I guess the worker loads the tangram source [here](https://github.com/tangrams/tangram/blob/master/src/scene.js#L255-L267)?

bcamper: Yes that's right, the worker is loaded there. Background on how/why/alternatives here: https://github.com/tangrams/tangram/issues/252

dmvaldman:
Seems budo is a good way to go. The launch script would be:

```
budo src/module.js:dist/tangram.debug.js --live -- -t [ babelify --optional runtime ] -t brfs -s Tangram
```

This obviates the need for the node `build_dev.js` script and can simply be run via command line/npm script. Not getting syntax errors to appear, however..

I believe the shaderSources module is still needed unless we want to remove the brfs approach for glslify. Right?

bcamper:
I think you can get this to work with `brfs` the same way you would with `glslify` (both just generate a string), wherever each shader is reference, you would use the `brfs` call instead of referencing `shaderSources`, e.g. 

In https://github.com/tangrams/tangram/blob/brfs-valdman/src/styles/style_manager.js#L30:

`ShaderProgram.addBlock('global', shaderSources['gl/shaders/accessors']);`

becomes

`ShaderProgram.addBlock('global', fs.readFileSync(__dirname + '/shaders/accessors.glsl', 'utf8'));`

dmvaldman:
Sure that works. I can replace those parts of the codebase.

bcamper:
Oh except [here](https://github.com/tangrams/tangram/blob/ac07620a124c0b15716e2b9cb455dad56d437740/src/styles/style.js#L324) they are referenced dynamically, but that could be refactored so each `Style` directly referenced its shader string, instead of the name of a value within `shaderSources`, e.g.:

In https://github.com/tangrams/tangram/blob/ac07620a124c0b15716e2b9cb455dad56d437740/src/styles/polygons/polygons.js#L15:

`vertex_shader_key: 'styles/polygons/polygons_vertex'`

could become something like:

`vertex_shader_base: fs.readFileSync(__dirname + '/shaders/styles/polygons/polygons_vertex.glsl', 'utf8')`

(It may have been prematurely abstracted)

dmvaldman:
Removed the watchify script in favor of an npm script using budo. Now a developer just needs to run (after npm install)

```
npm run watch
```

To have a live-reloading browser and incremental building of tangram.debug.js

bcamper:
:raised_hands:

dmvaldman:
Seems a worker test living on your CI server (/tangram.test-worker.js) is failing

bcamper:
Seems the tests may be failing because the test bundle (`make test-ci`) isn't being built as a UMD module, and we removed the bit in `module.js` that set the `Tangram` symbol globally?

dmvaldman:
Yeah I think that's what's going on. I wasn't able to test locally, so I pushed up new dist files (was going to force revert after). Then I noticed the makefile was creating a special test distribution. It would be good to sit down and get the local testing up and running.

dmvaldman:
Spent all day fiddling with karma configs :-( That little green check mark makes it all worth it though. The last commit changes the testing process slightly. Instead of building two bundles, `tangram.test.js` and `tangram.test-worker.js`, now there is just one build `tangram.test.js` (which is a staging test bundle of Tangram) and the individual test files are loaded by the karma configuration file. This eliminates the redundancy and gives an easier path to moving to simpler npm scripts and away from the makefile.

bcamper:
:boom:

dmvaldman:
What do you think about putting this on a path to merging with master? I'd like to replace the other make commands with npm scripts, and update the brfs calls.

bcamper:
Sounds good! Once that's done I'll give it a whirl to see if I notice anything, otherwise good to go.

package.json
```diff
@@ -64,13 +74,14 @@
   "devDependencies": {
     "babel-runtime": "5.8.29",
     "babelify": "6.4.0",
-    "browserify": "6.1.0",
+    "browserify": "10.0.0",
+    "budo": "^8.2.1",
     "chai": "1.9.2",
     "chai-as-promised": "4.1.1",
-    "derequire": "2.0.3",
```
bcamper: There was a reason we were using `derequire`, but I don't remember what it is. @louh should know (I believe he added it).

makefile
```diff
@@ -1,61 +0,0 @@
...
-	dist/tangram.debug.js
-
-debug: dist/tangram.debug.js
-
-dist/tangram.debug.js: .npm $(shell ./build_deps.sh)
-	node build.js --debug=true --require './src/module.js' --runtime | $(DEREQUIRE) > dist/tangram.debug.js
-
-dist/tangram.min.js: dist/tangram.debug.js
-	$(UGLIFY) dist/tangram.debug.js -c warnings=false -m -o dist/tangram.min.js
-	@gzip dist/tangram.min.js -c | wc -c | awk '{ printf "%.0fk minified+gzipped\n", $$1 / 1024 }'
```
bcamper: Spitting out the gzipped size has kept me conscious of bundle size (vs. proactively checking it periodically), I think it's a nice thing to keep.

louh: Derequire is used to uglify the original require() calls, which remain in
the bundled package. Without it, other browserify-based applications that
require('tangram') will attempt to parse Tangram's require() calls, but
fail because Tangram's dependencies are not present in the parent
application's dependency list.

dmvaldman: I think so too. I was stuck at escaping the double quotes in the printf statement... (enclosing double quotes are introduced by the npm script being a string). Any ideas?

package.json
dmvaldman: I'll look into this after @louh's explanation. I think with the standalone version of Tangram, derequire is no longer necessary. Will test.

dmvaldman: Thanks @louh, you're right, passing through derequire is necessary if Tangram is to be required via browserify. Will fix.

makefile
dmvaldman: Here's the line to be executed, but it needs to be able to be wrapped in double quotes in the npm script, and I couldn't figure out how to escape the double quotes needed for the printf statement
```
gzip dist/tangram.min.js -c | wc -c | awk '{ printf "%.0fk minified+gzipped\n", $1 / 1024 }'
```

dmvaldman:
Besides the gzip command, which I'd still like to include, this branch is ready for testing to incorporate into master! @bcamper

dmvaldman:
To summarize the diff:

* uses `brfs` and inline calls to shader source text
* more efficient build to karma testing suite by bundling tests into the karma config
* uses `budo` for incremental building and a live reloading browser (`npm run watch`)
* migrates from makefile to npm scripts. Common scripts described below:

```
	npm run test // lints, builds tangram.test.js, and runs karma (for use on circle CI)
	npm run test-local // lints, builds, tangram.test.js and runs karma locally (single run)
	npm run karma-start // sets up karma
	npm run karma-run // runs tests after setup
	npm run build // bundles tangram.debug.js and minifies to tangram.min.js
	npm run build-debug // bundles tangram.debug.js
	npm run watch // runs budo for incremental building and live reloading
```

makefile
bcamper: This appears to work:

`gzip dist/tangram.min.js -c | wc -c | awk '{kb=$1/1024; print kb}' OFMT='%.0fk minified+gzipped'`

https://github.com/tangrams/tangram/pull/291/commits/0fd7c997723798a538adf5fb8313cceea272b9a3

src/material.js
```diff
@@ -1,7 +1,9 @@
-import shaderSources from './gl/shader_sources'; // built-in shaders
 import GLSL from './gl/glsl';
 import {StyleParser} from './styles/style_parser';
 
+let fs = require('fs');
```
bcamper: It feels kind of weird to be mixing node `require` along with ES6 `import`, no? (Babel just transpiles the imports to requires.)

package.json
```diff
@@ -64,13 +77,15 @@
   "devDependencies": {
     "babel-runtime": "5.8.29",
     "babelify": "6.4.0",
-    "browserify": "6.1.0",
+    "browserify": "10.0.0",
+    "browserify-derequire": "^0.9.4",
```
bcamper: After a couple hard-to-debug cases where patch/minor upgrades had different behavior, I eventually started pegging all packages to exact versions for consistency (e.g. anyone building on any machine at any time should get the same thing). What do you think?

src/styles/lines/lines.js
```diff
@@ -9,13 +9,17 @@ import {buildPolylines} from '../../builders/polylines';
 import Geo from '../../geo';
 import Utils from '../../utils/utils';
 
+let fs = require('fs');
+let shaderSrc_polygonsVertex = fs.readFileSync(__dirname + '/../polygons/polygons_vertex.glsl', 'utf8');
```
bcamper: After we talked about this I realized that by including the `readFileSync` here, we are duplicating the strings for the polygon shaders (which was one of the motivations for them being shared, though not the only one). It's probably not a huge amount of extra weight, but we could avoid that by importing the `Polygon` style and referencing its shader source properties?

bcamper:
`build.js`, `build_deps.sh`, and `watch.sh` (old watcher) can all be removed now too, right?

bcamper:
Head's up I switched the local port back to 8000, just because it's what we've used historically and is likely to cause the least disruption (we often have branches or local modified copies of demos or Tangram Play that point to local copies of Tangram on 8000, for testing them in tandem).

src/material.js
dmvaldman: I originally tried this, and after some digging I found there is a problem: https://github.com/babel/babelify/issues/81

It looks like for now, we will need to keep this with a require statement, though I agree it's ugly

package.json
dmvaldman: I think that's a good idea. So you would like me to peg `browserify-derequire` at 0.9.4?

src/material.js
bcamper: Oh yeah this rings a bell. It looks like Matt actually went ahead and made this to work with babel: https://github.com/Jam3/brfs-babel

But since it sounds like this may crop up again, maybe not worth the effort to switch it out just to avoid a little ugliness :)

package.json
bcamper: Yeah let's peg at the current versions (we can make an effort to review and update packages every few months). So this one and the other new ones (budo, karma-browserify, etc.). Thanks!

dmvaldman: Seems `shrinkwrap` pegs your dependencies, and also their dependencies...etc etc. Another option.

src/styles/lines/lines.js
dmvaldman: Leading towards exporting the shaders from the polygons module and importing them from lines.js

```
// polygons.js
export {
    shaderSrc_polygonsVertex,
    shaderSrc_polygonsFragment
};
```

```
// lines.js
import {shaderSrc_polygonsVertex, shaderSrc_polygonsFragment} from '../polygons/polygons';
```

Thoughts?

bcamper: Sure, that works. You prefer that to importing the whole `Polygons` symbol and referencing them as:

`vertex_shader_src: Polygons.vertex_shader_src`

Either works for me.

dmvaldman: Yeah. It's more of a bikeshedding philosophical choice, as `lines` shouldn't depend on `polygons` so it shouldn't import it.

bcamper: Makes sense. It might actually make sense for `Lines` to inherit from
`Polygons`, but that's outside the scope of this branch and we can revisit
later to consider that.

dmvaldman:
Alright, pegged the dependencies, reused the same shaders in `lines.js` from `polygons.js` and deleted the deprecated build/watch scripts. I also declared the shader sources as `const`. Are they mutated eventually in the pipeline? If so I can revert this change.

bcamper:
`const` is only going to ensure that the reference of the property defined as `const` doesn't change (e.g. `shaderSrc_spotLight` itself shouldn't be set to a different string at run-time). But once those are copied to a style property like `vertex_shader_src`, all bets are off. `ShaderProgram` does the actual shader string construction, and these strings are [passed in / copied](https://github.com/tangrams/tangram/blob/brfs-valdman/src/styles/style.js#L326-L327) when that is instantiated anyway, so in summary everything should work (and you'd know right away if it didn't anyway... :)

dmvaldman:
👍  I'll be changing the contributing.md file to reflect the changes. Sorry to not do this with the PR.

PR #298, opened 2016-04-21
Optimize shader_program.updateUniform

dmvaldman:
This PR optimizes the `uniform` and `updateUniform` methods of a shader program by removing the es6 argument rest operator and explicitly setting the uniforms on the GPU by calling the uniform method for the type of the uniform. Since we avoid the rest operator, all uniforms are passed as arrays (or typed arrays).

It also automatically enforces the `false` argument when setting a `uniformMatrix3fv` and `uniformMatrix4fv` uniform. This prevents the creation of some unnecessary arrays.

I've also changed the arguments to `updateUniform` to take a `uniform` type instead of passing in the uniform's key to make the function more literal and match the other APIs such as `unsureUniforms` etc. 

The method that sets the uniform on the GPU has been refactored out as `commitUniform`.

### DIFF
--- a/src/camera.js
+++ b/src/camera.js
@@ -51,13 +51,13 @@ export default class Camera {
     setupMatrices (matrices, program) {
         // Model view matrix - transform tile space into view space (meters, relative to camera)
         mat4.multiply(matrices.model_view32, this.view_matrix, matrices.model);
-        program.uniform('Matrix4fv', 'u_modelView', false, matrices.model_view32);
+        program.uniform('Matrix4fv', 'u_modelView', [false, matrices.model_view32]);
 
         // Normal matrices - transforms surface normals into view space
         mat3.normalFromMat4(matrices.normal32, matrices.model_view32);
         mat3.invert(matrices.inverse_normal32, matrices.normal32);
-        program.uniform('Matrix3fv', 'u_normalMatrix', false, matrices.normal32);
-        program.uniform('Matrix3fv', 'u_inverseNormalMatrix', false, matrices.inverse_normal32);
+        program.uniform('Matrix3fv', 'u_normalMatrix', [false, matrices.normal32]);
+        program.uniform('Matrix3fv', 'u_inverseNormalMatrix', [false, matrices.inverse_normal32]);
     }
 
 }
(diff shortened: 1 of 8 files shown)


src/camera.js
```diff
@@ -274,11 +274,11 @@ class IsometricCamera extends Camera {
     }
 
     setupProgram(program) {
-        program.uniform('Matrix4fv', 'u_projection', false, this.projection_matrix);
+        program.uniform('Matrix4fv', 'u_projection', this.projection_matrix);
 
-        program.uniform('3f', 'u_eye', 0, 0, this.viewport_height);
+        program.uniform('3fv', 'u_eye', [0, 0, this.viewport_height]);
```
bcamper: So I think the behavior of `3f` and `3fv` (and the related variants) are actually functionally the same now from the user-facing API perspective? Meaning this should work with either method interchangeably? That's fine if so, just something that occurred to me.

bcamper: To elaborate: what I mean is, when using `3f`, the individual components of the passed vector will be sent to GL as separate arguments, while with `3fv` the whole array will be sent as one argument. But either way the same array value should cause the same result whether `3f` or `3fv` is specified.

dmvaldman: That's correct. Is there any advantage to one way or another?

bcamper: I don't think so, just confirming for my own understanding, since these
traditionally have different signatures but we've collapsed them (which is
fine).

I suppose it might suggest using the underlying GLSL types instead (float,
vec3, mat4, etc), but something to think about later.

src/view.js
```diff
@@ -302,8 +302,8 @@ export default class View {
 
     // Set general uniforms that must be updated once per program
     setupProgram (program) {
-        program.uniform('2f', 'u_resolution', this.size.device.width, this.size.device.height);
-        program.uniform('3f', 'u_map_position', this.center.meters.x, this.center.meters.y, this.zoom);
+        program.uniform('2fv', 'u_resolution', [this.size.device.width, this.size.device.height]);
```
bcamper: We still have unnecessary arrays being created in cases like this, but these aren't actually "new" arrays since the spread operator was creating them before anyway. But we can still factor them out later if it helps (not in the scope of this PR).

dmvaldman: Yeah if we want to avoid the variable-length arguments then we necessarily have to wrap the arguments in an array. The only only option I see is to have multiple `uniform` methods, one per type, such as `program.uniform2f(name, x, y)`.... but if you wanted to track the values in JS land you'd still need to create an array (or object with `x`, `y`, etc keys)

bcamper:
**CPU profiling**

Eliminates a large amount of CPU time when profiling an animated scene (continuous rendering). On non-animated scenes the effect is less pronounced, since it only comes into play when panning or zooming the view (but those are important times to reduce CPU and avoid UI jank!).

Before:
screen shot 2016-04-22 at 10 46 43 am

After:
screen shot 2016-04-22 at 10 45 14 am

Issue #243, opened 2016-02-10
Add synthetic label positions for non-tiled polygons


burritojustice:
Importing the transitland operator polygons and used labels text visible:true -- seeing text at many strange positions rather than one central position.

http://tangrams.github.io/tangram-play/?scene=https://gist.githubusercontent.com/anonymous/78d5327116ac44d26a07/raw/50bfb7e5b086a2697c6b1ccbe5fe97fff12dec13/transitland-operators.yaml#10/41.4556/-72.7487

![image](https://cloud.githubusercontent.com/assets/3979711/12965723/3ae8c0ca-d00d-11e5-820a-e309ca8d5859.png)

bcamper:
Thanks. I hadn't thought of that, but yes it would be possible to derive label positions (centroids or weighted in some other manner) from the original non-tiled data source. It's not happening now because the data has already been tiled client-side (by geojson-vt) by the time the Tangram styling pipeline is applied. I'll leave this ticket open to explore this in the future though.

bcamper:
cc @blair1618 for future consideration

bcamper:
Closed via #302

PR #302, opened 2016-04-22
Add support for GeoJSON polygon centroid labeling

dmvaldman:
This PR is a fix for #243 

It defines a new feature in the GeoJSON pipeline to calculate the centroids of polygons for labeling so that when the polygon is tiled, multiple redundant labels are not calculated for each tile.

screen shot 2016-04-22 at 4 02 49 pm

In the photo above, the centroid labels are in blue, and the tiled labels are in yellow. We only want the centroid labels to appear. The expectation is for the user to not draw the text for the polygons in the data-source, but to filter the data based on points with the property `label_centroid` (this name can easily be changed) and to draw labels for only those points instead. Resulting in a rendering like the one below

screen shot 2016-04-22 at 4 09 08 pm

This PR also supports MultiPolygons (creating a label for each one at its centroid), but doesn't currently support nested polygons (to create annulus-like shapes), whose "centroid" is a less well-defined (but possible, if we would like to add support for it).

@burritojustice @nvkelso @bcamper

### DIFF
--- a/src/geo.js
+++ b/src/geo.js
@@ -178,18 +178,22 @@ Geo.geometryType = function(type) {
 };
 
 Geo.centroid = function (polygon) {
-    let n = polygon.length;
-    let centroid = [0, 0];
+    // Adapted from https://github.com/Leaflet/Leaflet/blob/c10f405a112142b19785967ce0e142132a6095ad/src/layer/vector/Polygon.js#L57
+    let p0, p1, f;
+    let x = 0, y = 0, area = 0;
+    let len = polygon.length;
 
-    for (let p=0; p < polygon.length; p++) {
-        centroid[0] += polygon[p][0];
-        centroid[1] += polygon[p][1];
+    for (let i = 0, j = len - 1; i < len; j = i, i++) {
+        let p0 = polygon[i];
+        let p1 = polygon[j];
+        let f = p0[1] * p1[0] - p1[1] * p0[0];
+
+        x += (p0[0] + p1[0]) * f;
+        y += (p0[1] + p1[1]) * f;
+        area += f * 3;
     }
 
-    centroid[0] /= n;
-    centroid[1] /= n;
-
-    return centroid;
+    return [x / area, y / area];
 };
 
 Geo.multiCentroid = function (polygons) {
--- a/src/geo.js
+++ b/src/geo.js
@@ -179,7 +179,6 @@ Geo.geometryType = function(type) {
 
 Geo.centroid = function (polygon) {
     // Adapted from https://github.com/Leaflet/Leaflet/blob/c10f405a112142b19785967ce0e142132a6095ad/src/layer/vector/Polygon.js#L57
-    let p0, p1, f;
     let x = 0, y = 0, area = 0;
     let ring = polygon[0]; // only use first ring for now
     let len = ring.length;
--- a/src/geo.js
+++ b/src/geo.js
@@ -181,11 +181,12 @@ Geo.centroid = function (polygon) {
     // Adapted from https://github.com/Leaflet/Leaflet/blob/c10f405a112142b19785967ce0e142132a6095ad/src/layer/vector/Polygon.js#L57
     let p0, p1, f;
     let x = 0, y = 0, area = 0;
-    let len = polygon.length;
+    let ring = polygon[0]; // only use first ring for now
+    let len = ring.length;
 
     for (let i = 0, j = len - 1; i < len; j = i, i++) {
-        let p0 = polygon[i];
-        let p1 = polygon[j];
+        let p0 = ring[i];
+        let p1 = ring[j];
         let f = p0[1] * p1[0] - p1[1] * p0[0];
 
         x += (p0[0] + p1[0]) * f;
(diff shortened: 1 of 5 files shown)


meetar:
Yesssssssss 👍

bcamper:
Awesome! Couple things I can think of:

- We should probably make this an optional feature, in case people don't need or want additional features injected into their data. We can do this by adding a property that can be set in the source definition in the scene YAML. Any properties in the YAML already passed to the data source [constructor](https://github.com/tangrams/tangram/blob/master/src/sources/geojson.js#L15-L22), so we just need to choose a suitable name and then reference it there. For example:
   - In scene YAML:
   ```
sources:
   stuff:
      type: GeoJSON
      url: stuff.geojson
      generate_centroids: true
   ```
   - In `GeoJSONSource` constructor:
   ```
    constructor(source) {
        ...
        this.generate_centroids = source.generate_centroids;
    }
   ```
- TopoJSON support: in addition to GeoJSON sources, we'd like to be able to this to TopoJSON. The `TopoJSONSource` class is derived from `GeoJSONSource`, but also overrides `parseSourceData()` to convert the data to GeoJSON. We should be able to just insert the same preprocess call [there](https://github.com/tangrams/tangram/blob/master/src/sources/topojson.js#L16).

burritojustice:
\o/

dmvaldman:
@bcamper can do!

nvkelso:
_Dude, this is so cool! :)_

In Who's On First land, we've made a distinction between **math** centroids, and **label** centroids [1], because they are often different. 

Then in the Vector Tiles service, we calculate server-side label centroids for landuse polygons (and other layers) that span multiple tiles (the same problem you're solving client side for arbitrary GeoJSON), and mark the result: `label_placement: yes`.

It would be nice for Tangram to be consistent with these other projects.

**Proposal, 1a:**

* `generate_label_centroids: true` for the sources block.

While your code probably only generates a single centroid today, multipart geometries could someday be a thing. Plus in the future you could generate multiple centroids per feature per zoom, even at some x-y spacing interval. But I digress! Plural `centroids` so we don't regret the singular form later.

**Proposal, 1b:**

* `label_placement: yes` should be marked on the resulting label centroids

Your example Tangram scene file doesn't need the `$geometry: point` with this approach, though it's useful I guess to show the concept? But we sometimes mark line geometry in vector tiles with `label_placement: yes`.

[1] WOF is mainly relying on Mathew Bloch's excellent MapShaper library for generating label centroids, with light manual curation afterwards.

bcamper:
@nvkelso do you know where the MapShaper code for generating those label centroids is? Whether we want to support that probably comes down to generation speed on the client.

nvkelso:
See: https://github.com/mbloch/mapshaper

I'm not suggesting we do anything super complicated in Tangram today, just emphasizing the **label**'ing part of the centroid :)

dmvaldman:
@nvkelso happy to adapt both those naming conventions. For "multipart geometries" do you mean something other than `MultiPolygon` features? Also I'm happy to look into dealing with non-convex polygon labeling in the future, with mapshaper or another tool.

nvkelso:
@dmvaldman Yes, multipart geoms = MultiPolygon, but where each part gets it's own label placement (optional, often zoom dependent), versus only giving the largest lobe a label placement (the default).

dmvaldman:
So far all the points mentioned (except zoom-dependent label placement) in this discussion have been implemented:

- Filtering key renamed to `label_placement`
- Top-level YAML toggle `generate_label_centroids : true` to preprocess polygon centroids in the sources field. If left out, no preprocessing happens.
- TopoJSON support

bcamper:
@dmvaldman I realized there's an issue with multi-layer data sources. Some data sources have a concept of multiple names layers, and some don't (the latter get a default layer name `_default`). This happens in the `getLayers()` method, which is being called *after* [the centroid calculation](https://github.com/tangrams/tangram/pull/302/files#diff-3d2d77a033e415fb8269c33915ebb3aaR106).

To support both single and multi-layer data sources, we should move the centroid calculation step to after `getLayers()` is called. In addition, instead of being able to pass `preprocessFeatures()` a single collection of features, you'll have to loop over the `source.layers` object, which will have keys indicating layer names with a GeoJSON `FeatureCollection` value for each, e.g.:

```
{
   counties: { type: 'FeatureCollection', features: [...] },
   states: { type: 'FeatureCollection', features: [...] },
   ...
}
```

bcamper:
Also, after looking at some sample data more closely, I'm thinking we should perhaps be just creating a single centroid per `MultiPolygon` (instead of one per constituent polygon). This is because a `MultiPolygon` is a set of polygons indicating one logical feature. For an example like states in the U.S., any multi-polygon state is currently getting labeled multiple times, e.g. Washington State has a number of small islands, each of which will get labeled as "Washington". So it seems the safest thing to do is to simply generate one label centroid, for the first polygon (we can tune this in the future along with smarter label logic).

dmvaldman:
Makes sense. For the latter point would it be best to place the label in the largest polygon?

bcamper:
@dmvaldman yeah probably, it means we have to calculate area, but at least it's a one-time operation, and definitely more likely to be the best placement :) There's a `Geo.polygonArea` method btw.

dmvaldman:
The two recent commits should fix both those issues @bcamper. I may be missing some sane error checking though. Is it always safe to assume a `features` key is defined in a layer, for instance?

bcamper:
Thanks - I also realized when testing this that we weren't supporting single-object (non-`FeatureCollection`) TopoJSON objects, and that our centroid calculation was dumb/wrong. Both fixed now!

bcamper:
@dmvaldman could you please make a PR for the new syntax in the tangram-docs repo? @meetar can show you where. We won't actually merge it until we cut a new release that includes this code, but we should be in better habit of having the docs written/queued up when the feature itself is done.

Issue #290, opened 2016-04-10
Unified point + text rendering style


bcamper:
Currently, `points` (which can be rendered by shaders or optionally configured to draw a sprite) and `text` are two separate rendering styles. This works fine for standalone labels (e.g. roads) or icons, but presents a challenge for typical composite point/text rendering combinations, such as POI icons:

- **Styling is cumbersome**, especially because the point and text components are two independent entities that have no knowledge of each other or logic controlling their rendering in tandem.
   - For example, text must be manually placed relative to the point: to apply a bottom anchor to a text label appearing below an icon, an additional `offset` must be provided to move the text downwards half the height of the icon; setting `anchor: bottom` on the text is also required (as expected), but only positions the text such that its top is directly below the geometric feature point, without taking into account any space occupied by the icon itself.
   - This is further exacerbated in cases where the icon size itself changes across zoom levels; the text offset must also track this size! Here's an abridged example from Bubble Wrap:
   ```
   draw:
      icons:
         size: [[13, 12px], [16, 16px], [18, 19px]]
      text:
         anchor: bottom
         offset: [[13, [0, 6px]], [16, [0, 8px]], [18, [0, 10px]]] # offset tracks half icon size
   ```
- **Placement logic is limited** due to the lack of direct link between the two components. For example, we cannot currently make a text label only display when its linked icon is visible (we can set the `priority` of the icon to be higher than the text, but this does not guarantee that they will display together).

Proposal
===
- The most straightforward way to create a unified rendering style that allows for styling points and text in tandem is to extend the existing `points` style to have an *optional* text component. 
- This text component would appear as a `text` block within a `points` draw rule (similar to the `outline` block within a `lines` draw rule). For example, to create an icon with a text label, using a style `icons` that has `base: points`:

   ```
   draw:
      icons:
         sprite: ...
         size: 16px
         text:
            font: ...
   ```
- A `text` block attached to a `points`-based style would have slightly different default values (again, similar to outlines):
   - **Anchor**:
      - `anchor` would default to `bottom` instead of `center` (though it would be possible to composite a text label over a sprite by setting `anchor: center` and `collide: false`)
      - The point and text can have separate `anchor` values: the `anchor` of the `text` controls the text's placement *relative to the size and position* of its "parent" point, while the `anchor` of the `points` portion moves the *entire entity* (point + text) relative to the underlying feature's point.
   - **Offset**:
[...]

PR #306, opened 2016-05-02
Add text-rendering support to points style

bcamper:
This PR creates a **unified rendering style** for points and text, by extending the existing `points` style to have an *optional* `text` component. See #290 for background.

**`text` block within `points`**
- The `points` style (and any other styles with `base: points`) now supports an optional `text` block within its draw rule. (This is similar syntax-wise to the `outline` block within a `lines` draw rule.)
- This feature is intended for cases where a point (which can be either a sprite, or a shader-drawn point) has a text label meant to be displayed alongside it, such as POIs. Notably, the point is **required**, while the text is **optional** (see collision rules below).
- The `text` block allows for text to be styled with the same syntax as the `text` rendering style, but with different default values that take into account the "parent" point (see below).
- Example: to create an icon with a text label, using a style `icons` that has `base: points`:
```
   draw:
      icons:
         sprite: ...
         size: 16px
         text:
            font: ...
```
- Without an attached `text` block, the `points` style continues as a standalone points-only style as it is today.

**Text behavior**
The default text style behavior is adjusted to account for the parent point:
- **Anchor**:
  - `anchor` defaults to `bottom` instead of `center` (though it is possible to composite a text label over a sprite by setting `anchor: center` and `collide: false`).
  - The point and text can have separate `anchor` values:
    - The `anchor` of the `text` controls the text's placement *relative to the size and position* of its parent point.
    - The `anchor` of the `points` portion moves the *entire entity* (point + text) relative to the underlying geometry point.
- **Offset**:
  - Text is automatically offset to account for its anchor relative to its parent point (see description above).
  - Further manual offset is possible with the `offset` parameter, which moves the text in screen space, e.g. text with `anchor: bottom` will automatically be placed below the sprite, and an additional `offset: [8px, 0]` in the scene file would move the text another 8 pixels to the right.
- **Priority**:
  - The text's `priority` is assigned a default value of `0.5` below the `priority` of its parent point (numerically this means the priority is `+0.5`, since lower numbers are "higher priority"). This can be explicitly overridden by setting a `priority` value in the `text` block, though the text's priority may not be set higher than that of its parent point. This is similar to `outline` handling, where the `order` of the outline cannot be higher than the line fill. (In both cases, the values are capped to their highest/lowest allowed values.)
  - For example, in this case, the icon has `priority: 3`, so the text portion is assigned a priority of `3.5`:
   ```
      draw:
         icons:
            ...
            priority: 3
            text:
               ...
    ```
- **Collision**:
  - The point is required, but its text is optional: while the `points` portion of the style will render according to its collision test, the `text` portion will only render if **both** it and its parent point passed collision tests, e.g. if the point is occluded, then the text won't render either, even if it is not occluded.
  - Different collision behaviors can be achieved by setting the `collide: false` flag on either or both of the point and text:
    - Both `collide: true` (default): nothing will overlap, text will only be rendered if point also fits.
    - Points `collide: false`, text `collide: true`: all points will render, text will render if it fits.
    - Points `collide: true`, text `collide: false`: points will render if they fit, in which case their attached text will also render, even if it overlaps something else.
    - Both `collide: false`: all points and text should render, regardless of overlap.

**Migration/Compatibility**
- Because only new additive syntax is introduced for `points`, and the existing syntax for `points` and `text` is maintained, existing scene files are **fully backwards compatible**.
- The standalone `text` style continues to exist for cases where you **only** need to render text (e.g. no icon). Common examples include road labels for lines (which also have additional text layout logic for angle and in the future curves) and building/address labels for points.
- The migration path is straightforward: in cases where an existing `points`/`text` pair already exists, the `text` portion can simply be indented under `points`, and parameters adjusted (for instance, removing custom `offsets`).


bcamper:
cc @karimnaaji

Issue #267, opened 2016-03-09
Investigate line width precision loss


bcamper:
@blair1618 noticed that lines seem to lose some of their width when overzoomed multiple levels (particularly now that we are overzooming from z16). For example, a 10px line renders consistently from z16-20, but starting at z21, it slims down to around 8px, and at z24 it disappears entirely. This appears to be an issue with precision, given that we are encoding line extrusion vector and width information in a signed short (16-bit int). I verified that hypothesis by switching back to using raw floats in the VBO, which showed no overzooming artifacts.

Given internal tile scale of [0, 4096], we start to run up against the limits of a short as the compounding effects of 2^z overzooming kicks in. The "slimming" effect is likely when the values start to get quantized to an insufficient bit-range, finally disappearing when they collapse into a single point and/or zero width.

We can probably tune the storage of these attributes to address this.

nvkelso:
+1, I've noticed this as well. Setting to a meter unit often solves the problem?

/cc @sensescape

bcamper:
What is a reasonable max zoom we should be supporting? There's some built-in limit (not yet exactly known) between how many zoom levels we can overzoom a given tile, e.g. if we stop requesting at z16, we can use it up to maybe z21 but I don't know how much further. (Another way to solve in the future would be to support tile zoom *ranges*, so use z16 up to z20 or z21 and then request another tile if needed; not ideal server-side either but there's series of tradeoffs.)

nvkelso:
Zoom 16 overzooming does have limitation:

20/48.85744/2.34749

<img width="707" alt="screen shot 2016-03-09 at 13 39 41" src="https://cloud.githubusercontent.com/assets/853051/13651493/70c6f2d4-e5fc-11e5-8165-dba15cb8d09c.png">

Almost all our maps stop at zoom 20 at this point, supporting zoom 21 would be nice. But we'd need a different tile approach to support local areas with increased detail like you mention. That's probably useful to pursue in the mid-zooms anyhow. Once that's working should be easy to extend to zoom 21 tiles.

nvkelso:
Same area in OSM:
https://www.openstreetmap.org/edit#map=21/48.85752/2.34733

<img width="466" alt="screen shot 2016-03-09 at 13 43 45" src="https://cloud.githubusercontent.com/assets/853051/13651635/04c83efc-e5fd-11e5-8bc4-ce1288793420.png">

/cc @zerebubuth @rmarianski

nvkelso:
[...]

PR #310, opened 2016-05-17
Increase line width precision for overzoomed tiles

bcamper:
Line width is stored in a signed short GL attribute. Previously, the units used were the tile's coordinates, which (in Tangram JS) have a scale of 0-4096. When tiles are overzoomed, those units cover more and more screenspace. This causes precision loss and eventually lines disappear entirely, as their width collapses to zero. In Bubble Wrap, for example, this causes 1px building outlines to disappear at z20 (when overzoomed from z16 tiles). See #267 for more info.

This PR adjust the units for line width to multiply by the number of overzoom levels squared, e.g. units for z16 tiles shown at z20 are multipled by 4^2, or 16x. This ensures that the same level of precision is maintained for line widths at all zooms, regardless of overzooming.

Verified this in Differ, and added a z20 test case with building outlines:

0la oekasrqaaaaasuvork5cyii

### DIFF
--- a/src/styles/lines/lines.js
+++ b/src/styles/lines/lines.js
@@ -86,9 +86,9 @@ Object.assign(Lines, {
 
         // convert to units and relative change from previous zoom
         // NB: multiply by 2 because a given width is twice as big in screen space at the next zoom
-        style.width = width * context.units_per_meter;
+        style.width = width * context.units_per_meter_overzoom;
         style.next_width = (next_width * 2) - width;
-        style.next_width *= context.units_per_meter;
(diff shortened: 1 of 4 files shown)


bcamper:
@blair1618 ES also has this issue. This seemed like the most expedient way to fix it (at least on this side), do you have any thoughts?

blair1618:
I applied the same conceptual change to polylines in tangram-es (https://github.com/tangrams/tangram-es/pull/739) and it resolves the problem there as well.

PR #312, opened 2016-05-20
Add textured lines w/support for dash array patterns

bcamper:
This PR adds support for textured lines, with additional support for line dash patterns built in (using syntax similar to SVG/Canvas-style "dash array"). This is commonly used for features such as paths, administrative boundaries, stairs, or railway lines.

**Dash Patterns**
The simplest (and likely most common way) to create line patterns is through a dash pattern. This is done with a new `dash` parameter, as part of a style definition.

- The syntax of `dash` is an array defining a pattern of alternating dashes and spaces, e.g. `[2, 1]` creates a pattern of dashes that are each 2 units long, separated by spaces that are 1 unit long.
- The unit of the dash pattern is the *width* of its line, e.g. a value of `[1, 1]` creates a series of square dashes (separated by square spaces).
- If the dash pattern contains an *odd* number of entries, it is *repeated* to form an even pattern ([as in SVG](https://www.w3.org/TR/SVG/painting.html#StrokeDasharrayProperty)). This means that the `[1, 1]` example above is equivalent to just `[1]`. Similarly, `[3, 1, 1]` would become `[3, 1, 1, 3, 1, 1]`.
- Example of a simple dash pattern, applied to park paths:
```
styles:
   dashed:
      base: lines
      dash: [2, 1]
```
tangram-1464027788197
- The dashes are colored using the feature color as assigned by the layer's `draw` group (aka vertex color).
```
draw:
   dashed:
      color: yellow
```
tangram-1464027797249
- By default, the "spaces" in the `dash` pattern are transparent. Alternatively, an opaque background color can be assigned with the `dash_background_color` parameter, which is useful for typical "stairs" or "railway"-like patterns (note: the dash background color parameter only uses the RGB component of the specified color, since the alpha channel is used for encoding dash/space internally), e.g.:
```
styles:
   dashed:
      base: lines
      dash: [1, 1]
      dash_background_color: grey
```
tangram-1464027836777
- Unlike the dash segment color, the background color is constant across the style and is unaffected by feature/vertex color.
- A line's `outline` can also be used in conjunction with a dashed line to create more complex patterns, with either the outline or inside line having a dash pattern (or both):

tangram-1464027892685
tangram-1464028007658
tangram-1464028575781

**Line Textures**
The dash patterns are rendered using an underlying, generic line texture feature.

- The `texture` style parameter has been extended to `lines`.
- If provided, the texture is repeated across the line, with the width of the texture matching the width of the line, and the texture y coordinate scaled to match the aspect ratio of the image (its height over its width).
- The pattern specified with `dash` is rendered into a texture, which is then set to the style's `texture` property.
```
styles:
   arrows:
      base: lines
      texture: arrow.png
```

tangram-1464038836553

**Implementation Notes**
- Dashed line patterns have transparency by default, but must often respect feature `order` (for interleaving paths or rail lines with roads, etc.). Rather than force all dashed lines to draw with an `inlay` blend mode (which would be limiting from a style authoring perspective and isn't supported in ES right now anyway), this implementation uses simple alpha testing (or "cutout"). Pixels in the line texture with alpha values below 0.5 are `discard`'ed when alpha blending is unavailable (`blend: opaque`, `add`, or `multiply`); when alpha blending is available (`blend: overlay` or `inlay`), the pixel's alpha is set to `0.0`. While it's generally preferable to avoid a shader branch and/or `discard`, I believe this is a reasonable compromise for a rendering feature usually reserved for a small number of features. It does however lead to poor aliasing in some cases.
- The `dash` pattern is [rendered to a texture](https://github.com/tangrams/tangram/blob/dasharray/src/styles/lines/dasharray.js) at run-time, as a simple set of alternating dashes and spaces (a value of 1 in the dash array is rendered as a single pixel) . The alpha channel is used to encode whether the pixel is a "dash" or "space". In the future, this could possibly be improved by rendering an SDF for antialiased patterns and/or other dash "shapes" (e.g. rounded cap / circle dash segments, currently more advanced patterns like this can be achieved through custom shader styles). An atlas of dash patterns could also be kept, to reduce texture use (render all patterns to one texture), or in case the same dash pattern is used by multiple styles.
- Line UVs show severe distortion in some cases. We should investigate either more intelligent UVs that take into account line curvature (maybe using a Bezier curve to calculate UVs in the shader? could be higher quality but expensive), or perhaps rendering some line patterns with different geometry that is aware of the pattern (e.g. adjusting the pattern to an optimal fit within each line segment, avoiding pattern stretch at segment joins, etc.).

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -333,6 +333,7 @@ layers:
             filter: { kind: ferry }
             draw:
                 lines:
+                    style: dashed
                     color: '#8db3ce'
                     width: [[14, 1px], [18, 2px]]
                     outline:
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -140,6 +140,9 @@ styles:
                     // Just visualize the grid lines directly
                     color = vec4(vec3(1.0 - min(line, 1.0)), 1.0);
 
+    dash:
+        base: lines
+        dasharray: [2, 1]
 
 sources:
     mapzen:
@@ -321,8 +324,9 @@ layers:
             filter: { kind: path }
             draw:
                 lines:
-                    color: '#bbb'
-                    width: [[15, 0px], [18, 2px]]
+                    style: dash
+                    color: white
+                    width: [[15, 0px], [18, 3px]]
                     outline:
                         width: 0
         ferry:
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -140,9 +140,9 @@ styles:
                     // Just visualize the grid lines directly
                     color = vec4(vec3(1.0 - min(line, 1.0)), 1.0);
 
-    dash:
+    dashed:
         base: lines
-        dasharray: [2, 1]
+        dash: [2, 1]
 
 sources:
     mapzen:
@@ -324,7 +324,7 @@ layers:
             filter: { kind: path }
             draw:
                 lines:
-                    style: dash
+                    style: dashed
                     color: white
                     width: [[15, 0px], [18, 3px]]
                     outline:
--- a/src/builders/polylines.js
+++ b/src/builders/polylines.js
@@ -32,8 +32,9 @@ export function buildPolylines (
         tile_edge_tolerance,
         texcoord_index,
         texcoord_scale,
-        texcoord_normalize,
+        texcoord_width,
         texcoord_ratio,
+        texcoord_normalize,
         scaling_index,
         scaling_normalize,
         join, cap,
@@ -69,7 +70,7 @@ export function buildPolylines (
         texcoords: texcoord_index && [],
         texcoord_normalize,
         min_u, min_v, max_u, max_v,
-        v_scale: 1 / ((width * texcoord_ratio) * v_scale_adjust), // scales line texture as a ratio of the line's width
+        v_scale: 1 / ((texcoord_width * texcoord_ratio) * v_scale_adjust), // scales line texture as a ratio of the line's width
         total_dist: 0,
         num_pairs: 0
     };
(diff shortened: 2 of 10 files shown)


dmvaldman:
😻

bcamper:
This should be complete. @blair1618 @karimnaaji @tallytalwar any ES commentary?

bcamper:
Used [rstats](http://spite.github.io/rstats/) to compared a scene that just draws all lines (as a kind of worst case scenario), with dashed and solid line variants, and only saw marginal performance difference in frame time.

karimnaaji:
This looks great! When trying this I naively thought I could apply both dash + texture before looking into the details of the implementation:
```yaml
dashed:
    base: lines
    dash: [1, 10]
    texture: arrow.png
```
I can see why we don't support that in the dash part, since the problem would be to get the correct uvs along the line geometry for the texture to stretch and match exactly the size of the non-discarded part of the dash.. But I think that we should be put a note (at least in docs) about those option being mutually exclusive.
On -es I can expect it to be probably a little more aliased on regular screens.

bcamper:
@karimnaaji good point, we definitely do need to mark `dash` as taking priority. We *could* try to apply a dash pattern that masks against an existing texture, but from what I've seen so far, it's pretty hard to get a texture to look good anyway (maybe it will get easier if we can get line offsets in the tiles so we can make line patterns continuous across tile boundaries), but until then I'd prefer to wait and see how often people try to use their own textures to see if there's a need there (since it would require more code paths that would have to bind multiple textures, one for the dash pattern and one for the provided texture).

bcamper:
Comfortable enough having discussed this with the team to merge, and can make tweaks before 0.8 release if needed.

PR #313, opened 2016-05-23
Element array squash

dmvaldman:
This PR provides GL support for element arrays. Currently the polyline feature is the only feature to make use of them (with points and polygons to follow). 

The polyline builder in `src/builders/polylines.js` was rewritten to make use of element arrays and to improve the runtime performance. It also more closely mimics the es code. Further optimizations were made, such as [not creating line caps at the beginning of tile boundaries](https://github.com/tangrams/tangram/blob/element-array-squash/src/builders/polylines.js#L117), and [dynamically setting the triangle count](https://github.com/tangrams/tangram/blob/element-array-squash/src/builders/polylines.js#L452-L461) for a fan joint or round line cap. Because of the latter, you no longer need to specify a zoom filter for rounded lines in a scene file.

All in all, we get a significant speed up in JS execution (~30%) in typical use. We also use less memory on the GPU side as a result of switching to element arrays (30-40%), and create far fewer triangles for joins and caps (30% - 67% depending on zoom).

### DIFF
--- a/src/builders/polylines.js
+++ b/src/builders/polylines.js
@@ -104,7 +104,7 @@ export function buildPolylines (lines, width, vertex_data, vertex_template,
         coordNext = line[1];
 
         // If first pair of points is redundant, slice and push to the lines array
-        if (coordCurr[0] == coordNext[0] && coordCurr[1] == coordNext[1]) {
+        if (Vector.isEqual(coordCurr, coordNext)) {
             if (line.length > 2) {
                 lines.push(line.slice(1));
             }
--- a/src/builders/polylines.js
+++ b/src/builders/polylines.js
@@ -2,7 +2,7 @@
 
 import Vector from '../vector';
 import Geo from '../geo';
-import { default_uvs, outsideTile, isCoordOutsideTile} from './common';
+import {outsideTile, isCoordOutsideTile} from './common';
 
 const zero_vec2 = [0, 0];
 
@@ -54,11 +54,10 @@ export function buildPolylines (lines, width, vertex_data, vertex_template,
     }
 
     // Texture Variables
-    var min_u, min_v, max_u, max_v, v_scale;
+    var v_scale;
     if (texcoord_index) {
         texcoord_normalize = texcoord_normalize || DEFAULT.TEXCOORD_NORMALIZE;
         texcoord_ratio = texcoord_ratio || DEFAULT.TEXCOORD_RATIO;
-        [min_u, min_v, max_u, max_v] = texcoord_scale || default_uvs;
         v_scale = 1 / (width * texcoord_ratio * v_scale_adjust); // scales line texture as a ratio of the line's width
     }
 
@@ -77,8 +76,7 @@ export function buildPolylines (lines, width, vertex_data, vertex_template,
         scaling_normalize,
         v_scale,
         texcoord_index,
-        texcoord_normalize,
-        min_u, min_v, max_u, max_v
+        texcoord_normalize
     };
 
     // Buffer for extra lines to process
@@ -430,11 +428,13 @@ function addFan (coord, nA, nC, nB, uvA, uvC, uvB, isCap, context) {
 
         if (context.texcoord_index !== undefined) {
             if (isCap){
+                // UV textures go "through" the cap
                 affine_uvCurr = Vector.rot(affine_uvCurr, angle_step);
                 uvCurr[0] = affine_uvCurr[0] + uvC[0];
-                uvCurr[1] = affine_uvCurr[1] * context.half_width * context.v_scale + uvC[1];
+                uvCurr[1] = affine_uvCurr[1] * context.half_width * context.v_scale + uvC[1]; // scale the v-coordinate
             }
             else {
+                // UV textures go "around" the join
                 uvCurr = Vector.add(uvCurr, uv_delta);
             }
         }
--- a/src/builders/polylines.js
+++ b/src/builders/polylines.js
@@ -9,20 +9,21 @@ const zero_vec2 = [0, 0];
 // Build tessellated triangles for a polyline
 const CAP_TYPE = {
     BUTT: 0,
-    SQUARE: 2,
-    ROUND: 3
+    SQUARE: 1,
+    ROUND: 2
 };
 
 const JOIN_TYPE = {
     MITER: 0,
     BEVEL: 1,
-    ROUND: 3
+    ROUND: 2
 };
 
 const DEFAULT = {
     MITER_LIMIT: 3,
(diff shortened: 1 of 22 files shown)


dmvaldman:
Recent commits include support for element arrays in both polygon (with updated verison of earcut, which requires an `npm install`) and point builders.

src/builders/polylines.js
```diff
@@ -40,579 +43,444 @@ export function buildPolylines (
...
+        // Loop backwards through line to a tile boundary if found
+        if (closed_polygon && join_type === JOIN_TYPE.MITER) {
+            var boundaryIndex = getTileBoundaryIndex(line);
+
+            if (boundaryIndex !== 0) {
+                // create new line that is a cyclic permutation of the original
+                var newLine = line.splice(boundaryIndex);
+                Array.prototype.push.apply(newLine, line.slice(1));
+                newLine[newLine.length] = newLine[0];
+                lines.push(newLine);
```
bcamper: We can't mutate the underlying geometry while building, because it can be referenced by later code, such as:

- Drawing the same feature with multiple draw groups (fill a polygon and also draw an outline around the perimeter)
- When the scene is rebuilt (either to switch styles, or tiles go in and out of view, etc.)

So the builders need to be idempotent to the arguments passed in.

Currently this issue is manifesting itself for styles that both draw both a polygon fill and a line outline for buildings, such as Bubble Wrap. You can repro by loading Bubble Wrap, and after initial page load, calling `scene.rebuild()`, which leads to earcut exception, because the geometry has been modified by a previous call to the line builder and is no longer valid for triangulation.

bcamper: Separate but related issue: I'm seeing cases, such as with building feature `-3696564` (which you can isolate with `filter: { id: -3696564 }`) where new lines that are only a single element long (e.g. not a valid line) are being inserted.

src/builders/polylines.js
```diff
@@ -40,579 +43,444 @@ export function buildPolylines (
         miter_limit
     }) {
 
-    var cornersOnCap = corners_for_cap[cap] || 0;         // default 'butt'
-    var trianglesOnJoin = triangles_for_join[join] || 0;  // default 'miter'
+    var cap_type = cap ? CAP_TYPE[cap.toUpperCase()] : CAP_TYPE.BUTT;
```
bcamper: It feels unnecessary here to have to do a string transform to uppercase here (for no functional difference as far as I can tell besides a naming convention). Can we just use the lowercase which is what the official style spec is, and avoid any extra work (related: it isn't actually specified that uppercase or mixed case cap/join values from the scene file should work, and I don't know what the ES behavior is).

dmvaldman: Can do. You're right, it was just for a stylistic uppercase reason that I personally like for constants, but shouldn't come with a performance cost.

dmvaldman:
- 96cf00b fixes the `lines` mutation so make the `buildPolylines` function not have side effects
- 1d57c84 uses lowercase keys to match the scene file for defining join and cap types
- 80c5faa fixes textures for round caps and a square cap not existing for the beginning of a line (this bug may also exist on the es codebase)

dmvaldman:
- 3f85d80 Fixes the texture coordinate bug that would cause diagonal texture artifacts (uv coordinates were incorrectly assigned)
- 6b83951 Includes a guard to gracefully degrade when element indices overflow and no Uint32 index support is available

bcamper:
I'm not overly concerned about enforcing total consistency in variable name convention (e.g. `under_score` vs. `camelCase`), even if that would be ideal (and if I was starting the project today I would use camel case anyway). But, perhaps we could change a couple of these to keep consistency within classes where possible, specifically:

- In `VBOMesh`: `toggleElementArray` and `elementBuffer`
- In `VertexData`: `elementBuffer`
- In `VertexElements`: members are camel case (e.g. `hasOverflown`), but at least it's internally consistent, so optional to change

What do you think?

The old polyline builder had mixed styles anyway so I suppose we can leave it...

src/gl/vbo_mesh.js
```diff
@@ -12,6 +12,7 @@ export default class VBOMesh  {
         this.gl = gl;
         this.vertex_data = vertex_data; // typed array
         this.vertex_layout = vertex_layout;
+        this.vertex_elements = options.vertex_elements;
         this.buffer = this.gl.createBuffer();
```
bcamper: Maybe we should rename this `vertex_buffer` to distinguish, since we now have a second (element) buffer?

dmvaldman: Yeah I think consistency is best. My unconscious habit is to camelCase, so I need to break that. I also default to `var` for the same reason, and we should be consistent with that too.

dmvaldman: On a related note what do you think of passing the `options.vertex_elements` as an argument to the `VBOMesh` constructor as opposed to inside of the `options` object?

bcamper: Yep that makes sense (but still support it being null for `drawArrays`)

dmvaldman: I think `vertex_buffer` is a little confusing, as this suggest a buffer of vertices, which is what's stored in `vertex_data`. We could rename `vertex_data` to `vertex_buffer` and `vertex_elements` to `element_buffer` or `indices_buffer`.

dmvaldman:
- 4e1f0a9 under_score styling
- 5c2df73 removes `texcoord_scale` dependency in polyline builder
- 279dff0 implements texture coordinates that "pass through" line caps (needs to be augmented to use `texcoord_width` when merged into master
- 95351b0 puts `vertex_elements` on equal footing with `vertex_data`. Both are passed in as arguments into `VBOMesh` instead of inside of an `options` argument

src/gl/vbo_mesh.js
bcamper: Well `_buffer` (in my original implementation) was for the GL buffer itself (from GL terminology, e.g. `gl.createBuffer`). I was using `_data` as raw incoming ArrayBuffer data (so yes potential confusion over two "buffers"). In that nomenclature it would be `vertex_data` and `element_data` for the JS data, and `vertex_buffer` and `element_buffer` for the GL buffers. Or some other pairing would work, but I do think something that identifies these as distinct is good.

dmvaldman: Those changes sound good to me.

bcamper: Cool I can handle that post merge.

bcamper: Updated in https://github.com/tangrams/tangram/commit/04018d555aad0248fb8ad8128da000a020db58ed

PR #322, opened 2016-06-01
Optimize feature filters and config sync

bcamper:
This PR optimizes a few pieces of the feature/geometry build process, namely:

- "Simple" key/value filter conditions (for both feature properties, and `$`-style context properties like `$zoom`) are pulled out and given separate, optimized treatment. This includes matches on single or multiple values, e.g. `kind: rail`, or `landuse: [residential, commercial, industrial]`.
  - These matches are put into a simplified data structure, and are matched on in "native" JS, rather than in the dynamically compiled JS filters. (As with similar optimizations for zoom matching, these conditions are removed from the filter before it is compiled.) Property matches happen *after* zoom conditions (if any), but *before* remaining dynamic JS filter (if any).
   - See 
[`buildPropMatches()`](https://github.com/tangrams/tangram/compare/optimize-filters#diff-deabab718e02af3782a643dcdbbde01aR162), [`doPropMatches()`](https://github.com/tangrams/tangram/compare/optimize-filters#diff-deabab718e02af3782a643dcdbbde01aR193), and the [execution of these](https://github.com/tangrams/tangram/compare/optimize-filters#diff-deabab718e02af3782a643dcdbbde01aR401).
- Filters and their list of parent draw rules are calculated [lazily](https://github.com/tangrams/tangram/compare/optimize-filters#diff-deabab718e02af3782a643dcdbbde01aR392), the first time the layer is encountered (e.g. a feature matches its parent rule, or it is a top-level rule).
  - This means any JS function compilation for either filter or draw parameters are deferred until needed, amortizing their cost and decreasing initial startup time to process the scene config.
- Compiled JS functions are now serialized into strings (before the config is sent to the workers) only in cases where the scene config is sourced from a JS object.
  - When the scene source is a URL/file (the most common case), we know that it will not contain any compiled JS functions (it may contain strings that will be compiled into functions later).
  - The scene source can be a JS object either in cases of `scene.load(object)`, or when `scene.updateConfig()` or `scene.rebuild()` is called (in the latter two cases, it is a JS object modified in place).
  - This speeds up initial config parse/sync time to workers.
- Minor misc changes to remove ES6 iterators that blocked V8 optimization path.

Together these changes show substantial improvements of between **20%-45%** less build time.

With Bubble Wrap, in NYC at [40.697299008636755/-73.97094726562501], with a 1020x540 viewport, build times (average of 20 rebuilds) in milliseconds:

master/0.7.2 | optimize-filters | % reduction
-----------------|----------------------|--------------
694 | 380 | 45.24%
855 | 638 | 25.38%
1166	 | 859 | 26.33%
1077 | 739 | 31.38%
1148 | 876 | 23.69%
1034 | 817 | 20.99%
1020 | 767 | 24.80%
1163 | 865 | 25.62%
964 | 756 | 21.58%
832 | 656 | 21.15%
517 | 383 | 25.92%
518 | 376 | 27.41%

### DIFF
--- a/src/scene.js
+++ b/src/scene.js
@@ -132,7 +132,12 @@ export default class Scene {
 
                 // Only retain visible tiles for rebuilding
                 this.tile_manager.pruneToVisibleTiles();
-                return this.updateConfig();
+
+                // Scene loaded from a JS object may contain functions which need to be serialized,
+                // while one loaded from a URL does not
+                return this.updateConfig({
+                    serialize_funcs: (typeof config_source === 'object')
+                });
             }).then(() => {
                 this.updating--;
                 this.initializing = null;
@@ -646,7 +651,7 @@ export default class Scene {
     // Rebuild all tiles
     // sync: boolean of whether to sync the config object to the worker
     // sources: optional array of data sources to selectively rebuild (by default all our rebuilt)
-    rebuildGeometry({ sync = true, sources = null } = {}) {
+    rebuildGeometry({ sync = true, sources = null, serialize_funcs } = {}) {
         return new Promise((resolve, reject) => {
             // Skip rebuild if already in progress
             if (this.building) {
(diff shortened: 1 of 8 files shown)


bcamper:
Note: I consider this approach an interim one, because there is increasing logic duplication between the "native" and "compiled" JS function filter checks. Eventually we should look at reimplementing filters and perhaps a pure-native-JS approach (which would be most similar to ES) could be faster overall, with less code. Or perhaps some hybrid is still faster.

PR #326, opened 2016-06-09
Add support for custom fonts

bcamper:
Adds supports for custom web fonts, either loaded directly by Tangram, or via an external CSS stylesheet.

- The scene file has a new optional, top-level `fonts` block, which is a mapping of **font definitions**.
- The *key* of each font definition is the font **family** name, and the *value* is one or more **font face definitions** (a single object, or an array of objects).
- Each **font face definition** is an object with the following possible parameters:
  - `weight`: defaults to `normal`, may also be `bold` or a numeric font weight, e.g. `500`
  - `style`: defaults to `normal`, may also be `italic`
  - `url`: the URL to load the font from. For maximum browser compatibility, fonts should be either `ttf`, `otf`, or `woff` (`woff2` is [currently supported](http://caniuse.com/#search=woff2) in Chrome and Firefox but not other major browsers). As with other scene resources, `url` is relative to the scene file.
- Alternately, instead of an object, a font face definition can be set to the special string value **`external`**, which indicates that the font will be loaded via an external CSS stylesheet (such as via Google Fonts).
  - The `external` option is JS-only syntax. Tangram JS still needs to know that these custom fonts will be used, even if it is not responsible for loading them, because it must wait to render text until they have loaded. There is currently no ES equivalent, so this is a no-op in ES.
- All font face parameters are *optional*.
  - If a `url` is specified, the font is loaded by injecting one or more CSS `@font-face` rules into the document head.
  - If no `url` is specified, Tangram JS will assume that the font has already been loaded via external CSS.

Example of a font with a single face definition:
```
fonts:
    Montserrat:
        url: https://fonts.gstatic.com/s/montserrat/v7/zhcz-_WihjSQC0oHJ9TCYL3hpw3pgy2gAi-Ip7WPMi0.woff
```

Example of a font with multiple face definitions:
```
fonts:
    Open Sans:
        - weight: 300
          url: fonts/open sans-300normal.ttf
        - weight: 300
          style: italic
          url: fonts/open sans-300italic.ttf
        - weight: 400
          url: fonts/open sans-400normal.ttf
        - weight: 400
          style: italic
          url: fonts/open sans-400italic.ttf
        - weight: 600
          url: fonts/open sans-600normal.ttf
        - weight: 600
          style: italic
          url: fonts/open sans-600italic.ttf
        - weight: 700
          url: fonts/open sans-700normal.ttf
        - weight: 700
          style: italic
          url: fonts/open sans-700italic.ttf
        - weight: 800
          url: fonts/open sans-800normal.ttf
        - weight: 800
          style: italic
          url: fonts/open sans-800italic.ttf
```

Example of a custom font loaded via external CSS:
```
fonts:
    Poiret One: external
```

And in the document:
`<link href='https://fonts.googleapis.com/css?family=Poiret+One' rel='stylesheet' type='text/css'>`

Any fonts specified in the `fonts` block can then be referenced in the `font` block when rendering with `text` or `points`-based styles. Tangram JS will block text rendering while custom fonts load, using the [`FontFaceObserver`](https://github.com/bramstein/fontfaceobserver) polyfill for [CSS Font Loading Module Level 3 events](https://drafts.csswg.org/css-font-loading/). 

This PR replaces the earlier implementation from #284.

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -475,10 +475,9 @@ layers:
                     text:
                         text_source: global.language_text_source
                         font:
-                            family: Helvetica
+                            family: Montserrat
                             size: 12px
                             fill: black
-                            weight: bold
 
         # add generic icon at high zoom
         generic:
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -41,6 +41,10 @@ scene:
     background:
         color: '#9dc3de'
 
+fonts:
+    Montserrat:
+        url: https://fonts.gstatic.com/s/montserrat/v7/zhcz-_WihjSQC0oHJ9TCYL3hpw3pgy2gAi-Ip7WPMi0.woff
+
 textures:
     pois:
         url: images/poi_icons_32.png
--- a/index.html
+++ b/index.html
@@ -12,6 +12,7 @@ http://mapzen.com/tangram
     <title>Tangram: WebGL for OpenStreetMap</title>
     <link rel="stylesheet" href="demos/css/main.css" />
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.0.0-rc.1/leaflet.css" />
+    <link href='https://fonts.googleapis.com/css?family=Montserrat' rel='stylesheet' type='text/css'>
   </head>
 
   <body>
--- a/index.html
+++ b/index.html
@@ -12,7 +12,6 @@ http://mapzen.com/tangram
     <title>Tangram: WebGL for OpenStreetMap</title>
     <link rel="stylesheet" href="demos/css/main.css" />
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.0.0-rc.1/leaflet.css" />
-    <link href='https://fonts.googleapis.com/css?family=Montserrat' rel='stylesheet' type='text/css'>
   </head>
 
   <body>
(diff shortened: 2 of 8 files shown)


blair1618:
In the case where no URL is given: what does a '"truthy" value mean in YAML?

We should also decide on behavior for conflicting font face definitions. For example:
```yaml
fonts:
  CustomFontFamily:
    - weight: 400
      style: bold
      url: some/url/
    - weight: 400
      style: bold
      url: some/other/url
```
Since face definitions are ordered, we could simply declare that the last definition for a face is used.

This could hypothetically be enforced by the syntax itself, if the keys for faces were objects defining the parameters, e.g.
```yaml
fonts:
  CustomFontFamily:
    { weight: 400, style: bold }: some/url
```
But this syntax isn't well supported by the YAML parser in tangram-js (or github, it seems).

bcamper:
Discussing with @blair1618, rather than having any truthy value used to indicate externally loaded fonts, perhaps we should only accept a single, more explicit value, such as the string 'external', 'CSS', or 'DOM':

```
fonts:
    Poiret One: external # font will be loaded via external CSS
```

This is more precise and hopefully would be less opaque and more self-documenting for scene authors (especially those from less technical backgrounds, for whom values like `{}`, `null` or `true/false` may not be as obvious here).

blair1618:
^ I like this. It will be much better at preventing accidents like putting a font URL right after the family name - we can now issue an error for that.

bcamper:
Switched to requiring `external` for external CSS-loaded fonts, and updated description.

meetar:
will be documented with https://github.com/mapzen/mapzen-docs-generator/pull/161

PR #346, opened 2016-06-28
Re-initialize web workers when data source custom scripts change

bcamper:
Tangram loads any custom data source scripts into web workers at the time of worker instantiation, in a single `importScripts()` call along with itself. This method has mostly been settled on as the most reliable, with the loading of large scripts (like Turf.js) into web workers at run-time flaky on some browsers.

However, the previous code only did this when workers were initially created, meaning that loading any subsequent scenes that had *different* custom scripts (or if the initial scene had no custom scripts, followed by loading a scene that did include them) would fail.

This PR instead checks to see if a newly loaded scene has different custom scripts from those previously loaded, and if so, simply destroys and re-creates the web workers (this was the simplest way to accomplish this under the current architecture, and this is a rare scenario in any case).

### DIFF
--- a/src/scene.js
+++ b/src/scene.js
@@ -218,15 +218,7 @@ export default class Scene {
 
         this.sources = {};
 
-        if (Array.isArray(this.workers)) {
-            log.setWorkers(null);
-            this.workers.forEach((worker) => {
-                worker.terminate();
-                WorkerBroker.removeWorker(worker);
-            });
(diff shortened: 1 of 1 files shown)


bcamper:
This should fix the issue observed in differ here which can be used to verify the fix, I've verified locally and will upon next Tangram release as well: https://github.com/tangrams/differ-tests/issues/4

PR #349, opened 2016-06-30
Add option to require text when drawing point

bcamper:
We added the ability to attach `text` to a `points` (or `points`-based style) draw group in v0.8.0. However, in that implementation, the text portion is **optional**, while the point itself is **required**: that is, the text will only draw if it's parent point is visible, but the converse is not true (the point will still draw, even if the text does *not* fit).

This PR adds an option to **require** the text: the point and text are co-dependent and will draw or not draw as a single unit, with a new `required` option:

```
draw:
   points:
      color: red
      size: 6px
      text:
         required: true # only draw parent point if text can also be drawn
         font:
            ...
```

The default behavior is unchanged (e.g. `required: false`), for backwards compatibility (*however this is not necessary if we feel that `required: true` is a better default*).

Following are some examples with/without required text. Note that in some cases, the number of culled POIs is modest/expected, while in others, many more are removed, likely due to tile boundaries (compound icon/text with long labels just occupy a lot of space). This is a concern, and perhaps reason to keep the default to not require text. We can improve with future label improvements such as multiple label anchor placements (e.g. if label doesn't fit on bottom, try placing on right, etc.), and moving label collision into screenspace (on the main thread) to allow labels to cross tile boundaries (a bigger but ultimately important change).

output_kbnnpf

output_l6br2o

output_l8vupr

output_38mxej

**Road Shields**
Related work in this PR allows linked text and points to not collide with one another, which helps to enable road shields. Further work is likely required for full road shield support, but an initial example has been added to the bundled Tangram demo:

screen shot 2016-06-29 at 7 22 58 pm

screen shot 2016-06-29 at 7 23 49 pm

screen shot 2016-06-29 at 7 24 29 pm


bcamper:
@nvkelso @karimnaaji seeking feedback on *default behavior* for requiring / not requiring text, and for the syntax terminology (e.g. `required: true` or something else).

bcamper:
@dmvaldman we can consider merging the additional "linking" code here, which allows for single-direction (text requires point but not vice versa) or two-way links (text and point require each other, so either both or neither draw), into a future label grouping / compound mechanism as a follow up to your other label work. But I'd like to move ahead with adding this functionality for now.

bcamper:
This adds new behavior but is backwards compatible (if `require: true` is made the default, it will not be exactly compatible, but v0.8.1 scenes would still render properly, just with different possible collision outcomes), my inclination is to include it in a v0.8.2 release, rather than deferring to v0.9 (or bumping up v0.9 just for this).

nvkelso:
If icons default to no text and show up generally, the moment text is added I vote for defaulting it to `required: true` but allowing people to set it to `required: false`. (I would set `required: false` on a few features in house styles today, probably including peaks, waterfalls, train stations, and info spots – but the vast majority should be `required: true`.)

I favor releasing this with a v0.9 release and bumping up the version now. Then later releases could "restore" more labels using the multiple position option.

bcamper:
If we are going to make required text the default (which is OK with me), then maybe we should change the semantics to `optional`, with default `optional: false`, but you can set it to `optional: true` if desired.

I'll bump this to v0.9 instead and combine with some other upcoming features (exact set TBD).

karimnaaji:
Since the keyword affects the visibility of the parent, what about having it a little more verbose, under the parent `point` directly? 

`text_optional` or `require_text`:
```yaml
draw:
   points:
      color: red
      size: 6px
      require_text: true
      text:
         font:
```

bcamper:
Either is equally implementable, I guess from my perspective it seems more natural to have it under `text`, but I'm happy to put it under `points` if that's the consensus.

What do you think @nvkelso @blair1618?

meetar:
In Karim's proposal, if require_text: true is specified but a text block is
not, would the style fail?

nvkelso:
> In Karim's proposal, if require_text: true is specified but a text block is not, would the style fail?

I could see reading it that way, but there's already a way to accomplish that via `filter: { name: true }` so yeah loose English? ;) 

So I think it's okay, and I like saying `require_text` more than `required` or `optional` as it says what it is requiring.

meetar:
Right, I'm not asking about how to achieve something but how to define the
behavior: would there be a default text style if none was provided in that
case? I assume not, so I'd assume that the style would fail and the point
would not be shown, but I'm not 100% sure how all the related pieces of
syntax are defined.

On Thu, Jul 7, 2016 at 1:15 PM, Nathaniel V. KELSO <notifications@github.com

blair1618:
I think either of these syntax options could make sense, but `points: { require_text: ... }` could be interpreted to have a very different meaning than what we were originally discussing (this is what Peter was asking about). In a casual reading, `require_text: true` sounds like it means: 

"For this point to be drawn, it must have a 'text' draw rule and the text must not be occluded"

whereas we originally wanted an option that means: 

"For this point to be drawn, _if_ there is a 'text' draw rule the text must not be occluded". 

The first meaning could be useful in some situations, but it is clearly different from (and impossible to denote with) a syntax where `required`/`optional` is a member of the `text` draw rule itself. 

It seems to me that the best and simplest choice right now is something like:
```yaml
points:
  ...
  text:
    optional: true
    ...
```
A more specific name like `allow_occlusion` may be helpful though.

bcamper:
Thanks. I would like to stick with this then, I think it is most consistent with existing syntax:

```
points:
  ...
  text:
    optional: true
    ...
```

meetar:
documented with https://github.com/tangrams/tangram-docs/commit/fd3fd495d1fa4a847a4b45b262033587b5d89fb2

Issue #344, opened 2016-06-24
Multiple anchor placements w/fallbacks


bcamper:
Fallback through a list of possible anchor placements, see https://github.com/tangrams/tangram-es/pull/806.

bcamper:
released in 0.9.0!

PR #351, opened 2016-07-08
Multiple anchor placements for point labels

bcamper:
Support for specifying multiple `anchor` values on point labels. This is to go with similar upcoming functionality in ES: https://github.com/tangrams/tangram-es/pull/806. Completes #344.

During label placement, each anchor position is tried in the order provided. This substantially improves the density of placed labels.

Example:
```
draw:
   points:
      ...
      text:
         anchor: [bottom, right, left, top]
         ...
```

Live preview (thanks to precog!):
https://precog.mapzen.com/tangrams/tangram/multi-anchor/#35.667086218359884,139.75703186094174,17.39925782629959

Before/after:
output_vg1b0b
output_qgegkn
output_qcpvwd

Note that while this does work for standalone `text`, it's likely more useful for `text` attached to `points`, such as a POI icon/text pair.

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -415,7 +415,6 @@ layers:
                     color: '#bddec5'
 
     road_labels:
-        visible: false
         data: { source: mapzen, layer: roads }
         filter: { name: true, aeroway: false, tunnel: false, railway: false, not: { kind: rail }, $zoom: { min: 10 } }
         highway:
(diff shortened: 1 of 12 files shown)


bcamper:
Note: as is, this does not change the **default** anchor values, which are `anchor: center` for standalone `text`, and `anchor: bottom` for `text` attached to `points`.

Given the improved placement, my inclination is to change the default for attached text to take advantage of this, for example to:
`anchor: [bottom, right, left, top]`

Or even something like (maximum placements, but diagonals can feel a bit less natural):
`anchor: [bottom, right, left, top, bottom-right, bottom-left, top-right, top-left]`

@nvkelso?

nvkelso:
WOW, this is fantastic!

Slight variation to prefer top before right (more placements are better):

anchor: [bottom, top, right, left, bottom-right, bottom-left, top-right,
top-left]

PR #358, opened 2016-07-15
Add ability to load scenes from zip files

bcamper:
This PR adds the ability to load scene "bundles" from zip files. Any relative resources (imported scene files, images, fonts) are read from within the zip bundle.

The purpose of this feature is to improve distribution and management of complex scenes, particularly by allowing users to distribute a single self-contained file with images and/or fonts.

 **Scene Bundles**
A new `SceneBundle` class provides an interface for loading scene files. Given a scene URL to load, the scene bundle object loads that file, and resolves relative resources for any files loaded within the bundle. For "standard" scene YAML files (e.g. not zip bundles), this continues to work as prior scene loading has. The bundle stores a root URL for the scene, and prepends this root to any relative paths. If a resource has a remote URL, the bundle returns that as-is.

To support zip files, a `ZipSceneBundle` subclass adds additional behavior:
- When a scene zip file is loaded, the zip bundle first loads the zip file and decompresses all files within it, keeping an internal list of its resources.
- Zip files are determined by a `.zip` extension, other filenames will be treated as standard (non-zip) bundles.
- The zip bundle identifies a "root scene YAML" (*more on this below*), and loads and parses it as YAML, as a standard bundle would.
- To resolve relative resources, the zip bundle **creates a blob URL** (using `createObjectURL`) for the file's contents. This allows all bundles (standard and zip, and possible future types such as locally cached) to provide a common URL-based interface for retrieving any scene resource.
- As with non-zip bundles, remote resource URLs are passed through as-is.

Each imported scene (using `import`) is instantiated as its own scene bundle, which allows both zip and non-zip bundles to freely import further scenes of either type.

**Root Scene File**
**This is likely the only outstanding user-facing issue to decide before moving ahead with this feature**, though it is a significant one. When a scene zip bundle is loaded, we need to identify which YAML file *within* that bundle should be considered the "root". There are several approaches to this with varying levels of flexibility and complexity. My bias here is towards something that keeps things easy for the user (without becoming very complex to implement). The current behavior in this branch is:

- The root scene must be in the root folder level of the zip file, e.g. not within a sub-folder.
- The root scene file must **either** be a YAML file with the same name (minus extension) as the bundle itself (e.g. `bubble-wrap.yaml` for `bubble-wrap.zip`), **or** (if no file with that name is found) a single YAML file (with any name) at the zip bundle's root folder (e.g., a file named `bubble-wrap.yaml` would load for a bundle called `bubble-wrap-test.zip`, provided there were no other YAML files in the zip's root folder, though other scene imports could be in sub-folders).
- Using the **same-name-as-zip-file** is a straightforward rule to me that follows our existing convention for house styles: the repo and root scene yaml files are consistent, and all files in the repo could be zipped without modification (this is a key for me, not requiring the user to make modifications to their scene to allow it to be zipped).
  - However, this rule falls short in cases (that I encountered several times when testing this branch) in which I wanted to have multiple versions of the scene available for testing, e.g. `bubble-wrap-woff.zip` to test a version that loads WOFF fonts, etc.
- The best balance may be to simply keep the *other* rule instead, requiring that there be a **single YAML file in the zip's root level**. This puts some restrictions on authoring, but also fits all of our current house styles (including the new [Tron](github.com/tangrams/tron) structure), and is both easy to implement and to message to users in the case a root file is not found.
- Other options considered:
  - Require the root YAML to have a constant name such as `scene.yaml` or `root.yaml`. This is straightforward but I believe places too many restrictions / lack of compatibility with common scene patterns (I don't want to require people to rename their root scene -- or any other files -- before zipping).
  - Allow the root YAML to be in a sub-folder. Could still be combined with the "single file" rule, if the *first* sub-folder containing *any* YAML files must still have a single YAML. It may be hard to argue this complexity is worth it (and it's harder to explain), though it's worth noting that it's common, both for manually and automatically created archives to have their contents in a folder (for less messy un-archiving in-place). For example, github's [download links](https://github.com/tangrams/bubble-wrap/archive/gh-pages.zip) include the repo and branch name, e.g. `/bubble-wrap-gh-pages/bubble-wrap.yaml`. The current behavior can handle this case.
  - I experimented with parsing all scene files in the zip, and building a graph of their `import`s, to find the root. This gets complex quickly, especially in cases where relative paths may be used in imports (`import: ../other-folder/file.yaml`).

**JSZip**
Zip processing is done using [JSZip](https://stuk.github.io/jszip/), which appears to be the most robust JS-based zip library by far, and provides a modern Promise-based API. The primary downside is its large size: the current release minified is still 97k! JSZip provides full zip encode/write capabilities in addition to reading zip files. Since we don't need write capability, I have [forked the library](https://github.com/Stuk/jszip/compare/master...tangrams:read-only) and removed as much code as possible (we've followed a similar approach with js-yaml with good results). This reduces the minified library to **83k**, which is an improvement, though not as much as I had hoped. JSZip is designed with a single interface for reading/writing, which makes it a bit harder to remove code; it may be possible to reduce this further and warrants more investigation. There is an [outdated issue](https://github.com/Stuk/jszip/issues/170) regarding this that we should inquire on.

All told, with the read-only branch of JSZip, this branch increases the Tangram gzipped minified build to **153k**, up from 124k. While this is a substantial increase, I believe it's worth it for a feature that will enable significantly better management of increasingly complex scenes moving forward -- especially for new projects such as [Tron 2.0](https://github.com/tangrams/tron) and others that make heavy use of [Blocks](tangrams.github.io/blocks).

**Font Loading**
This branch also necessitated some changes to custom font loading:
- We now test for and use the native `FontFace` API for adding new font faces to the document. This API is long-supported in Chrome and Firefox, and will be supported in the upcoming Safari 10. Other browsers continue to fallback to using CSS `@font-face` injection.
- When fonts are loaded from zip files and the browser does *not* support the native `FontFace` API, the font data must first be converted to a base64-encoded `data` URL (which is used as the font source in the injected CSS). See the [fontloader polyfill](https://github.com/bramstein/fontloader/blob/master/src/fontface.js#L146-L153) for an example.
- There is a known issue with OpenType/TrueType fonts not loading properly from zip bundles on IE11 (but *do* load properly on Edge). WOFF fonts (much preferred) load on both IE11 and Edge. We may want to continue to investigate, but this is a reasonable limitation (the scene will still render with fallback fonts).

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "gl-mat4": "1.1.4",
     "gl-shader-errors": "1.0.3",
     "js-yaml": "tangrams/js-yaml#read-only",
+    "jszip": "3.0.0",
     "match-feature": "tangrams/match-feature#v1.2.0",
     "pbf": "1.3.2",
     "strip-comments": "0.3.2",
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "gl-mat4": "1.1.4",
     "gl-shader-errors": "1.0.3",
     "js-yaml": "tangrams/js-yaml#read-only",
-    "jszip": "3.0.0",
+    "jszip": "tangrams/jszip#read-only",
     "match-feature": "tangrams/match-feature#v1.2.0",
     "pbf": "1.3.2",
     "strip-comments": "0.3.2",
--- a/src/module.js
+++ b/src/module.js
@@ -33,6 +33,7 @@ import FeatureSelection from './selection';
 import CanvasText from './styles/text/canvas_text';
 
 import yaml from 'js-yaml';
+import JSZip from 'jszip';
 
 // Make some modules accessible for debugging
 var debug = {
@@ -66,6 +67,7 @@ if (Thread.is_main) {
     // Allows FontFaceObserver to use polyfill (without needing to include its own duplicate polyfill)
     if (window.Promise === undefined) {
         window.Promise = Promise;
+        JSZip.external.Promise = Promise;
     }
 }
 
--- a/src/scene.js
+++ b/src/scene.js
@@ -15,7 +15,7 @@ import TileManager from './tile_manager';
 import DataSource from './sources/data_source';
 import FeatureSelection from './selection';
 import RenderStateManager from './gl/render_state';
-import CanvasText from './styles/text/canvas_text';
+import FontManager from './styles/text/font_manager';
 
 import {Polygons} from './styles/polygons/polygons';
 import {Lines} from './styles/lines/lines';
@@ -981,7 +981,7 @@ export default class Scene {
         this.createDataSources();
         this.loadTextures();
         this.setBackground();
-        CanvasText.loadFonts(this.config.fonts);
+        FontManager.loadFonts(this.config.fonts);
 
         // TODO: detect changes to styles? already (currently) need to recompile anyway when camera or lights change
         this.updateStyles();
--- a/src/scene_bundle.js
+++ b/src/scene_bundle.js
@@ -25,9 +25,7 @@ export class ZipSceneBundle extends SceneBundle {
         this.zip = null;
         this.files = {};
         this.urls = {};
-
-        // Make root file name by substituting or .yaml extension to .zip file name
-        this.root = url.split('/').pop().split('.zip').shift() + '.yaml';
+        this.root = null;
     }
 
     load() {
@@ -37,11 +35,8 @@ export class ZipSceneBundle extends SceneBundle {
             return Utils.io(this.url, 60000, 'arraybuffer')
                 .then(body => this.zip.loadAsync(body))
                 .then(() => this.parseZipFiles())
-                .then(() => {
-                    let url = this.urlForZipFile(this.root);
-                    return Utils.loadResource(url);
-                })
-                .catch(e => { throw e; });
+                .then(() => this.loadRoot())
+                .catch(e => Promise.reject(e));
         } else {
             return Promise.resolve(this);
         }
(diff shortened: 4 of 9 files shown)


bcamper:
@blair1618 @nvkelso long description here but I would like comments specifically on the **Root Scene File** proposals/section, this is the major issue to resolve here.

The rest may be interesting in passing but is mostly JS implementation internals :)

blair1618:
A+ for scene bundling!

There's no perfect solution for the "root scene" convention, but I would consider a permutation of the rules you've suggested:
```
if (bundle contains "scene.yaml") {
  load "scene.yaml" as root scene
else if (bundle root directory contains a single yaml file) {
  load the single yaml file as root scene
} else {
  error
}
```
This does things that seem good to me:
 - The name of the zip archive is irrelevant, so you can rename archives for different versions or branches and it won't break
 - Allows all "house" styles to be zipped and loaded with no changes
 - Provides an escape hatch for disambiguation in case you _need_ to have multiple yaml files at the root (maybe for CI config or something)

If that third bullet doesn't seem super important then I would be in favor of requiring a single yaml file at the bundle root, for simplicity.

nvkelso:
+1 for Matt’s proposal.

bcamper:
Latest should include the new proposed scene root rules.

nvkelso:
Will one scene file bundle be able to load another scene file bundle? I'd like to start distributing Sprite packs this way for icons and "include" them into other scene files.

bcamper:
Yep!

tallytalwar:
@bcamper's reign of super descriptive PR continues :D.

I agree with @blair1618 suggestion, but I also liked the idea of having project name (zip file name) same as the *Root Scene File* name. So if I may suggest another addition to the suggested `if` block:

```
if (bundle contains "scene.yaml") {
  load "scene.yaml" as root scene
} else if (bundle contains yaml having the same name as the zip file/project file) {
  load this yaml (having the same name as the zip file)
} else if (bundle root directory contains a single yaml file) {
  load the single yaml file as root scene
} else {
  error
}
```

This also gives more flexibility to the user.

nvkelso:
+1

bcamper:
I'd like to avoid having more than two rules for determining root scene file (I don't think more justifies the additional logic and documentation/explanation to user for diminishing use case returns).

Having let this sit a week, I'm actually inclined to reduce this to just **one** rule, which would be:
**bundle root directory must contain a single yaml file**. 

Looking at the real-world use cases, I don't see where this would be a problem, and it fits the structure of our current house styles. It puts some limitation on the user's organization of scenes by requiring a sub-folder for any imported scenes that will be included in the bundle, but I don't think this is any more restrictive than requiring a specific scene file name (whether matching bundle or fixed to a constant name).

blair1618:
One rule sounds nice to me. 

When we say "a single yaml file", does that mean one file with the extension ".yaml"? The ".yml" extension is also common, though Tangram files have traditionally used ".yaml". I'm mostly thinking about ".travis.yaml" and "circle.yml" files, which I believe are required to be at the root level of a repository.

bcamper:
While it's not difficult to support both `.yaml` and `.yml`, I'd be comfortable (for simplicity all around) just requiring the `.yaml` extension. It's what we use in all our demos and docs. We could also consider adding (if needed) an explicit "format" parameter when loading scenes, to specify YAML vs. zip explicitly, rather than relying purely on file extension. For that matter, we could provide a means to specify an alternate scene root path within the zip explicitly. In both cases, I lean towards starting simple, and adding more options as needed based on usage/feedback.

bcamper:
Merging with single-YAML rule, we can tweak before the next release goes out if we want.

Issue #336, opened 2016-06-20
RFC: Articulated Labels


dmvaldman:
Currently our label placement algorithm tries to place straight labels on segment midpoints if they can fit. Often this is too restrictive, especially for curvy roads, because often there is no road segment where a label fits entirely. Two approaches are to:

#### Break labels on spaces

By splitting a label such as _"Washington Blvd"_ into two - _"Washington"_ and _"Blvd"_ - we can place the label on (internal) segment ends in addition to segment midpoints, where the label is split and follows the angles of the previous and next line segments. For labels that are three or more words, such as _"George Washington Blvd"_, we would still split them into a set of pairs ([_"George Washington"_, _"Blvd"_] and  [_"George"_, _"Washington Blvd"_]) and try to fit the possibilities in the set, taking the first that fits.

The GL geometry (for a split label) would consist of creating two quads offset from an anchor point. We could do this by generalizing the point builder to include an offset (to be applied before any rotation). We can also consider a polyline geometry which would generalize to the case of more than two quads (described below).

**Pros**:
- Simple
- Keeps typography integrity (kerning, ligatures, etc)
- Works for all languages (text shaping, etc)

**Cons**:
- Label placement is still geometry dependent. Can't place label anywhere along a curve, or space them out regularly, etc. 
- Doesn't generate as many new label placements as a "break on characters" approach.
- For curved lines made of several short segments, the label cannot "curve around" the line; it can only kink in one place.

#### Break labels on characters

By breaking labels on any character we can place labels arbitrarily. The placement function doesn't need to consider midpoints or ends of segments, but can simply place based on some arbitrary spacing.

The GL geometry for such a label would consist of several quads in a chain like a polyline (but with no joins or miter vectors. I think we would need to modify the line builder to work with this type).

**Pros**:
- Complete control over label placement, as we are not dependent on the underlying geometry.
- Can generate many more labels
- Can allow for realtime dynamic label placement on zoom (though this would require development work in other areas)

**Cons**:
- Can mangle the typography
- No clear way to extend to languages like Arabic that incorporate text-shaping

cc @bcamper @hjanetzek

bcamper:
[...]

PR #360, opened 2016-07-19
Articulated labels merged

dmvaldman:
I'd like to start a discussion about my work with articulated (curved) labels @bcamper @nvkelso @sensescape. Here's a summary of current progress. 

General curved (think bezier) labels would require special treatment for handling typographic considerations (kerning, ligatures, text-shaping, etc) and right-to-left (RTL) text support. For a first measure we've decided to break labels on spaces, which skirts these issues.  Try it out on [precog](https://precog.mapzen.com/tangrams/tangram/articulated-labels-merged/).

## Examples

Marin county

marin

The hills of LA

LA_hills

## Right to Left Text

Once labels are broken up and rendered in parts, one has to think about right-to-left languages. This PR has RTL support as well as "mixed-text" (mixed RTL and LTR in a label) support. 

kink-hebrew-kinked

kink-zenglish-arabic-english-mixed

There is one place where mixed-text label rendering can fail, however, and that is where a word of all numbers is between a strongly-LTR and strongly-RTL language (or a similar pattern of a "weak" word between two different "strong" words). Here's an example of such a string.

screen shot 2016-07-11 at 4 44 07 pm

This is quite rare, but needs mentioning.

## New Parameters

There are a few new parameters in the label placement process that I've exposed, and others that can be revisited:

### `line_exceed`

In places with windy roads, we can fit in more labels than before. Previously, we were using a parameter `line_exceed` in Tangram (modifiable in the scene file for Tangram JS, but not documented) which would allow for labels to be placed even if they exceeded the length of the road segment. This value was set to a very high value in order to get a good enough label placement. However, these labels could shoot off awkwardly of a curved road. We can now consider reducing this value without incurring as strong a penalty for label density. For example, here's the Yosemite valley with the value reduced to 50% (labels can be twice as long as their road segments).

line_exceed_reduced

### Multiple Placement

This PR also introduces a potential different strategy for label placement (not used in the images above). Previously, for every road geometry we would create one label. I've experimented with generating multiple labels per road (a straight label at every segment midpoint and a kinked label at every internal segment endpoint). This has the benefit of adding more labels, at the expense of doing more collision logic.

You can toggle this setting in the scene file, by adding a `placement: many`option under any `text` field used for labels.

### Articulated Label Spread

For kinked labels, there is a free parameter introduced which is the "spread" between the two kinked parts. I call this parameter the `spread_factor` and it is available in the scene file to fiddle with. I think it's best to choose a value that works and stick with it, but didn't think I should be the one to decide this value :-). It's easiest to explain this value in a picture

img_4622

The boxes are the bounding boxes of the kinked label. A spread of 1 shifts the labels along their axis so that their corners touch. This enforces that the labels could never overlap, however, they may be spaced too far away, aesthetically speaking. Currently the code defaults to 0.5.

You can toggle this setting in the scene file, by adding a `spread_factor: 1`option under any `text` field used for labels.

## TODO

- [x] Make "partially upside-down" labels choose an orientation based on their size (and angle?)
- [x] Handle outline textures correctly
- [x] "Smarter" label placement strategy for placing labels along a line. Should take into account high `line_exceed` factors and articulated vs straight label choice.
- [x] Add documentation

## Future Directions
- Break labels along other metrics, like on 3-letter combinations
- "Average" out road geometry so that labels fit better globally on a road (by looking ahead and behind multiple line segments) and not simply locally


bcamper:
Nice! Diagrams are helpful :)

For the new parameters, we should leave them exposed while we tune the
behavior, then consider if some/all should remain in the scene file, or
should instead be set to fixed values (which could vary by zoom or other
criteria we select). This also will depend on how relevant they are to ES,
so we can keep scene syntax as cross-platform as possible.

nvkelso:
Awesome work! The precog link really shows off the combination of better "articulation", multiple icon + text label placements, and shields :) Together these feel like a huge leap forward and reason enough to do another Tangram release. 

`spread_factor` I think a value of 0.5 like you propose is a good default, we can tweak once we get this tested in a house style and before the next Tangram release. 

I'm curious about **Multiple Placement**, can you add screenshot GIF for that, please?

`line_exceed` Now that we're labeling more features, I think dropping the `line_exceed` default down makes sense. We can play with this in a house style like Bubble Wrap to determine a good default. I anticipate allowing **more** overlap for major roads, and _less_ overlap for minor roads. **¿Is this measured in px or in % of total character width or?**

Can you talk me thru way some labels here are working but some not?

screen shot 2016-07-19 at 21 26 52

1. **JFK Dr** Why doesn't `John` and `Dr` "kink" down to the line more? _It feels like the angle is being calculated at the point nearest the line geometry instead of being averaged for the run (take the center of the run, find the line intersection, angle the run based on the average angle of the line between the start and end of the text run?)_
2. **Nancy Pelosi Dr.** I think reducing the `line_exceed` overlap would prevent this feature from labeling in JS? What impact would that have on ES?
3. **MLK Jr. Dr** Why is `Martin Luther` angled so far away away from the line geometry it's labeling? Doesn't appear to be kinking on all the whitespace?

Another set of examples ([link](https://precog.mapzen.com/tangrams/tangram/articulated-labels-merged/#40.73530896308074,-124.14633122089184,13.337416837082612)) showing angled text away from line feature:

screen shot 2016-07-19 at 21 44 27

Regarding `Make "partially upside-down" labels choose an orientation based on their size (and angle?)` I think preferring the longer run to be right side up is key – though some labeling systems also add in a "don't let too much of this label go upside down" failsafe parameter, too that when exceeded blocks the label from rendering. 

I'm obligated to ask... After articulated labels what's the big chunk of work needed for full curved (ala bezier) labels? Is it incremental or require another complete rebuild? Is there something that could get us there faster as an approximate intermediate step? 

_And finally, your blue map background is killing me in your demo style ;)_

nvkelso:
[DC example](https://precog.mapzen.com/tangrams/tangram/articulated-labels-merged/#38.893669673676676,-77.014576,12.016776438619237) where I'd expect label to be articulated:

screen shot 2016-07-19 at 21 53 11

nvkelso:
Example of upside down [Beach Drive in DC](https://precog.mapzen.com/tangrams/tangram/articulated-labels-merged/#38.943982406624976,-77.01875851542364,13.016776438619237):

screen shot 2016-07-19 at 21 54 23

dmvaldman:
Here's some replies to your feedback in the order it was given:

### Multiple Label Placement

In areas where the label density is low, multiple label placement gives a good improvement to label density

www gifcreator me_qnnut1

However in places with high label density, multiple label placement doesn't give much of an improvement. 

www gifcreator me_l44gdm

Having consistently better results with multiple label placement can happen with some further modifications to how we do our collision pass. In theory, adding a bunch more labels prior to the collision step should only improve the label density after the collision step, but our current collision implementation depends on the arbitrary order (within the same priority) the labels are collided in. This can have the effect that one long label can be placed at the expense of several other labels (that it obscures), by virtue of it being placed first. This would decrease the total number of labels on screen.

### Why are some labels bending the wrong way, or not bending enough?

The articulated label placement looks locally at the underlying road's line geometry. A kinked label would look at the previous line segment's angle, and the next line segment's angle, and that's it. It could be that the segments two/three/etc steps back/forward are at wildly different angles, and it wouldn't capture this longer range curvature. We could think about doing such a looking forward and backward and "simplifying" the line's geometry with an averaging process. When continuously zooming in/out within a zoom-level, this may have a jarring effect, but I haven't tested.

A quick fix is reducing the `line_exceed` parameter. In your "Martin Luther King Jr. Dr." example, that label isn't placed if we reduce the `line_exceed` value from its current default 80%, to just 77%.

This is also true for your next example in the DC area. Reducing the `line_exceed` to 60% gives

screen shot 2016-07-20 at 11 30 13 am

And reducing it further to 50% gives 

screen shot 2016-07-20 at 11 30 27 am

In general reducing `line_exceed` encourages using kinked labels, but often also reduces the label density in general.

### Why don't labels kink on every space, if they are made up of >2 words?

This would also require some kind of line geometry averaging step. Otherwise a kinked label word in the middle of the label would need to match very closely the length of the line segment it fits over. More discussion is [here](https://github.com/tangrams/tangram/issues/336).  I think this is a good direction to explore further to generalize the current work to more than 1 kink.

### Why are some labels upside-down?

Glad you asked! This is one of the TODO items. Though this was discussed in our Graphics meeting, for posterity I'll mention here that when a kinked label crosses the vertical axis on both segments, one half will always be right-side up and the other half "upside down". In your "Beach Drive NW" example, the "NW" is right-side up. 

I see two solutions: 

- We can be smarter about which side we choose to be right-side up, by always ensuring the larger part of the kinked label is right-side up. One side will still be upside down though.
- We throw out this kind of label before the collision process. The label may still be placed, just from a different candidate on the line.

dmvaldman:
I forgot to add, there's also an `articulated: true` parameter exposed in the scene file (under any `text` style) to toggle whether straight or mixed (kinked and straight) labels appear

dmvaldman:
All the new parameters, for reference. 

screen shot 2016-07-20 at 3 27 10 pm

Note, as Brett mentioned. These are for testing purposes, and we can consider exposing or fixing or configuring (based on zoom level, etc) any of them.

dmvaldman:
756ee49 Fixes upside-down label segments by simply throwing them away from out of the label generation process, and going the next available line segment to place the label on.

sensescape:
Thanks David! Amazing work! Thanks for all the detailed explanation : )

dmvaldman:
There was a problem with font stroking, where since stroke width is not taken into account when calling Canvas' `measureText` method, texture artifacts would appear when the label is split up, e.g.,

screen shot 2016-07-25 at 4 30 35 pm

This is "kind of" fixed in 44b0727. However, if the stroke width is made big enough (where there is no vertical line that can cleanly separate the words of a label), it'll still look bad:

screen shot 2016-07-25 at 4 29 58 pm
screen shot 2016-07-25 at 4 29 09 pm

The only way I can think to fix this generally would be to render each word of the label completely independently in the texture, and space them manually. This comes with a lot of added complexity though. Would like to know what you think @bcamper

bcamper:
Something in this branch (I think recent, maybe the changes to point label anchors or other commits around there?) appears to have introduced bugs with the collision of point labels (could be cases where `cull_from_tile` isn't working? or could be more widespread).

Here is this branch vs. master, note that the highway shields are overlapping themselves and other labels:

output_k5urcf

bcamper:
Further observation does suggest this is an issue specifically with point labels with `cull_from_tile: true` applied.

dmvaldman:
You're right Brett. 5a82ffb should fix the issue.

bcamper:
@dmvaldman thanks but the diff on that commit just seems to have some debug drawing in it?

dmvaldman:
:-( I added the wrong file. I've since fixed (reverted to previous commit and forced push) in ad76b22

bcamper:
I'd like post some comparison screenshots here when I have more time, but for now, thinking about `line_exceed`, there are different levels of quality for label placements, ranked something like:

- 1st: A straight label that fits 100% within the line segment.
- 2nd: A kinked label that fits 100% within its two line segments.
- 3rd: A label that exceeds the line segment(s), but is within the `line_exceed` tolerance (straight labels may always be preferable?)

I think a problem we are suffering from now is that we take the **first acceptable** placement we find (at least I think this is how it's working? please correct me if not). Instead, I think we ought to be "scoring" each label placement (keeping only those that meet the minimum acceptability requirements, e.g. fit within the `line_exceed`), so that we end up with higher quality overall. Something like:

- Try to place a straight label on all line segment midpoints.
  - If a label fits 100% within a segment, use this placement and go to next label.
  - If a label is longer than the segment but within the `line_exceed`, save it as a "candidate" placement. If a previous non-ideal candidate was already found, keep whichever one has the lower exceed factor (e.g. keep a reference to the best non-ideal candidate placement found so far).
- If no 100% placement found, next try all corner/kinked placements.
  - If a label fits 100% (each kinked segment fits within its line segment), use this placement and go to next label.
  - If one or both segments is longer but within the `line_exceed`, compare it to any previous candidate placements as described above. Perhaps kinked placements should be scored with more nuance, depending on whether one or both segments exceeded its segment's length.
- If all placements are exhausted, and a non-ideal candidate is available, use that placement. Otherwise, discard label.

It also may be reasonable to have different `line_exceed` values for straight vs. kinked labels, or based on other criteria (it's a % so presumably should already scale properly with line segment and zoom levels).

Basically, I am concerned that if we lower the `line_exceed` too far, we are going to lose too many labels, *but*, if we were preferring kinked labels that fit, we could keep a higher `line_exceed` because those placements would still be better than nothing (IMO). We might still lower the overall exceed value, but I think we can higher quality placements overall with a scoring system.

bcamper:
Separate topic and thinking ahead to "more than 1 break / 2 label segments": the problem as I understand it is that by only breaking on *word boundaries*, we are predetermining the **shape of the label**, and then trying to find a way to **fit that label shape** onto the shape of the line. This is why it's impractical for us to place labels with multiple kinks, since it's so unlikely that the line shape *just happens* to fit 3 or more word lengths.

Really the proper way to do this is to **make the label conform to the line**: first calculate **the shape of the line**, and then use this to determine **where the label string should be broken** (e.g. not always on a space). While it would be expensive to send the entire line geometry to the main thread (at least I expect so), perhaps it would be reasonable to send the *lengths of each line segment*, which would allow the placement phase to occur on the main thread and more appropriately determine the best sub-strings to render.

I know I have brought up concerns about interfering with text shaping in the past, hence the suggestion to only break on spaces. But, given the RTL detection code we now have, I think we could very reasonably apply a modified process to those labels, either requiring them to be broken on spaces (maybe constraining them to something like the current 1-kink approach), or simply disabling kinking and requiring them to be placed straight.

Side note: we already calculate line segment lengths in the polyline builder. If we stored these values for a given feature geometry, other processes like label placement could make use of it and we could avoid duplicate work (or vice versa, a label might be placed first and the line segment lengths could be cached for use by the polyline builder).

bcamper:
Sorry one more :)

I do think we need to properly address the issue with "stroke bleed" between words. But I think following the conclusion above, it probably already makes sense to shift towards rendering separate text strings (whether they be separate words, or substrings or words that have been broken to fit a line). It just seems like we will need this flexibility for a few reasons.

dmvaldman:
We can definitely do some sort of improved label placement strategy. Previously in the code, we only put one label down for a line, but we've generalized this [here](https://github.com/tangrams/tangram/blob/articulated-labels-merged/src/styles/text/text.js#L191-L196) to iterate through the line and place a label at every midpoint and interior join. This could be made less naive, and take into account a metric for "goodness" of fit.

In your second point, you're absolutely right that we're fitting the geometry to the label, as opposed to the label to the geometry. We can do the latter, as you say, by splitting a label on characters (or groups of characters) instead of words. However, this does make us need to support bidi text in its entirety, as my simpler single-pass solution won't work in general. It will also break some typographic qualities like text-shaping, ligatures, kerning etc. I propose experimenting with fully curved text for strictly LTR or RTL alphabets, and adopting the kinked strategy for bidi-text.

For the main thread label rendering, we could avoid sending geometry data by just sending all the collision boxes (and texture boxes?) instead. No line geometry computation needs to be done unless you want to adaptively add new labels. If we send all the labels (or all the labels in consideration for collision), then I believe that's all we'd need to send. Maybe I'm wrong here?

I agree that stroke bleeding should be solved. I guess one of the particular questions I have is what to do for a straight label: is it rendered as two (or more) quads?

bcamper:
Cool - having the code generalized for iterating over different placement options is great. Can we try some strategies as discussed above -- preferential/scored, e.g. "best" rather than "first good enough" -- for this branch? I'd like to see the impact on quality of placements. I think right now that a high `line_exceed` is causing some labels to prematurely place on a straight line, when a better fitting straight or kinked placement may exist further on the line. But I'd prefer not to "solve" that just by lowering the `line_exceed` because I think it may drop too many labels.

> I propose experimenting with fully curved text for strictly LTR or RTL alphabets, and adopting the kinked strategy for bidi-text.

👍 Agree this is a good strategy. It's fine to have RTL/bidi go through a different path. I'm very interested in seeing a more general solution, but may make sense to wait till this branch ships since it's already quite long-running. (Let's see what it looks like with updated placement strategies discussed above.)

> For the main thread label rendering, we could avoid sending geometry data by just sending all the collision boxes (and texture boxes?) instead. No line geometry computation needs to be done unless you want to adaptively add new labels. If we send all the labels (or all the labels in consideration for collision), then I believe that's all we'd need to send. Maybe I'm wrong here?

My thinking here was related to the initial text rendering phase: if we wanted to find the **ideal** fit of the text to the line, we wouldn't know where to break the text (or how many substrings to break it into) without knowing the shape of the line (if not full geometry, maybe just line lengths/angles?). Since we can only measure text on the main thread (unless we go all-in on something like opentype.js, a much bigger change), it seems we need some form of this data within `CanvasText`? We could pre-break the text by groups of 3 chars (or something like that) which would be *better* than the current break-on-space approach, but would still be "fitting the line to the label" (just with more resolution). Does this make sense? Definite pros/cons with any approach.

Once the actual bounding/collision boxes have been calculated, yes, it should be fine to just have those on the main thread.

bcamper:
Sorry, missed this:

> I agree that stroke bleeding should be solved. I guess one of the particular questions I have is what to do for a straight label: is it rendered as two (or more) quads?

While it's desirable to minimize the number of quads, it seems it will be difficult to do that for labels that are potentially articulated, since we won't know if they are straight or kinked until the placement phase in the worker, right? (To have multiple words in a single quad, they'd have to be rendered side by side in Canvas like we do today.)

It's probably OK to have those straight labels render as multiple quads, since there is often an order of magnitude fewer text features compared to surrounding roads, buildings, etc. 

But we could still continue to render labels that don't even attempt to articulate (e.g. for points/POIs) as a single quad, right? This will also avoid having to change the logic for multi-line labels / line wrapping (currently all lines get rendered into one texture and are drawn as one quad).

dmvaldman:
To your last point, yes, the only text features that go through the articulation path are road labels. I included this logic [here](https://github.com/tangrams/tangram/blob/articulated-labels-merged/src/styles/text/text_settings.js#L44) and [here](https://github.com/tangrams/tangram/blob/articulated-labels-merged/src/styles/text/canvas_text.js#L109) (I think this excludes POIs...I'll double check).

bcamper:
FYI I just re-merged master into this branch to keep it from getting behind. No actual conflicts just a couple minor things.

nvkelso:
Just clarifying that this is limited to LINE geometry types, including
river lines and boundary lines (not just roads).

dmvaldman:
The text that can currently articulate has geometry type `LineString` or `MultiLineString`. Is that a necessary and sufficient condition?

nvkelso:
Yes and yes :)

dmvaldman:
There's a shiny green button on here now. The additions are:

+ **Stroke fix** - each word of the label is now it's own quad, rendered separately and drawn on its own line in the texture atlas. We also cache each word so no words are drawn twice unless they have different styles. This will also set the stage for breaking up labels on more general patterns than on spaces (for instance, 3 letter groups)

+ **Placement strategy** for articulated/straight labels that's more robust than our previous `line_exceed` heuristic. It cascades between an optimistic straight placement, then optimistic kinked placement, then "good enough" straight placement, then "good enough kinked" placement, or bust.

There are no longer scene file hooks that there were previously as I believe we've achieved sensible defaults, and exposing them as parameters would be more confusing than helpful. If you want to fiddle, though (especially with the label placement strategy), the place to do so is here:

+ [label placement strategy](https://github.com/tangrams/tangram/blob/articulated-labels-merged/src/styles/text/text.js#L219-L227) and [default quality cutoffs](https://github.com/tangrams/tangram/blob/articulated-labels-merged/src/labels/label_line.js#L11-L12)
+ [maximum articulated angle](https://github.com/tangrams/tangram/blob/articulated-labels-merged/src/labels/label_line.js#L10)
+ [spread between articulated labels](https://github.com/tangrams/tangram/blob/articulated-labels-merged/src/labels/label_line.js#L27)

dmvaldman:
Would be good to get a review on this PR @bcamper @nvkelso. [Precog link](https://precog.mapzen.com/tangrams/tangram/articulated-labels-merged/)

bcamper:
Thanks, I'm seeing noticeably better results with the most recent settings. Will give a more thorough test-drive over the next couple days with the aim to merge.

nvkelso:
In Oakland I'm seeing significant improvements! 

But San Francisco's Golden Gate Park is missing most road labels at most zooms (admittedly they are super long names). 

I'll do more testing and add screenshots soon.

_n

nvkelso:
I pointed Bubble Wrap at the precog Tangram from this branch and have screenshots! 

But I'm seeing significantly LESS labels on major roads than before (curved and not curved), even at high zooms like z18 where I was expecting to see just as many labels.

At many zooms I think we'll still want some additional amount of "overrun" that the text is able to extend past the road geometry (for major roads, not minor roads) that is more than default used for minor roads and other line labels. Is that configurable by layer in the scene file now?

z12 in San Francisco:
z12

z13 in San Francisco:
z13

z15 in San Francisco:
z15

z16 in San Francisco:
z16

z18 in San Francisco:
z18

dmvaldman:
Thanks for the in-depth comparison shots! I'll have a closer look why certain labels are being dropped. In theory we could have different tolerances for major/minor roads, zoom dependence would be trickier.

bcamper:
@dmvaldman can we try putting the allowed max overrun for straight labels back to the value it has in v0.9.x, so kinked labels are still preferred but we allow these? Then see if you can recreate these views please.

@nvkelso I'd like to avoid putting the "overrun" and similar values in the scene file for now, because they are tightly coupled to the current JS internal implementation for label placement, so they don't directly transfer to ES (the usual bar for adding scene file syntax is if we at least expect/plan to support a parameter in both platforms), and they may change as we keep improving behavior in the next few releases. But, we should definitely make sure we aren't shipping this with values that remove (too many) existing labels.

bcamper:
Looks like there is something off with left-aligned text:

http://tangrams.github.io/differ/?1=https%3A//github.com/tangrams/differ-tests/blob/gh-pages/unit-tests/text/text01.json&2=&lib1=0.9&lib2=https%3A//precog.mapzen.com/tangrams/tangram/articulated-labels-merged/dist/tangram.debug.js&ignore1&ignore2&go

We should also look at why fewer labels are placed on some of the line label tests in there.

(You can safely ignore those fringe-y differences where the text appears in the same place but there's a small diff anyway -- it's an alpha issue in the pixel diff utility that @meetar is going to look into.)

bcamper:
Here's a text anchor example (actually looks like some other anchors are off too, besides left):

screen shot 2016-08-31 at 11 28 12 am

dmvaldman:
These differ tests are great! Didn't know we had such encompassing unit tests. Will look into this more. Thanks.

dmvaldman:
Turns out the labels are shifted in these differ tests because they overlap tile boundaries and are moved into the tile. Setting `cull_from_tile : false` makes the differ tests the same. 

This raises the question of what would we like the default behavior to be in this case?

nvkelso:
Should they move to the other side of the point (icon) instead of just
“into the tile” if that toggle is true?

dmvaldman:
In general that's what will happen, but in this case the user has specified an anchor position for the label in the scene file (top-left, etc)

nvkelso:
Then should the entire label fail?

dmvaldman:
Yeah it should fail. It isn't failing in master (it's placing over a tile boundary). @bcamper and I were talking about it this morning and we think it's a good idea to remove the `move_into_tile` option completely (for point labels they don't make sense, and for line labels that can shift along a line geometry the option should be implied). This will be a separate pull request though that removes the `move_into_tile` logic and invalidates point labels that overlap tile boundaries.

dmvaldman:
@nvkelso pointed out a significant bug that caused many straight labels to be dropped. It was fixed in 9d43298. Here are some recent screenshots comparing master and this branch. I've lowered the `line_exceed` parameter from master, so still some straight labels are not appearing as they were before, but overall I think the placements are better. There's also a small spacing difference between many labels. I'm not exactly sure what it's attributed to yet.

z12 San Francisco
z12

z13 San Francisco
z13

z14 San Francisco
z14

z15 San Francisco
z15

z17 Glenwood Cemetery
z17

nvkelso:
@dmvaldman Screenshots look much better! Couple questions:

1. How can I try this myself? Loading off `https://precog.mapzen.com/tangrams/tangram/articulated-labels-merged/dist/tangram.debug.js` doesn't seem to give me the same results as in your screenshots.
2. Can you include more examples in Golden Gate Park that match my earlier screenshots, please?

dmvaldman:
Not sure why the precog link wouldn't give the same results... but here are the images you requested. It can be a little hard to tell which is the articulated one and which is master (sorry I should have marked them up before making GIFs but I have to run for the day!)

z15 (When Great Hwy (towards left) faces the toward the water, that's master)
z15-park

z16 (when cabrillo on the upper left is cut off, that's articulated)
z16-park

z17 (when 10th ave appears on the upper left, that's articulated)
z17-park

bcamper:
@nvkelso what tile version are you using? We found that some of your earlier shots were on 1.0, which didn't match ours...

nvkelso:
@bcamper Indeed, I'm on v1 tiles instead of v0.10 tiles! I'll try again... New pictures are looking much better!

bcamper:
Given the apparent similarity of label density and the verrrry long-running branch here, I'd like to get this merged and continue to make behavior adjustments (which will likely be an on-going process) on master.

@dmvaldman I did notice in some quick benchmarks that this branch is noticeably slower in generating labels than master. This is unfortunate but not really surprising due to the increase in Canvas `measureText` calls and additional geometry (i.e. more than one quad for some labels). That said I didn't drill down deeply enough to identify if those are the main culprits. I think we should do a follow-up pass (after the v0.10 release) where we look at identifying bottlenecks and optimizing.

nvkelso:
+100!

dmvaldman:
I'll take a closer look at the performance differences as well.

nvkelso:
OMG branch title now matches reality, party!

PR #367, opened 2016-07-27
Build tiles incrementally

bcamper:
This changes the tile build process so that groups of styles can be sent from the worker to the main thread for display as they are done, rather than blocking on all styles completing before any of the tile is shown.

It can be tried here: https://precog.mapzen.com/tangrams/tangram/incremental-build/

The primary motivation for this is to reduce the **time to initial visibility** of a tile, especially because icon and text styles can have longer build times due to text rendering and collision steps.

For this reason (and after some experimentation), I have split styles into two groups: **collision styles** and **non-collision styles**. When all of the styles in a group finish building, they are sent to the main thread where their GL meshes are instantiated and they begin rendering. *Usually* this means the non-collision group will finish first, but it's possible for them to finish in any order (for instance, it's possible a very dense network of lines could take a long time to tesselate).

This also adds some **fade in** logic for points and text styles, to ease the visual transition when new data comes in. Currently, fade in is only applied when those styles are not replacing previously visible ones (e.g. which will continue to be displayed as proxy tiles while the new data loads). This is to avoid a flicker effect where text/icons that were already present "re-fade-in". It's difficult to avoid this by fading some features but not others because there is no knowledge across tiles that individual features were previously present. I know we have put considerable effort into similar issues for ES (@karimnaaji), and more can be done here to improve, but it's likely a bigger project so deferring for now.

Lastly, the `WorkerBroker` handling of transferable objects has been generalized, so that a new `withTransferables` wrapper object can be used to transfer objects as part of *method calls*, in addition to via *return values* (only the latter was previously supported, just due to lack of need).

### DIFF
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -466,7 +466,7 @@ Texture.getInfo = function (name) {
 // Called from worker, gets info on one or more textures info from main thread via remote call, then stores it
 // locally in worker. 'textures' can be an array of texture names to sync, or if null, all textures are synced.
 Texture.syncTexturesToWorker = function (names) {
-    return WorkerBroker.postMessage('Texture.getInfo', [names]).
+    return WorkerBroker.postMessage('Texture.getInfo', names).
         then(textures => {
             for (var tex of textures) {
                 Texture.textures[tex.name] = tex;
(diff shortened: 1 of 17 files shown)


bcamper:
Verified no visual changes in Differ (with addition of debug flag to disable fade-in for consistent comparisons).

PR #371, opened 2016-07-29
Label boxes

dmvaldman:
Adds a `debug_settings` file (I suppose this could be a JSON file, instead of JS, but es6 has no built in support for importing JSON files) for viewing label collision boxes and texture boxes. Likely this debug settings file will grow with more flags over time.


bcamper:
This is good, except we should default them off ;)

I also realized that we should find a way to draw collision boxes for the `points` style (used for icons/sprites/etc., the base class of `text`). But this will need to be a little different, since there is no Canvas drawing step to add the box to. It could probably be added to the shader, since (as far as I recall) the collision box of a sprite should be the same as its visible area (not true for text labels). (I suppose we might consider changing that behavior in the future too though! :)

We should just merge this one first though to get the benefits for text, where it's more needed anyway.

dmvaldman:
Set default label box visualizations to false and migrated `profile_geometry_build_time` into the debug settings in b4e6994

bcamper:
Actually on closer examination, I don't think that profile option made sense as a flag, because it's more of a one-time operation (not something you need to turn on/off). So I moved it to a method call option instead: https://github.com/tangrams/tangram/pull/371/commits/afdb0fa62d584b60441b44edc587d8207acf0cd3

We will have need of more debug flags in the future though so I like the pattern!

PR #378, opened 2016-08-11
Remove link to OSM US hosted vector tiles

pnorman:
This service is shutting down and recommending people use Mapzen's hosted one


meetar:
Thanks Paul! Will merge this as soon as we sort out its failing CircleCI test :/

pnorman:
Can you rerun master? I **really** can't see why this change would cause a failure, so suspect that master is broken.

meetar:
It's nothing related to this change – CircleCI is having issues of its own, I'll run this past our engineering team as soon as they're in.

bcamper:
Yeah just a Circle/Sauce Labs hiccup (happens frequently unfortunately).

PR #394, opened 2016-08-24
Data sources can specify min/max display zooms and bounding box

bcamper:
This adds the ability for data sources to control when tiles are requested and displayed, with three new parameters:

- `min_display_zoom`: tiles will not be requested or displayed *below* this zoom
- `max_display_zoom`: tiles will not be requested or displayed *above* this zoom
- `bounds`: tiles will not be requested or displayed *outside* of this bounding box, specified as latitude/longitude in `[w, s, e, n]` format.

I think the behavior is straightforward and the primary discussion to be had here is around the **terminology**:

- Because of our existing `max_zoom` parameter, which controls overzooming (tiles will not be requested above this zoom, but they *will* be displayed, with tile geometry scaled up), I added a `display` qualifier in these parameter names. We could consider other naming as well.
- The `bounds` are currently specified as a single, flattened 4-element array of lat/lng values, in the order `[w, n, e, s]`. This was largely arbitrary, and we can easily use another format if desired. I considered using sub-arrays like `[[w, n], [e, s]]` to better show the coordinate groupings, but I felt this might be more error-prone than a flat list. _**Update**:_ based on discussion below, changed to `[w, s, e, n]` order._

Example:
```
    mapzen:
        type: TopoJSON
        url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
        min_display_zoom: 9
        max_display_zoom: 18
        bounds: [-74.1274, 40.5780, -73.8004, 40.8253] # [w, s, e, n]
```

### DIFF
--- a/src/scene.js
+++ b/src/scene.js
@@ -838,11 +838,11 @@ export default class Scene {
             });
         }
 
-        // Mark sources that generate geometry tiles
+        // Mark sources that will generate geometry tiles
         // (all except those that are only raster sources attached to other sources)
         for (let layer of Utils.values(this.config.layers)) {
             if (layer.data && this.sources[layer.data.source]) {
-                this.sources[layer.data.source].geometry_tiles = true;
+                this.sources[layer.data.source].builds_geometry_tiles = true;
             }
         }
     }
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -170,6 +170,9 @@ export class NetworkSource extends DataSource {
 
     _load (dest) {
         let url = this.formatUrl(this.url, dest);
+        if (!url) {
+            return Promise.resolve(dest);
+        }
 
         let source_data = dest.source_data;
         source_data.url = url;
@@ -221,6 +224,16 @@ export class NetworkTileSource extends NetworkSource {
         super(source);
 
         this.tiled = true;
+
+        if (Array.isArray(source.bounds) && source.bounds.length === 4) {
+            this.bounds = source.bounds;
+            let [w, n, e, s] = this.bounds;
+            this.bounds_meters = {
+                min: Geo.latLngToMeters([w, n]),
+                max: Geo.latLngToMeters([e, s])
+            };
+        }
+
         this.url_hosts = null;
         var host_match = this.url.match(/{s:\[([^}+]+)\]}/);
         if (host_match != null && host_match.length > 1) {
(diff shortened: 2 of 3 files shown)


bcamper:
cc @nvkelso for terminology feedback

blair1618:
If both `max_zoom` and `max_display_zoom` are specified, which takes precedence? i.e. for
```
mapzen:
  max_zoom: 16
  max_display_zoom: 15
```
Will tiles at zoom 16 be requested or displayed? I would guess "no", but just want to check. 

I like the flat-list `bounds` syntax, though I would've expected the order to be `[w, s, e, n]` (i.e. from lower to higher values).

nvkelso:
LGTM!

bcamper:
@blair1618 you're right, in your example:

```
mapzen:
  max_zoom: 16
  max_display_zoom: 15
```

...no tiles would be requested at z16 because `max_display_zoom` would short-circuit the process first.

I'm good with `[w, s, e, n]` instead, makes sense.

nvkelso:
Looks like desktop GIS and web mapping tools do several different things:

* [GDAL](http://www.gdal.org/ogr2ogr.html) [xmin ymin xmax ymax]
* [Leaflet](http://leafletjs.com/reference.html#latlngbounds) L.latLngBounds(southWest, northEast); and map.fitBounds([ [40.712, -74.227],  [40.774, -74.125] ]);
* [Google Maps](https://developers.google.com/maps/documentation/javascript/reference#LatLngBounds)  (sw?:LatLng|LatLngLiteral, ne?:LatLng|LatLngLiteral)

So both Leaflet and Google Maps are doing [ymin xmin ymax xmax] eg `[s, w, n, e]` sorting on lat lng instead of x y.

I'm personally more used to GDAL, but since Tangram sits tightly in the Leaflet + Google Maps on native Android space seems like we should follow their lead?

bcamper:
We already prefer longitude before latitude in other places like [camera `position`](https://mapzen.com/documentation/tangram/cameras/#position), so I'd prefer to stick with that. It's more "graphics-y" (x before y), and on the maps side it seems muddled anyway rather than there being a hard and fast rule (Mapbox GL is also longitude-first). For native APIs (outside the scene file), there may be object constructors that allow different patterns as well.

nvkelso:
Internal consistency is better than world consistency ;)

bcamper:
I think we're good here, merging - ping if any further issues.

PR #395, opened 2016-08-24
Add ability to filter by screen area

bcamper:
This adds an option to filter feature properties by **zoom-independent screen area**, that is, by the number of square pixels they cover.

This is done through a new `px2` unit type that can be applied to range filters, e.g.:

`filter: { area: { min: 500px2 } }`

This example filters the feature's `area` property by the number of *square mercator meters* that cover a *500 pixel screen area* at the current zoom level. This means that the `area` property must be in *square mercator meters*, as the property provided by Mapzen vector tiles is.

As with other pixel-based values in the scene file, the `px2` units are expressed in *logical pixels* (or "CSS pixels"), meaning they are interpreted at a pixel density of 1, and are automatically scaled up for higher density displays.

This syntax can be used to significantly simplify more cumbersome per-zoom filters such as:

```
filter:
  - { $zoom: { min: 9 }, area: { min: 100000000 } }
  - { $zoom: { min: 10 }, area: { min: 10000000 } }
  - { $zoom: { min: 12 }, area: { min: 1000000 } }
  - { $zoom: { min: 15 }, area: { min: 10000 } }
```

The new range filter units are supported by a new version of the `match-feature` module, which enables range filters to be transformed to account for per-zoom scale:
https://github.com/tangrams/match-feature/commit/9152e1711ccf3b4278d5116b7095bc79515c0304

**Note:** This PR does *not* add an automatically derived area property for polygon features. That is, this filtering can only currently be applied if the data source already contains a suitable area property (it does not need to be named `area`, as any property name can be specified in the filter, but it must already exist in the data source). A future extension could be to add the option to derive an area property for each feature, but this could add considerable overhead (and isn't necessary for all data sources), so would likely be an opt-in flag on the data source.

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -214,10 +214,11 @@ layers:
 
     landuse:
         data: { source: mapzen }
+        filter:
+            - { $zoom: { min: 16 } }
+            - { area: { min: 500px2 } }
 
         areas:
-            filter:
-                function() { return $geometry === 'polygon' && $zoom >= feature.min_zoom }
             draw:
                 polygons:
                     order: global.feature_order
@@ -248,16 +249,8 @@ layers:
 
         labels:
             filter:
-                $geometry: point
                 label_placement: yes
                 kind: [park, forest, cemetery, graveyard]
-                any:
-                    # show labels for smaller landuse areas at higher zooms
-                    - { $zoom: { min: 9 }, area: { min: 100000000 } }
-                    - { $zoom: { min: 10 }, area: { min: 10000000 } }
-                    - { $zoom: { min: 12 }, area: { min: 1000000 } }
-                    - { $zoom: { min: 15 }, area: { min: 10000 } }
-                    - { $zoom: { min: 18 } }
             draw:
                 icons:
                     sprite: tree
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "gl-shader-errors": "1.0.3",
     "js-yaml": "tangrams/js-yaml#read-only",
     "jszip": "tangrams/jszip#read-only",
-    "match-feature": "tangrams/match-feature#v1.2.0",
+    "match-feature": "tangrams/match-feature#v1.3.0",
     "pbf": "1.3.2",
     "strip-comments": "0.3.2",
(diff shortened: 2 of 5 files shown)


bcamper:
cc @nvkelso for comment and @blair1618 for ES considerations

nvkelso:
👍

blair1618:
Such concision! 👏 

This does require a bit of new wiring for filters, since range filters values haven't allowed a unit suffix before now. Still, it shouldn't be terribly bad to implement on ES, and it'll be worth the work. 

Does this apply only to range filters? Filtering by exact area isn't likely to be useful, but I'm just thinking about consistency for edge cases.

bcamper:
@blair1618 I agree it's a bit asymmetrical for `px2` units to only apply in this specific case, but I'm actually not sure there's a clear way to apply them elsewhere anyway: an "exact comparison" property match (besides being very unlikely for area as you mention) wouldn't work because there's ambiguity as to whether it's a numeric value with units attached, or a string value (not such ambiguity exists for range comparisons since it's an explicitly numeric value, which is also true of other contexts where units are applied). I'd also be hesitant to add overhead for a check that seems to have dubious real-world use. While this isn't totally satisfactory either, a user who wanted to do an exact comparison *could* write a function filter using `$meters_per_pixel` to do this.

bcamper:
@blair1618 any further thoughts? I'd like to merge this, but I think any further changes would likely be additive rather than significantly changing this implementation

blair1618:
Maybe what's weird is that we're putting units on a value that's being compared to feature values that don't have units. Instead of adding a unit suffix it might be more consistent to create variations on the "range" filter functions, like `min_px2` and `max_px2`, which perform comparison with against a feature value multiplied by a factor of square-pixels-per-meter.

bcamper:
Hmm... I get the distinction, but it feels like it makes for more awkward / less expected (from my perspective anyway) scene syntax. All application of units is pretty context-sensitive already, and `min` and `max` explicitly expect numbers already.

blair1618:
Well ok, I don't feel that strongly about the difference. Let's move forward with the "px2" suffix.

PR #396, opened 2016-08-25
Tall labels

dmvaldman:
Allows for a `max_lines` parameter for text fields to truncate excessively tall labels. The current default is 5 lines. This would require documentation if implemented.

### DIFF
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -302,7 +302,7 @@ class MultiLine {
         this.height = 0;
         this.lines = [];
 
-        this.ellipsis = '…';
+        this.ellipsis = '...';
         this.ellipsis_width = context.measureText(this.ellipsis).width;
 
         this.max_lines = max_lines;
@@ -388,9 +388,9 @@ class MultiLine {
         let line = multiline.createLine(line_height);
 
         // First iterate on space-break groups (will be one if max line length off), then iterate on line-break groups
-        for (let w=0; w < words.length; w++) {
-            let breaks = words[w].split('\n'); // split on line breaks
-            let new_line = (w === 0) ? true : false;
+        for (let i = 0; i < words.length; i++) {
+            let breaks = words[i].split('\n'); // split on line breaks
+            let new_line = (i === 0) ? true : false;
 
             for (let n=0; n < breaks.length; n++) {
                 if (!line){
@@ -406,7 +406,8 @@ class MultiLine {
                 let spaced_word = (new_line) ? word : ' ' + word;
 
                 // if adding current word would overflow, add a new line instead
-                if (text_wrap && line.exceedsTextwrap(spaced_word)) {
+                // first word (i === 0) always appends
+                if (text_wrap && i > 0 && line.exceedsTextwrap(spaced_word)) {
                     line = multiline.advance(line, line_height);
                     if (!line){
                         break;
@@ -425,7 +426,7 @@ class MultiLine {
                 }
             }
 
-            if (w === words.length - 1){
+            if (i === words.length - 1){
                 multiline.finish(line);
             }
         }
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -93,6 +93,7 @@ export default class CanvasText {
         // First iterate on space-break groups (will be one if max line length off), then iterate on line-break groups
         for (let w=0; w < words.length; w++) {
             let breaks = words[w].split('\n'); // split on line breaks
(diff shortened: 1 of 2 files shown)


dmvaldman:
Feel free to rebase all these :-P

bcamper:
Do you expect this will merge cleanly with the articulated labels branch? :)

src/styles/text/canvas_text.js
```diff
@@ -343,3 +342,107 @@ CanvasText.font_size_re = /((?:[0-9]*\.)?[0-9]+)\s*(px|pt|em|%)?/;
...
+
+// Private class to arrange text labels into multiple lines based on
+// "text wrap" and "max line" values
+class MultiLine {
+    constructor (context, max_lines = Infinity, text_wrap = Infinity) {
+        this.width = 0;
+        this.height = 0;
+        this.lines = [];
+
+        this.ellipsis = '...';
```
bcamper: I wonder if we should use the Unicode single-character ellipsis? I suppose font support might be iffy, so maybe not.

bcamper:
I haven't confirmed in-code, but it looks like there's a space being left in before the ellipsis?

screen shot 2016-08-29 at 11 16 37 am

dmvaldman:
I can do the merging if there's a conflict.
The space before the ellipsis is applied can be easily removed. Would you prefer that?

dmvaldman:
Also, nice debugging spot!

bcamper:
@dmvaldman re: ellipsis space, yes I think I'd prefer to not have a leading space, thanks.

bcamper:
Seeing some weird behavior with low `text_wrap` values, in this case `5`. Maybe anything shorter than the initial word?

tangram- 18 4488 40 7513 -73 9693_2016-08-29_11 23 22

dmvaldman:
Is the unexpected behavior you're seeing the artificial vertical space that seems to be added when the first word is larger than the allotted spacing?

bcamper:
Right - it appears to be inserting an empty leading line?

src/styles/text/canvas_text.js
```diff
@@ -351,8 +351,9 @@ class MultiLine {
         this.height = 0;
         this.lines = [];
 
-        this.ellipsis = '...';
+        this.ellipsis = '…';
```
bcamper: Sorry, I probably shouldn't have suggested this at all. We really should research handling across fonts more in depth before changing. I suspect it may not exist in some custom fonts and could just render with unexpected results?

bcamper:
I can't currently see any effect from changing the `max_lines` value at run-time (e.g. in Play, though theoretically a user could programmatically update this as well or hook it up to another GUI). 

I suspect this is because `max_lines` is not part of the `TextSettings` key, so `CanvasText` is caching text strings with different `max_lines` values under the same key.

dmvaldman:
Should be good now. Reverted the ellipsis to basic text, fixed the spacing bug, placed the `max_lines` into the `text-settings` key generation algorithm, and refactored a bit more.

bcamper:
This is looking good to me.

We should add test cases for `max_lines` to our Differ unit tests. Here's current 0.9.x vs. `tall-labels` branch:

http://tangrams.github.io/differ/?1=https%3A//github.com/tangrams/differ-tests/blob/gh-pages/unit-tests/text/text01.json&2=&lib1=0.9&lib2=https%3A//precog.mapzen.com/tangrams/tangram/tall-labels/dist/tangram.debug.js&ignore1&ignore2&go

The test cases are defined here: https://github.com/tangrams/differ-tests/blob/gh-pages/unit-tests/text/text01.json

@meetar can help walk you through how to add them!

bcamper:
Good call on factoring out the multi-line handling too :)

bcamper:
Merging but please leave any notes about new differ units tests there.

dmvaldman:
Added some unit tests to the differ-tests repo here https://github.com/tangrams/differ-tests/commit/7e0f85ecfafcb804821bf150a4e6b0cecabb37a2

Here are some results

dmvaldman:
The `max_lines` text attribute should also be specified in the docs for v0.10. Should I handle this or would you like to @meetar

bcamper:
@dmvaldman nice! I added the `needs-docs` label, and then we clear those out when we do the actual release. You could add docs for this in a branch, and then we can merge it when v0.10 is released (or just write it after we release for a small feature, I often do that too :)

bcamper:
(After each release, we go through all the issues tagged `needs-docs`, and then remove that label once we ensure they are documented.)

PR #407, opened 2016-09-06
More useful default lighting + remove alpha from lights

bcamper:
This PR has two parts:

### More useful default lighting
The previous default, if on `lights` are defined, has been to assume *100% omnidirectional diffuse light*. In other words, all geometry is lit with the vertex color (compounded by the material's `diffuse` component if present) at 100%.

The problem with this approach is that it is a poor default for extruded geometry, which end up showing as silhouettes with no depth.

This PR changes the default to a more useful value, a [directional](https://github.com/tangrams/tangram/compare/default-light?expand=1#diff-7a3c51617962d8b132a43dbe5d63e862R231) light that maintains 100% illumination for ground plane objects, with typical flat shading for extruded objects. The default is configured to provide *0.5 strength diffuse* and *0.5 strength ambient*. The default directional light angle in the XY plane is hardcoded for now, but could be made configurable if we want to provide a friendlier interface for controlling directional lights (since it turns out asking people to provide full XYZ vectors that happen to normalize to a length that provides the desired light intensity is unreasonable...).

Old:
tangram- 16 5072 40 7652 -73 9739_2016-09-06_17 21 54

New:
tangram- 16 5072 40 7652 -73 9739_2016-09-06_17 21 20

### Remove alpha from lights
In the course of changing the light default, I also discovered some very strange behavior with our lights and transparency in the scene. It turns out this is caused by *our lights having and "casting" an alpha component*. Examples of a point light casting full transparency outside its radius:

tangram- 16 7195 40 7643 -73 9716_2016-08-26_17 48 09 1

Cool-looking but ultimately wrong and unhelpful :)

To fix, this PR *changes all lighting components from `vec4` to `vec3`*, and changes all of the "light accumulation" logic (that sums the `light_accumulator_*` variables) to only operate on the RGB components (with alpha set to 1).

Additionally, to allow a material's alpha components to have some effect in while using the default "opaque" blend mode, the material's alpha is now multiplied into its RGB components, e.g. diffuse light is calculated as (note **material.diffuse.rgb * material.diffuse.a**):

`color.rgb += light_accumulator_diffuse.rgb * _color.rgb * material.diffuse.rgb * material.diffuse.a;`

Behavior is unchanged for other blending modes, where alpha is used in the blending equation. [See more here.](https://github.com/tangrams/tangram/compare/default-light?expand=1#diff-7a3c51617962d8b132a43dbe5d63e862R145
)

Here is a final side by side from Differ showing previous (left) and corrected behavior (right):
screen shot 2016-09-06 at 5 36 12 pm

### DIFF
--- a/src/light.js
+++ b/src/light.js
@@ -1,6 +1,7 @@
 import ShaderProgram from './gl/shader_program';
 import GLSL from './gl/glsl';
 import Geo from './geo';
+import Vector from './vector';
 import {StyleParser} from './styles/style_parser';
 
 let fs = require('fs');
@@ -222,13 +223,13 @@ class DirectionalLight extends Light {
         this.struct_name = 'DirectionalLight';
 
         if (config.direction) {
-            this.direction = config.direction;
+            this._direction = config.direction;
         }
         else {
             // Default directional light maintains full intensity on ground, with basic extrusion shading
             let theta = 135; // angle of light in xy plane (rotated around z axis)
             let scale = Math.sin(Math.PI*60/180); // scaling factor to keep total directional intensity to 0.5
-            this.direction = [
+            this._direction = [
                 Math.cos(Math.PI*theta/180) * scale,
                 Math.sin(Math.PI*theta/180) * scale,
                 -0.5
@@ -238,8 +239,15 @@ class DirectionalLight extends Light {
                 this.ambient = GLSL.expandVec3(0.5);
             }
         }
+        this.direction = this._direction.map(parseFloat);
+    }
 
-        this.direction = this.direction.map(parseFloat);
+    get direction () {
+        return this._direction;
+    }
+
+    set direction (v) {
+        this._direction = Vector.normalize(v);
     }
 
     // Inject struct and calculate function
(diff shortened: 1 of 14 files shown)


bcamper:
cc @blair1618 @patriciogonzalezvivo @meetar while there's some nuance here, I'm interested in getting this into the v0.10 release both because of the behavioral fixes, and because the new directional defaults help simplify our examples.

src/light.js
```diff
@@ -203,7 +221,25 @@ class DirectionalLight extends Light {
...
         this.struct_name = 'DirectionalLight';
 
-        this.direction = (config.direction || [0.2, 0.7, -0.5]).map(parseFloat); // [x, y, z]
+        if (config.direction) {
+            this.direction = config.direction;
+        }
+        else {
+            // Default directional light maintains full intensity on ground, with basic extrusion shading
+            let theta = 135; // angle of light in xy plane (rotated around z axis)
+            let scale = Math.sin(Math.PI*60/180); // scaling factor to keep total directional intensity to 0.5
```
blair1618: Is the `direction` vector normalized at some point? I'm guessing it is if this scaling is necessary, I just don't see it here.

bcamper: Yes it looks like it's normalized in the shader: https://github.com/tangrams/tangram/blob/master/src/gl/shaders/directionalLight.glsl#L20

But perhaps we should move that to CPU-side?

patriciogonzalezvivo: Better to doit once and for all... that for each pixel every time

blair1618: Ah I see, yeah tangram-es normalizes these CPU-side so I wanted to make sure the results wouldn't differ.

blair1618:
LGTM

bcamper:
Moved light direction normalization to CPU side.

PR #408, opened 2016-09-06
Add video capture methods

bcamper:
It turns out there is a surprisingly complete set of new [MediaRecorder APIs](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder) that allow you to [capture the contents of a Canvas element](https://hacks.mozilla.org/2016/04/record-almost-everything-in-the-browser-with-mediarecorder/) to a video stream. 

This PR adds new `Scene.startVideoCapture()` and `Scene.stopVideoCapture()` methods for capturing a WebM-encoded movie of a Tangram map.

`stopVideoCapture()` returns a promise that will resolve with `url` and `blob` properties, similar to the existing `Scene.screenshot()` API. In addition, a new `type` property is now returned with both screenshot and video promises, indicating the file type. Currently the `type` is always `png` for screenshots, and `webm` for videos.

This PR also consolidates screenshot and video behavior into a new `MediaCapture` module.

MediaRecorder functionality is currently supported in Chrome and Firefox. `startVideoCapture()` will return `false` on other browsers.

A simple video capture button is also added to the bundled demo.

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -383,6 +383,28 @@ Enjoy!
         };
         gui.add(gui, 'screenshot');
 
+        // Take a video capture and save to file
+        if (typeof window.MediaRecorder == 'function') {
+            gui.video = function () {
+                if (!gui.video_capture) {
+                    if (scene.startVideoCapture()) {
+                        gui.video_capture = true;
+                        gui.video_button.name('stop video');
+                    }
+                }
+                else {
+                    return scene.stopVideoCapture().then(function(video) {
+                        gui.video_capture = false;
+                        gui.video_button.name('capture video');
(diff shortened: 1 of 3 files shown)


bcamper:
cc @thisisaaronland

thisisaaronland:
There will be blog posts, right?

dmvaldman:
So cool!

bcamper:
I think you mean vlog posts, @thisisaaronland.

thisisaaronland:
Only if you can talk about it the way Jony Ives talks the manufacturing process.

bcamper:
Bead-blasted, lovingly crafted canvas streams.

thisisaaronland:
"We have polished each pixel in a bath of superior GL optimonium."

PR #425, opened 2016-10-13
Fix and speed up isRTL function

batje:
function was throwing errors and had an updated faster alternative from the original source on stackoverflow.

After updating from 0.8 to 0.10.4 I got the following errors on some maps, in some areas. I found that the original code that came from stackoverflow had an update from the code that was copied, and when I used that new updated code, the errors went away. However, I don't understand how to test if the original purpose of the code is still intact.

The exact error:
7c616e0a-0e16-440d-b1a5-64a3fd7865d7:1 Uncaught (in promise) Invalid regular expression: /^[-@[-`{-Â¿Ã—Ã·Ê¹-Ë¿â€€-â¯¿â€- â€¬â€¯-â¯¿]*[Ö‘-ß¿â€â€«â€®ï¬-ï·½ï¹°-ï»¼]/: Range out of order in character class: SyntaxError: Invalid regular expression: /^[-@[-`{-Â¿Ã—Ã·Ê¹-Ë¿â€€-â¯¿â€- â€¬â€¯-â¯¿]*[Ö‘-ß¿â€â€«â€®ï¬-ï·½ï¹°-ï»¼]/: Range out of order in character class
    at RegExp (native)
    at i (http://localhost/TheOtherMaps/dist/lib/tangram.min.js:17:25951)
    at a (http://localhost/TheOtherMaps/dist/lib/tangram.min.js:17:26064)
    at o (http://localhost/TheOtherMaps/dist/lib/tangram.min.js:17:26215)
    at http://localhost/TheOtherMaps/dist/lib/tangram.min.js:17:27514

PR #426, opened 2016-10-14
Compiled code no longer backwards

dmvaldman:
Resolves the issues brought up in https://github.com/tangrams/tangram/pull/425

Our unicode range strings to detect RTL text were being converted to their textual equivalents upon bundling. This caused some very bizarre errors, such as the appearance of unicode characters such as ﷽ in our source, in addition to much of our source code being _backwards_. 

The fix is to create the RegEx directly, instead of building it out of these strings that will get corrupted upon bundling. Thanks to @batje for spotting this!

### DIFF
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -415,13 +415,10 @@ CanvasText.cache_stats = { hits: 0, misses: 0 };
 CanvasText.texcoord_cache = {};
 
 // Right-to-left / bi-directional text handling
-// Taken from http://stackoverflow.com/questions/12006095/javascript-how-to-check-if-character-is-rtl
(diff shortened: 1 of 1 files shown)


meetar:
Niiiiiice. 🔍 👀 👌

nvkelso:
DUDE! :)

pnorman:
tnellecxe

dmvaldman:
rtl

bcamper:
@dmvaldman does this PR supersede #425?

dmvaldman:
It should be used instead of #425

bcamper:
Released in https://github.com/tangrams/tangram/releases/tag/v0.10.6

batje:
Thanks for that explanation @dmvaldman, I finally understand what I reported now!

dmvaldman:
I'm glad you submitted the issue! That was a wild rabbit hole to go down.

Issue #431, opened 2016-10-19
Italic text not accounted for in collision box


bcamper:
It looks like the slant from italic text is not accounted for in the collision box:

![screen shot 2016-10-19 at 10 15 26 am](https://cloud.githubusercontent.com/assets/16733/19522393/0be0af9c-95e5-11e6-89a8-b9248e196e62.png)

bcamper:
@dmvaldman FYI I usually close these when they are released to production, so we can do a final verification - I believe this fix is merged to master but not released until 0.11

dmvaldman:
Ah sorry. I thought the protocol was close on merge.

bcamper:
Part of the thinking there was to only notify external people that their issue is closed when it's released and they can actually get the benefit :)

bcamper:
Released in 0.11.0

PR #436, opened 2016-10-25
Fix collision for text styling that exceeds bounding box #431

dmvaldman:


bcamper:
@dmvaldman I was able to reduce where we need to pass through this italics info, because once it's set in the `layout`, it's already available where it's needed downstream https://github.com/tangrams/tangram/pull/436/commits/7aff9b8e82581c3d5dd7c858474690b8937fc9c8

PR #437, opened 2016-10-26
Revert move-into-tile functionality

dmvaldman:
Re-implement `move-into-tile` logic, but with `point` labels set to `move-into-tile = false`. This allows for text labels to be shifted into a tile to preserve features like neighborhoods, that aren't fixed to a particular geo location.

### DIFF
--- a/src/labels/label_point.js
+++ b/src/labels/label_point.js
@@ -1,5 +1,6 @@
 import Label from './label';
 import PointAnchor from './point_anchor';
+import Geo from '../geo';
 import OBB from '../utils/obb';
 import {StyleParser} from '../styles/style_parser';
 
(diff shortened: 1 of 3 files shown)


bcamper:
Updated Differ anchor unit tests and confirmed:

http://tangrams.github.io/differ/?1=https://tangrams.github.io/differ-tests/unit-tests/unittests01.json&2=&lib1=0.10&lib2=http://localhost:8000/dist/tangram.debug.js&ignore1&ignore2&go

meetar:
Is this `move_into_tile` entry still accurate? https://github.com/tangrams/tangram-docs/blob/92960852f67be03fc82cc4bcac7a044169255883/pages/draw.md#move_into_tile

dmvaldman:
Previous:

> [JS-only] Optional boolean. Default is true.
>
> Applies to text styles. Moves the label into the tile if the label would otherwise cross a tile  
> boundary. This should be set to false if using anchor/align functionality for text + icon 
> combinations – otherwise, text can get out of sync with expected position.

New:

> [JS-only] Optional boolean. Default is false for text-labels and true for point-labels.
>
> Applies to point and text styles. Moves the label into the tile if the label would otherwise cross a 
> tile boundary. This is automatically set to false if using anchor/align functionality for text + icon 
> combinations – otherwise, text can get out of sync with expected position.

How does this look @bcamper?

bcamper:
I believe the behavior we settled on is that this feature is only available for standalone `text` labels -- not for `points`, or for ` text` labels attached to a point.

dmvaldman:
Sorry, I reversed it. Should be "Default is true for text-labels and false for point-labels".

bcamper:
Yeah but the behavior doesn't actually exist for point labels, so "default
false" is not accurate... it can't be true :)

PR #440, opened 2016-11-01
Use simple box intersection function for collision

bcamper:
Here's a fun one: reviewing our dependencies to reduce library size, I tried replacing our [`box-intersect`](https://github.com/mikolalysenko/box-intersect) module with a simple homegrown box intersection function:

The simple version not only produces a **6k smaller** minified/gzipped bundle, it's also **25% faster** on a complex label scene (Bubble Wrap central London z17).

It's possible a fancier implementation will be more performant when we have more complex collisions (such as pooling all tile labels and colliding in screen-space), so we can revisit at that time. Until then, score another one for the Dumbest Thing Possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "box-intersect": "1.0.0",
     "brfs": "1.4.3",
     "csscolorparser": "1.0.3",
     "earcut": "2.1.1",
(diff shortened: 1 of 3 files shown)


meetar:
🏆💯

dmvaldman:
Perhaps we should advertise that this is the _new_ smallest and fastest box collision library for most practical use cases.

PR #441, opened 2016-11-02
Control point placement by line length ratio

bcamper:
#409 improved placement of points-on-lines -- for symbols like road shields and one-way arrows -- by adding placement strategies on line segment *midpoints* (`placement: midpoint`), or *evenly spaced* along the line (`placement: spaced`).

For road shield placement, we are using `placement: midpoint` to keep shields as far away from line vertices as possible. This is because shields placed at vertices are more likely to be ambiguous, such as where two highways intersect.

The code from #409 places no restrictions on the *length* of the line segment on which it places a point, so a shield (or other icon) may be placed on a line segment much shorter than itself (the icon may entirely cover the line segment). This appears problematic for shields because highway interchanges are more likely to simply into many short line segments, causing the same type of ambiguity that vertex placement did.

To improve this, this PR introduces a new parameter, `placement_min_length_ratio`, that specifies the **minimum line segment length as a ratio to the point being placed**. Here are some example values for this parameter:

- `placement_min_length_ratio: 1` (default value), will only place points on line segments that are *at least as long* as the point itself (the point must fit 100% along the line segment)
- `placement_min_length_ratio: 0` disables this behavior by allowing a point to place on a line segment of *any* length (minimum length of 0)
- `placement_min_length_ratio: 2` requires the line segment to be *at least twice* as long as the point
- `placement_min_length_ratio: 0.5` requires the line segment to be *only 50%* as long as the point

Here is a before/after example in Chicago. Note there are several shields placed very near to intersections that are pushed further away with this behavior enabled.

shield-min-ratio-chicago

This parameter is not a silver bullet, needs to be tuned (by zoom) or it can drop too many shields (not enough line segments that are long enough to fit), but in many cases it can provide significant improvement.

The parameter can be zoom-interpolated with stops, or accept a JS function. It is only used for `placement` strategies of `midpoint` and `spaced` (does not affect `vertex` placement).

This control was required because the desired behavior is somewhat dependent on the data source. Current Mapzen tiles still have enough short line segments at mid-zooms that the best ratio sometimes needs to be below `1` (or there is an insufficient # of shields placed). If we introduced further road network generalization in the future (through something like [skeletron's](https://github.com/migurski/Skeletron) approach), we would want to re-tune these.

The example screenshot above was using these values across zooms (should be tuned further by @sensescape and @nvkelso!):

```
placement_min_length_ratio:
  - [8, 0.1]
  - [9, 0.25]
  - [10, 0.5]
  - [11, 1]
  - [12, 1]
  - [13, 1.50]
  - [14, 2.0]
```

### DIFF
--- a/src/labels/point_placement.js
+++ b/src/labels/point_placement.js
@@ -4,15 +4,16 @@ import LabelPoint from './label_point';
 import {isCoordOutsideTile} from '../builders/common';
 
 const PLACEMENT = LabelPoint.PLACEMENT;
-const default_spacing = 50; // spacing of points along line in pixels
+const default_spacing = 80; // spacing of points along line in pixels
 
 export default function placePointsOnLine (line, size, options) {
     let labels = [];
     let strategy = options.placement;
+    let min_length = Math.max(size[0], size[1]) * options.placement_min_length_ratio * options.units_per_pixel;
 
     switch (strategy){
         case PLACEMENT.SPACED:
-            let result = getPositionsAndAngles(line, options);
+            let result = getPositionsAndAngles(line, min_length, options);
             // false will be returned if line have no length
             if (!result) {
                 return [];
@@ -56,9 +57,11 @@ export default function placePointsOnLine (line, size, options) {
                     0.5 * (p[1] + q[1])
                 ];
                 if (options.tile_edges === true || !isCoordOutsideTile(position)) {
-                    let label = new LabelPoint(position, size, options);
-                    label.angle = getAngle(p, q, options.angle);
-                    labels.push(label);
+                    if (!min_length || norm(p, q) > min_length) {
+                        let label = new LabelPoint(position, size, options);
+                        label.angle = getAngle(p, q, options.angle);
+                        labels.push(label);
+                    }
                 }
             }
             break;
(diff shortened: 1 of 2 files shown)


bcamper:
@nvkelso @blair1618 how does this look?

nvkelso:
Can you show a screenshot in London? I’m curious to see how it interacts
with repeat spacing to filter nearby shields from display.

bcamper:
Here's London (which is still pretty wonky in general with the plain road text labels).

There's no specific impact to repeat spacing, placement runs before repeat culling, so it reduces the set of candidates that may survive. You can also tune it with priorities.

output_q2ni5l

bcamper:
e.g. above, I think you could get more shields to place by removing or changing priority of those pre-existing text labels on roads, I think that is causing some weirdness here

blair1618:
To confirm some details: This is a new draw parameter for `points` styles that can be either a number, a JS function that returns a number, or stops, and the default value is 1. Is that right? 

What exactly are the two lengths that are compared? It seems like the metric we care about here is screen space size, but for linestrings that changes with tilt/rotation. For simplicity, we can consider only screen size at zero tilt for linestrings. For the point, it seems that we're taking the maximum dimension of the label. If a label has a "child" label, does that impact this measurement?

bcamper:
@blair1618 your assumptions are correct.

The min length takes the *max* of the point's width and height (this was simplest, could be changed), and converts into internal tile units: https://github.com/tangrams/tangram/pull/441/files#diff-7715e52b76e0d7236d7a669fa2d9a47dR12

You're right that the desired length is screen space here, adjusting for tilt would make sense (would be interesting to see the behavioral change here).

For child labels (`text` attached to `point`), there is no impact. The point (parent) alone is what is considered for placement (child text goes through a different path). This has no impact for road shields as the text is placed within the shield sprite anyway. Adding the text size to the minimum placement length sounds interesting but complex, especially considering dynamic anchor placement. So, I would continue to skip it :)

blair1618:
Cool, I can't think of any problems with this.

Issue #453, opened 2016-11-21
The Chinese label shows very vague. 


ligai322:
**EXPECTED RESULT:**

The text displayed in the green box is clear. 
But the text displayed in the red box is vague.
The two types of text displayed on the same layer, using the same style.
![image](https://cloud.githubusercontent.com/assets/3262486/20469570/dad098a2-afdb-11e6-9329-62cdd5e97525.png)

**ENVIRONMENT:**
OS : window 8.1
browser version : firefox 49.0.2   or google chrome 53.0.2785.116 m

**OTHER:**
 [this demo](http://tangrams.github.io/bubble-wrap/#4.375/40.463/78.215) has the same problem. 
![image](https://cloud.githubusercontent.com/assets/3262486/20469702/dd63fb4e-afdc-11e6-93da-f4cd1a1b0f60.png)

nvkelso:
Can you clarify if the text content is incorrect (something upstream in OpenStreetMap), or Tangram renderingor basemap style related?

Looking at the different colored boxes it's not obvious to me.

ligai322:
@nvkelso    [this demo](https://mapzen.com/tangram/play/?scene=https%3A%2F%2Fapi.github.com%2Fgists%2F2b273a3ca5531c706c6cf69cb09a007d#10.2917/30.2751/114.7863)
like this 
![image](https://cloud.githubusercontent.com/assets/3262486/20472882/ddde7a3e-aff7-11e6-8cac-738b7b9100e0.png)

ligai322:
There is another problem. When the label with stroke, The effect of the display is shadow, not what I want halo. so how do I need to achieve halo effect
![image](https://cloud.githubusercontent.com/assets/3262486/20473148/9c6d2dd2-aff9-11e6-972d-7dedf9b510e6.png)

nvkelso:
I still don’t know what to debug, can you help me?

> The text displayed in the green box is clear.
> But the text displayed in the red box is vague.

What is “clear” and “vague”? Is it the text content or how it is rendered?

meetar:
@nvkelso Some but not all of the labels are blurry – in this image, for
instance:

https://cloud.githubusercontent.com/assets/3262486/20469702/dd63fb4e-afdc-11e6-93da-f4cd1a1b0f60.png

The top boxed label is blurry, the bottom one is not.

This looks like something to do with texture definitions, @bcamper will
have more definite ideas.

bcamper:
[...]

PR #454, opened 2016-11-22
Round positions and sizes of labels to nearest pixel values

dmvaldman:
A potential solution to https://github.com/tangrams/tangram/issues/453

Run it on the differ [here](https://tangrams.github.io/differ/?1=https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap.yaml&2=https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap.yaml&lib1=https://mapzen.com/tangram/0.11/tangram.debug.js&lib2=https://precog.mapzen.com/tangrams/tangram/rounding/dist/tangram.debug.js&go)

unnamed-1

### DIFF
--- a/src/labels/label_line.js
+++ b/src/labels/label_line.js
@@ -1,6 +1,7 @@
 import Label from './label';
 import Vector from '../vector';
 import OBB from '../utils/obb';
+import Utils from '../utils/utils';
 
 const PLACEMENT = {
     MID_POINT: 0,
@@ -250,12 +251,15 @@ export default class LabelLine {
 
         switch (this.placement) {
             case PLACEMENT.CORNER:
-                position = segment[1].slice();
+                position = [
+                    Utils.roundToNearestPixel(segment[1][0]),
+                    Utils.roundToNearestPixel(segment[1][1]),
+                ];
                 break;
             case PLACEMENT.MID_POINT:
                 position = [
-                    0.5 * (segment[0][0] + segment[1][0]),
-                    0.5 * (segment[0][1] + segment[1][1])
+                    Utils.roundToNearestPixel(0.5 * (segment[0][0] + segment[1][0])),
+                    Utils.roundToNearestPixel(0.5 * (segment[0][1] + segment[1][1])),
                 ];
                 break;
         }
@@ -328,7 +332,7 @@ export default class LabelLine {
                     this.aabbs.push(aabb);
 
                     this.offsets[i] = [
-                        this.layout.offset[0] + nudge,
+                        this.layout.offset[0] + Utils.roundToNearestPixel(nudge),
                         this.layout.offset[1]
                     ];
 
@@ -356,7 +360,7 @@ export default class LabelLine {
                     this.aabbs.push(aabb);
 
                     this.offsets[i] = [
-                        this.layout.offset[0] + nudge,
+                        this.layout.offset[0] + Utils.roundToNearestPixel(nudge),
                         this.layout.offset[1]
                     ];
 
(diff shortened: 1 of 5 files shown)


pnorman:
Does this handle retina displays where the pixels you need to round to are at a higher resolution?

dmvaldman:
Yes. Things are rounded via [this](https://github.com/tangrams/tangram/blob/rounding/src/utils/utils.js#L181-L184) code, which takes into account screen resolution.

pnorman:
I'm seeing text with both that doesn't appear to be pixel aligned
image
image

bcamper:
@pnorman yeah, see the companion screenshot here: https://github.com/tangrams/tangram/issues/453#issuecomment-262388964

I think there are a couple fixes needed at different parts of the pipeline.

src/labels/label_point.js
```diff
@@ -50,8 +51,8 @@ export default class LabelPoint extends Label {
     }
 
     updateBBoxes () {
-        let width = (this.size[0] + this.layout.buffer[0] * 2) * this.layout.units_per_pixel * Label.epsilon;
-        let height = (this.size[1] + this.layout.buffer[1] * 2) * this.layout.units_per_pixel * Label.epsilon;
```
bcamper: I think this epsilon was originally added to prevent objects that were "perfectly" aligned (e.g. a `10px` sprite with a label positioned exactly `10px` below) from colliding one another. It may not be needed anymore (since then there have been explicit sprite/text label linking code added, and collision code changes), but we will need to look closely.

src/labels/label_line.js
```diff
@@ -250,12 +251,15 @@ export default class LabelLine {
 
         switch (this.placement) {
             case PLACEMENT.CORNER:
-                position = segment[1].slice();
+                position = [
+                    Utils.roundToNearestPixel(segment[1][0]),
```
dmvaldman: Yeah, you're right. I'll remove this rounding call.

dmvaldman: It wouldn't make sense to snap to a pixel based unit, as this would likely be converted to some non-integer position after converting to screen space. Your shader rounding method would catch any such rational point, so it's best not to round at all, I'd think.

src/labels/label_point.js
dmvaldman: Right. I realize now the epsilons can be left, because they do not effect the actual size of the texture, only the size of the collision area.

src/labels/label_line.js
dmvaldman: Or I guess since the tile units are 4096 x 4096, and screen space is 256 x 256, we could round to the nearest 1/16th? (floating point error in JS is another problem to consider here)

bcamper:
The view matrices will shift things off the pixel grid anyway, which is why
I think snapping in the shader itself is necessary.

bcamper:
Comparing master, the `snap-to-pixel` branch, and this one, so far the piece of this that I've found definitely effective is the snapping to the pixel grid within the Canvas rendering:
https://github.com/tangrams/tangram/pull/454/files#diff-56e91e9082620f32a1ac931d75dc9957

I've incorporated those changes in the other branch here:
https://github.com/tangrams/tangram/commit/06be2bfcc6df1f1c69d343829c13a479f7d72ccd

This makes sense, because the *position* of the sprites gets snapped in the shader, but this only works if the *texture itself* was drawn on a pixel grid. So I think the only parts we need to snap CPU-side are anywhere that drawing to the Canvas happens.

bcamper:
Closing - some of this integrated into recently released text pixel snapping.

PR #455, opened 2016-11-24
Alpha blending

donflopez:
Hi guys!

We have a problem with the alpha channel and how flat polygons are rendered without a background (transparent one) and a simple leaftlet layer behind.

Let's see an example:

* How it should be:

screen shot 2016-11-23 at 10 18 52

* How it is:

screen shot 2016-11-23 at 10 16 37

I've made some changes to tangram adding a new blend option called depthinlay (I know, the name is the worst, sorry), to add the alpha channel and depth_write option just like an opaque but adding the src_alpha and dst_alpha just like an overlay or inlay.

I don't know if this is the best way to do this, could you help me on this problem? (Also changing the name of the option if it is necessary a new option).

Thank you guys!

### DIFF
--- a/src/light.js
+++ b/src/light.js
@@ -106,50 +106,27 @@ export default class Light {
                 vec4 color = vec4(0.0);
 
                 // Keep material alpha channel when alpha blending is on
-                #if !defined(TANGRAM_BLEND_OPAQUE)
-                    #if defined(TANGRAM_BLEND_BUFINLAY)
-                        color.a = _color.a; // use vertex color alpha
+                #if !defined(TANGRAM_BLEND_OPAQUE) && !defined(TANGRAM_BLEND_DEPTHINLAY)
 
-                        #ifdef TANGRAM_MATERIAL_EMISSION
-                           color.rgb = material.emission.rgb * material.emission.a;
-                        #endif
+                    #ifdef TANGRAM_MATERIAL_EMISSION
+                        color = material.emission;
+                    #endif
 
-                        #ifdef TANGRAM_MATERIAL_AMBIENT
-                            color.rgb += light_accumulator_ambient.rgb * _color.rgb * material.ambient.rgb * material.ambient.a;
-                        #else
-                            #ifdef TANGRAM_MATERIAL_DIFFUSE
-                                color.rgb += light_accumulator_ambient.rgb * _color.rgb * material.diffuse.rgb * material.diffuse.a;
-                            #endif
-                        #endif
-    
-                        #ifdef TANGRAM_MATERIAL_DIFFUSE
-                            color.rgb += light_accumulator_diffuse.rgb * _color.rgb * material.diffuse.rgb * material.diffuse.a;
-                        #endif
-    
-                        #ifdef TANGRAM_MATERIAL_SPECULAR
-                            color.rgb += light_accumulator_specular.rgb * material.specular.rgb * material.specular.a;
-                        #endif
+                    #ifdef TANGRAM_MATERIAL_AMBIENT
+                        color += light_accumulator_ambient * _color * material.ambient;
                     #else
-                        #ifdef TANGRAM_MATERIAL_EMISSION
-                            color = material.emission;
-                        #endif
-    
-                        #ifdef TANGRAM_MATERIAL_AMBIENT
-                            color += light_accumulator_ambient * _color * material.ambient;
-                        #else
-                            #ifdef TANGRAM_MATERIAL_DIFFUSE
-                                color += light_accumulator_ambient * _color * material.diffuse;
-                            #endif
-                        #endif
-
                         #ifdef TANGRAM_MATERIAL_DIFFUSE
-                            color += light_accumulator_diffuse * _color * material.diffuse;
-                        #endif
-
-                        #ifdef TANGRAM_MATERIAL_SPECULAR
-                            color += light_accumulator_specular * material.specular;
+                            color += light_accumulator_ambient * _color * material.diffuse;
                         #endif
                     #endif
+                    #ifdef TANGRAM_MATERIAL_DIFFUSE
+                        color += light_accumulator_diffuse * _color * material.diffuse;
+                    #endif
+
+                    #ifdef TANGRAM_MATERIAL_SPECULAR
+                        color += light_accumulator_specular * material.specular;
+                    #endif
+
                 // Multiply material alpha channel into material RGB when alpha blending is off
                 #else
                     color.a = _color.a; // use vertex color alpha
--- a/src/light.js
+++ b/src/light.js
@@ -107,24 +107,48 @@ export default class Light {
 
                 // Keep material alpha channel when alpha blending is on
                 #if !defined(TANGRAM_BLEND_OPAQUE)
-                    #ifdef TANGRAM_MATERIAL_EMISSION
-                        color = material.emission;
-                    #endif
+                    #if defined(TANGRAM_BLEND_BUFINLAY)
+                        color.a = _color.a; // use vertex color alpha
 
-                    #ifdef TANGRAM_MATERIAL_AMBIENT
-                        color += light_accumulator_ambient * _color * material.ambient;
-                    #else
-                        #ifdef TANGRAM_MATERIAL_DIFFUSE
-                            color += light_accumulator_ambient * _color * material.diffuse;
+                        #ifdef TANGRAM_MATERIAL_EMISSION
+                           color.rgb = material.emission.rgb * material.emission.a;
                         #endif
-                    #endif
 
-                    #ifdef TANGRAM_MATERIAL_DIFFUSE
-                        color += light_accumulator_diffuse * _color * material.diffuse;
-                    #endif
+                        #ifdef TANGRAM_MATERIAL_AMBIENT
+                            color.rgb += light_accumulator_ambient.rgb * _color.rgb * material.ambient.rgb * material.ambient.a;
+                        #else
+                            #ifdef TANGRAM_MATERIAL_DIFFUSE
+                                color.rgb += light_accumulator_ambient.rgb * _color.rgb * material.diffuse.rgb * material.diffuse.a;
+                            #endif
+                        #endif
+    
+                        #ifdef TANGRAM_MATERIAL_DIFFUSE
+                            color.rgb += light_accumulator_diffuse.rgb * _color.rgb * material.diffuse.rgb * material.diffuse.a;
+                        #endif
+    
+                        #ifdef TANGRAM_MATERIAL_SPECULAR
+                            color.rgb += light_accumulator_specular.rgb * material.specular.rgb * material.specular.a;
+                        #endif
+                    #else
+                        #ifdef TANGRAM_MATERIAL_EMISSION
+                            color = material.emission;
+                        #endif
+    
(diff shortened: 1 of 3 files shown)


bcamper:
Hey @donflopez, thanks, could you provide more info on how you think the blending should occur, channel-wise? Is this only an issue that affects a raster/leaflet background behind?

Trying to better understand the problem - I've found it very finicky to get the WebGL blending options (and interaction with Canvas and HTML elements behind it) to work for all the different use cases!

Unfortunately I do not think the proposed solution here will work on its own. The reason we cannot usually have depth write on in conjunction with blending is that blending is highly dependent on draw order (which Tangram reorders based on tiles, grouping by style/shader program, and blend options). It may work by chance in your example, but if translucent geometry in front is drawn *before* any geometries behind it, the ones that are behind will never be drawn (because they will fail the depth test).

This post provides some background, and also explains why we have the `add` and `multiply` options for `blend`:
http://www.openglsuperbible.com/2013/08/20/is-order-independent-transparency-really-necessary/

There are specific issues for WebGL around how it is composited with the background, and those may be affecting your case (and we may be able to address those directly). This post provides some info:
http://webglfundamentals.org/webgl/lessons/webgl-and-alpha.html

Another option for your specific case may be one of the existing blend options, along with the [`blend_order`](https://mapzen.com/documentation/tangram/styles/#blend_order) parameter - unfortunately, the interaction of blending and ordering can be an Achilles' heel of WebGL/Tangram, it has definitely been one of the trickier aspects :)

donflopez:
Hi @bcamper!

Thank you! I'm starting with WebGL, so I tried to solve it for my self, sorry. I'll explain the problem that we have here.

We have to apply opacity and composition operations, like add, multiply... and we are using Tangram for the data layer and leaflet for the basemap.

Let's start with the opacity:

With mapnik, having two polygons with an opacity of 0.5, the result is the next one.

Mapnik version

With tangram this is the result.

Tangram version

With my... hack/modification/thing? What I have is the same color that mapnik, but the hidden part of the polygon doesn't appear making the color more dark like .75 of opacity (It is also happening on tangram without modifications).

Tangram hack

What if I use multiply or add?

Multiply:

screen shot 2016-11-30 at 15 59 13

Add:

screen shot 2016-11-30 at 15 59 39

Do I need to set a different lights?

Do you think that we have to use tangram to paint the base layer too to avoid this problem?

Could be something wrong with my configuration?

Thank you!

meetar:
Hi @donflopez – looks like there are a couple of things happening here, including a possible bug!

I tried to replicate your setup using the current Tangram build, so if there are bugs introduced in your branch my experiment won't address them.

First, assuming you have two polygons (the US polygon, and a second polygon) it looks as though they're being pre-composited somehow – in other words they are both being drawn as opaque, and then the result of that operation is given an alpha of its own. Shapes drawn with alpha should composite over each other properly with the `overlay` or `inlay` blend modes (`add` and `multiply` ignore alpha).

However, when investigating this, I found that Tangram may be processing alpha incorrectly even when used properly: an alpha value of `.5` results in a completely opaque polygon. So this may be causing the precompositing problem, and may also be related to the color differences you saw between the mapnik and Tangram results.

I have a test scene here which I think replicates your setup – in this case I'm bringing the raster layer in as a Tangram source, but it should behave the same when drawn as a separate Leaflet layer:

https://mapzen.com/tangram/play/?scene=https%3A%2F%2Fgist.githubusercontent.com%2Fanonymous%2F2b06f1bfa5a916612667a655130a133f%2Fraw%2F9391c30c5c81863ae81d0c45f13709d680f12976%2Fscene.yaml#4.971/36.393/-78.869

You can test the behavior by manipulating the `blend` mode, as well as the alpha in the `color` values (eg. the last number of `[0, 0, 0, .3]`). We're going to take a look at that suspicious `overlay` alpha behavior here as well.

donflopez:
Hi @meetar! Thank you!

In your example, there are two different layers one for the US polygon and one for the other, in my example those polygons are in the same layer.

The screenshots are made with your tangram branch, not mine (except the third one).

Ok! I can see that over .5 the alpha is not working. I've made this other tries.

Two layers, one with the two polygons (US and the other at the right) and the other one with a tiny polygon over the others.

First, layer (US + Big Polygon) with 0.1 opacity, the other (tiny polygon) with 0.05 opacity.

This is the result:

screen shot 2016-12-02 at 11 29 04

We can see another polygon that doesn't exist in the dataset (maybe matching with some tiles).

The expected result is too different with those levels of alpha.

screen shot 2016-12-02 at 11 32 06

meetar:
Thanks @donflopez – we're taking a close look at the alpha code now.

donflopez:
Thanks to you! If you need me to try something else, just ping me!

bcamper:
Should be fixed by 47fc96cfebd62d4283436ee484e51dda09a4dde4, which will be in the v0.11.1 release.

donflopez:
Hi @bcamper, thank you!

I've been using the fix and I can appreciate that now, the colour is correct but the alpha is still being weird.

I'm a noob on this, but I think that you've missed to multiply the `_color.a` to the material.xx.a. It seems more opaque as more polygons are piled up. I've added it to light.js and another problem appear, the tiles of the middle looks brighter than the outer ones, could it be related with the material?

Probably, what I'm saying makes no sense, sorry! I'm trying to help you and minimize the effort by your side.

Thank you again!!

bcamper:
Hi @donflopez - there may be some misunderstanding as to expected behavior.

With standard blending (e.g. `blend: overlay`), it is *expected* that total opacity will *increase* where multiple partially transparent surfaces are stacked. In the shader, the `_color` is the initial vertex color assigned in the `draw` block, and `color` is the final computed output color. Since `color.a` is initialized to `_color.a`, additional multiplications by material properties would be undesirable.

Note that the `add` and `multiply` blend modes offer alternatives. 

Can you please provide more information and examples as to the use case you are trying?

Likewise, it is hard to know what is happening with the "darker tiles" screenshot without a more complete example. I have not seen behavior like that before, but I also can't tell how the scene is constructed or what code it is running (0.11.0, master, modified master, etc.).

donflopez:
Yes @bcamper, I didn't explain it correctly, sorry.

With your actual master code my example looks like this: [Video Link](http://recordit.co/lkCPwxAyK5)

master

Multiplying by _color.a (your actual master + my changes): [Video Link](http://recordit.co/PMH5loY1Vt)

master_alpha

As you can appreciate the first GIF show more opaques layers, making them practically full opaque, but you also can notice the difference between the middle and the outers tiles this is more visible in the second GIF.

Thank you!

bcamper:
Hi @donflopez - thanks, but I don't think we can offer any further feedback unless you can provide the scene files for these examples - perhaps there's an unexpected interaction between materials or lighting, but it's hard for us to know because it doesn't look similar to our own test results. Can you please provide?

donflopez:
Hi @bcamper, don't worry, I've uploaded my demo so you can just play with it.

https://cartodb.github.io/tangram-carto/

Press GO button and play! It is using your tangram.min.js version from -> https://mapzen.com/tangram/tangram.min.js, in my examples I've used your master tangram branch.

The example is in gh-pages branch... so you can download it an make whatever you want.

Thank you all!

bcamper:
@donflopez if you can just export your final scene object, that would probably work for us. For example, if `scene` is the Tangram scene object:

`copy(JSON.stringify(scene.config))`

Then paste into a file.

donflopez:
Hi @meetar & @bcamper!! Of course, thanks!!

Here is the [config](https://gist.github.com/donflopez/67103c08825fec0366dc4e7dda1feb80)

Let me know if there is something more that I can do.

Thank you guys!

donflopez:
Hi guys, I'm so sorry, we found the error.

We have a buffer for the tiles, so we are repainting the tiles and the alpha was not working as expected! It was our fault.

Sorry for waste your time.

Now another issue is revealed, how can we work with buffer? Do you support it?

Thank you!

bcamper:
The buffer will always be drawn. This should be fine for some cases like lines, and plenty of polygon cases (Mapzen tiles currently have the buildings layer unclipped by tile boundaries, so building polygons extend past the tile and some portions are duplicated between tiles). Could you try clipping your polygons to the tile boundary (or at least the ones you want to draw with alpha)?

javisantana:
hey @bcamper, I think the question is wider, as you know some vector tiles providers use vector tiles with buffers (mainly because mapbox libraries work in that way). In our case we can generate tiles without buffer but that's not going to be the general rule so the question is will tangram support vector tiles with buffers?

bcamper:
@javisantana, for most rendering cases (all the ones we've had for our own tiles + base maps, anyway), there haven't been any issues with buffering and as I said we do buffer some elements. I've also used Tangram with Mapbox's buffered tiles without problems. 

For this alpha case specifically, the border rendering/overlap is an issue, yes (that we jut haven't run into ourselves yet). I don't currently know of any easy solution to that, since the rendering mechanism of Tangram is somewhat different -- it paints the entire tile in 3d space unclipped, vs. masking the tile against the border (with a stencil buffer... also harder to do properly with arbitrary 3d view such as in Tangram ES). We can certainly think about what such a technique would mean for Tangram, but I suspect that in addition to being a large change, it could conflict with other assumptions in the rendering pipeline.

Can you better explain what you see as the need for buffering on those polygons? I know that Mapbox GL (and possibly Mapnik, can't recall) needs buffering on lines to produce proper joins (BTW this is a case where Tangram actually **doesn't** need buffered lines **because** it renders without a stencil -- the lines from adjacent tiles line up pixel for pixel, when you use a stencil, you're actually chopping *off* part of the line you want at the tile border). Is that the case for polygons like these? (Side note, there's definitely some inefficiency in having such a large buffer as you have now, even if you're stenciling it out, there's a *lot* of overdraw where the same screen pixels are being partially processed multiple times.)

Another transparency technique I've been meaning to experiment with more is stippling or "screen door". It's certainly less flexible, but is still in use in a fair number of video games due to the complexity of mixing arbitrary 3d geometry with alpha (which is closer to Tangram's architecture as well). This article provides a nice example: http://n00body.squarespace.com/journal/2009/9/14/stippled-alpha.html

javisantana:
There is no need for buffering but the thing is most of the cached tiles out there will have buffers (not as larger as the ones we have now because of a bug)

As I said we don't need them in our case (but for labeling) and I don't see this as something really important for rendering basemaps but datavis is a kind of different

pnorman:
> and possibly Mapnik, can't recall

Yes - each tile (or metatile) is rendered independently of others, so it has to have any objects from other tiles which would influence the look of the edges.

The same will be true for Mapbox GL producing server-side raster tiles, and probably the same for server-side raster tiles from Tangram ES.

bcamper:
Yep, alpha (more important for dataviz) is a different case, we'll give it some thought.

For labels, it's true that we have sub-optimal label placement with regard to tile edges in Tangram JS currently. In Tangram ES, label placement is done in screen-space, so labels across tiles are aware of one another; we expect to move to this method for JS as well.

For rasters we have not entirely decided on a method, but yes, will likely include some form of meta-tiling (and many Tangram features like perspective building extrusion are not really compatible with static rasters anyway... though there are some fun use cases for static maps and we have that planned).

bcamper:
Going to close this as a PR, and open an issue specifically for alpha + buffering.

bcamper:
See #469.

PR #458, opened 2016-11-29
Improve text clarity by aligning sprites to pixel grid

bcamper:
It was recently pointed out, in #453, that some of our label text is undesirably blurry, and further investigation showed that this is because labels are often not aligned to the screen's pixel grid.

This PR addresses this in two ways:

- When text is drawn to a Canvas, it should be aligned to the Canvas's pixel grid. Line heights and string widths from the context `measureText()` function (which can return fractional values) now have a ceiling applied.
- When text labels are positioned on screen (model, view, etc. matrices are applied), the text will often not be screen-aligned. These are now "snapped" to the pixel grid, inside the vertex shader. Snapping continuously can cause a distracting jitter effect, especially during panning. To ameliorate this, the snapping is:
  - Not applied when the map is actively panning.
  - Applied with a transition animation when the map stops panning. (Google Maps appears to do something similar.) If desired, the snap transition can be disabled with the `Tangram.debug.debugSettings.suppress_label_snap_animation` flag.

This results in significantly better text clarity, especially for non-retina displays, small font sizes, and non-Latin scripts:

snap-to-pixel

eyaaaaasuvork5cyii

nbgqrgsp92waaaaasuvork5cyii

1fwc9c81wk838a49sbutheockaaaaaelftksuqmcc

### DIFF
--- a/src/leaflet_layer.js
+++ b/src/leaflet_layer.js
@@ -117,14 +117,15 @@ function extendLeaflet(options) {
                 map.on('zoomstart', this.hooks.zoomstart);
 
                 this.hooks.dragstart = () => {
-                    this.scene.view.panning = true;
+                    this.scene.view.setPanning(true);
                 };
                 map.on('dragstart', this.hooks.dragstart);
 
-                this.hooks.dragend = () => {
-                    this.scene.view.panning = false;
+                this.hooks.moveend = () => {
+                    this.scene.view.setPanning(false);
+                    this.scene.requestRedraw();
                 };
-                map.on('dragend', this.hooks.dragend);
+                map.on('moveend', this.hooks.moveend);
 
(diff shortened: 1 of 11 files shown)


nvkelso:
@bcamper Does this also sharpen the sprite artwork, too? It seems to looking at the 3 preview images, but hard to tell on my display.

dmvaldman:
It should!

bcamper:
@nvkelso yes, should have mentioned that too. It generally has a greater effect on text, but can help sprites as well; for sprites that are scaled from their original size, it doesn't necessarily help or hurt to align them to the pixel grid. Text is always drawn at 100% size.

nvkelso:
Hurray! Hopefully I’ll be able to look at Tangram on my 1x external display
now :)

PR #467, opened 2016-12-12
Improve text clarity for angled labels by supersampling text rendering

bcamper:
This is a follow-up to #458. The pixel-snapping technique used there can only be applied to fully upright labels, such as for POIs. A different technique can be used to improve the clarity of angled labels, such as for roads:

By rendering the underlying label texture at a larger initial size, we can get better results when it is sampled during rotation. @meetar and I tried a few values and found that rendering at 150% of display size seems to give the best results (200% introduces new aliasing / over-sharpness by actually providing too many pixels to sample from).  

This significantly improves clarity for non-Latin scripts and small font sizes (though helps for all diagonal labels in any script).

Animated before-and-after:
supersample-diag-text

Side by side diff:
4nilggkwdlyaaaaasuvork5cyii

@dmvaldman this may require careful merging with your upcoming changes since there is almost certainly code overlap, but the actual footprint here is pretty small.

### DIFF
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -20,10 +20,10 @@ export default class CanvasText {
     }
 
     // Set font style params for canvas drawing
-    setFont ({ font_css, fill, stroke, stroke_width, px_size }) {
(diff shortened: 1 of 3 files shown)


nvkelso:
Nice work!

Issue #464, opened 2016-12-07
Border point


donflopez:
Hi guys,

I have not seen a way to set a border for a point (with colour and width), Is there anyway to set it up? If not, is it planned?

Thank you!

meetar:
Hi @donflopez – we don't plan to add support for a `border` parameter on a point style – however you may be able to make a custom style with a shader which places a border around your point, and in some cases you may be able to simply draw one larger point behind another (although you'll have to turn off collisions for that).

donflopez:
Hi @meetar! Yep, I've seen those possibilities but, could the border be dynamic based on a feature of the data?

Thank you!

meetar:
We don't have any built-in functionality which matches that description, but if you use the double-point trick then you could set the color of the background point with a function, and the effect would be to color the border based on feature properties.

donflopez:
If we make a PR with this feature, would you consider to add it to tangram?

meetar:
We've actually been considering changing the points style behavior in
Tangram-JS, to match the Tangram-ES behavior in ways which would allow
borders to be easier to implement! Currently the default circular point
shape is drawn with a custom shader, and it may be better to switch to
using pre-drawn textures instead. So of course feel free to experiment, but
this specific behavior is currently subject to change, and without knowing
more about how you plan to modify it I can't be sure that our goals will
line up.

meetar:
Changing this to a feature request, for tracking: would like a way to add borders to points.

javisantana:
the problem with pre-draw textures if that it's hard to deal with situations where you need dynamic border width (based on a attribute for example). That's a pretty typical thing in data visualization (our use case)

bcamper:
Sorry, to clarify, we mean we may switch to textures that Tangram draws on the fly, according to the styling rules in the scene file. Terminology in this case, "pre-drawn" = "CPU-side / before the shader", vs. the current method which is drawn entirely within the shader :)

bcamper:
Though even then your by-attribute example is a good one where this might break down. That suggests a method of having one (or a limited number) of SDF-style point textures that are dynamically re-styled in the shader may be the best option. cc @blair1618

donflopez:
Hey!

We've been thinking about the way to draw points, here are our insights:

- Draw with pre-drawn textures.
![presentacion sin titulo](https://cloud.githubusercontent.com/assets/463462/21218820/78d59be0-c2b2-11e6-80b3-ce0ec56622c6.png)
[...]

PR #472, opened 2016-12-19
Add point outline

donflopez:
As in raster tiles, our point has the capability to have border (`marker-line-*`), this PR add this feature starting from a basic option.

At the moment, tangram options for a point are:

- size: 10px || [10, 20] 
- color: rgba/rgb/hex/name color

With this PR we've added this two new options:

- outline_width: 2px
- outline_color: rgba/rgb/hex/name color

After adding this two new options we have points with strokes.

I've followed the actual trigonometry of the point used in mapnik.

presentacion sin titulo

NOTE: I've tried to add/change all the notes from [here](https://github.com/tangrams/tangram/issues/464#issuecomment-267353576).


bcamper:
Hey, thanks for the submission! Before reviewing, could you please rebase without the unrelated alpha changes?

donflopez:
Of course! :)

donflopez:
I think it's done @bcamper, sorry about the noise!

bcamper:
Thanks for this addition -- I think it is mostly straightforward, but will need to review after the holidays. If all goes well we would schedule it for the next minor release, v0.12.0, likely in January, since it affects the public API.

donflopez:
You're welcome! Thank you!, Happy Christmas!

bcamper:
Hi @donflopez, looking at this now. I may need to rebase against some significant label changes that will be merged soon.

donflopez:
Great!! Let me know if you need anything.

Thanks @bcamper!

bcamper:
Sorry for the delay, still working on getting some other large changes merged so we can put this on top :)

bcamper:
There is no important reason for the 15% fade range (and in fact that might be the cause of #492, though I haven't confirmed that). The point drawing code is realllly old (because we are most often using sprites, but there are good cases for points), so it is a good time to be reviewing and improving it.

For using something like a fixed pixel size fade: note that everything in the fragment shader is in UV space, which ranges from 0-1 across the entire point; you don't have have direct access to the pixel size of the point in the fragment shader, but we can pass this information from the vertex shader if needed. (This is probably why we used something proportional originally, but I agree it's not the best.)

@patriciogonzalezvivo could you help look at our point drawing shader code and work with @donflopez to incorporate the outline so that we get the best transition between fill and outline color (if present), and then fade alpha at the point edges?

@donflopez separately, I made some minor adjustments in this branch (if you want to incorporate them, or I can replay them on top of your latest): https://github.com/tangrams/tangram/compare/point-outline

patriciogonzalezvivo:
Hi @donflopez and @bcamper! happy to help : )

Question, Do we spect to `#ifdef TANGRAM_POINT_OUTLINE` to be `true` always here https://github.com/tangrams/tangram/compare/point-outline#diff-60e93cbad892a92b22ab3f8baaf17594R65 ?? What happen if is `false`? according to https://github.com/tangrams/tangram/compare/point-outline draws no circular point... well just the billboard.

So far, from what I see in https://github.com/tangrams/tangram/compare/point-outline we have [this](http://editor.thebookofshaders.com/?log=170228211133).

Based on that I did a small prototype [here](http://editor.thebookofshaders.com/?log=170228213451) with no conditionals, and adding a ring of the half of `v_outline_edge` from the outer border in a transition range of 3 pixels. That's what we want right?

To implement this, we just should pass a `varying` from the vertex shader as @bcamper suggest with the width or height of the point (I'm assuming is a square). `a_shape` seams to have extrusion values for the coroners, right? So we can pass the `length(a_shape.xy)` from the vertex shader as a `varying float v_size` (or something like that) to replace the `(1./u_resolution.x)`. on line 35 of [here](http://editor.thebookofshaders.com/?log=170228213451)

Happy to make a PR if everybody thinks this is ok.

donflopez:
Hi! Thank you :)

> Question, Do we spect to #ifdef TANGRAM_POINT_OUTLINE to be true always here 

You're right!

This is my [latest shader code]http://thebookofshaders.com/edit.php?log=170301083112) that I've no pushed waiting for your comments (I'm a noob with shaders, sorry!)

I said 3px as an example but if it is a good size for the fade, great!

Maybe you were confused with my wrong implementation, but IMO, we need the 3px fade at the outer border of the point, and also a good transition between the fill and the outline as you did, my transition between colours was completely wrong.

What do you think @patriciogonzalezvivo & @bcamper?

bcamper:
@patriciogonzalezvivo thanks, we need a fade first from the inner fill color to the outline color (if there is an outline), and then from that to transparent. See @donflopez diagram at the top for where those transitions happen (plus some buffer / fade area as we've discussed here).

patriciogonzalezvivo:
@donflopez Seams you are calculating the alpha of the edge twice... why not just once at the end.

@bcamper with buffer/fade on the alpha you mean a small space to properly fade it with out cuts on the edges of the billboard right?

patriciogonzalezvivo:
@donflopez [here](http://editor.thebookofshaders.com/?log=170301152819) I made a new one... this one in theory fades on the edges... but the alpha blending on the editor is not great.

I didn't follow the logic of the `TANGRAM_FADE_START` completely. Also I notice that I was wrong believing that I should draw the edge on half size... should be fully sized, because the billboard will grow the extra half of the width... that's right? no?

In that case the alpha blending is constant and always apply at the end from `1.-aa_factor` to `1.`. Right?

In the line of optimizations, we can send the `varying float v_size;` from the vertex shader (`v_size = length(a_shape.xy)`) or we can send the aa_factor precomputed `v_aa_factor = (1./length(a_shape.xy))*TANGRAM_PX_FADE_RANGE` (is always cheeper to make calculations on the vertex shader because is down less times). What do you think?

Note also In this sketch that beside not using `TANGRAM_FADE_START` I'm using a different define call `TANGRAM_PX_FADE_RANGE` which is directly in pixels. Hope that's ok with you.

patriciogonzalezvivo:
I'm checking the resulting image and the results are pretty ok

slack-imgs

donflopez:
That's really cool @patriciogonzalezvivo! It's just how I've imagined.

It's perfect IMHO.

Thank you!

patriciogonzalezvivo:
@donflopez Yay! Team work! this was FUN! we should collaborate more often! : )

donflopez:
I'm very happy to collaborate, I'm learning a lot from you guys :)

bcamper:
@patriciogonzalezvivo can you incorporate shader updates to the `point-outline` branch? I made a couple fixes there and I don't see any additional functionality here since.

patriciogonzalezvivo:
sure I'll do!

patriciogonzalezvivo:
Hi! @donflopez there is something funky about the geometry when outlines are passing... makes all the points different size !!??? 

Any idea why is that? it should just grow the billboard two half of the outline_width... right?
screen shot 2017-03-01 at 2 43 31 pm

patriciogonzalezvivo:
Also seams `v_outline_edge` is never 0. : /

patriciogonzalezvivo:
I add the shader prototype here: https://github.com/tangrams/tangram/commit/22daa2e8e86f6de95668d6cf11b166777ff85e01

I believe the sizing issue is coming from JS when the geometry is generated... @donflopez could you give my a hand w it?

patriciogonzalezvivo:
Also seams growing twice the size that it should... I believe  
screen shot 2017-03-01 at 2 57 45 pm

patriciogonzalezvivo:
I'm using this style 
[scene.yaml.zip](https://github.com/tangrams/tangram/files/811888/scene.yaml.zip)

nvkelso:
It the stroke meant to have blue and then another outside stroke of black
at the very edge?

bcamper:
@nvkelso no but wait till we have an updated version then we can review :)

donflopez:
Hi @patriciogonzalezvivo! The points with different sizes are happening because you're using a polygon as a source, not a point. Change `building` by `places` and it should work as expected.

Another thing is that I'm precalculating the edge of the outline in [point builder](https://github.com/tangrams/tangram/pull/472/files#diff-c2f8624a39c2cee4f40230440cd8b682R69), the unique thing to do is remove the `*.5`.

```
color = mix(color, v_outline_color, (smoothstep(v_outline_edge-v_aa_factor, v_outline_edge, point_dist)) * step(.01,v_outline_edge) );
``` 

I thought that precalculate the point where to start to draw the outline was better for performance, but if it not the case, we can do it in the vertex shader. What do you think?

I've also made some improvements to in the point builder part, I'll upload them now.

On the other hand, I've noticed that if we use an outline size greater than the point size it start over drawing an outline, what should we do here? Set the width as the outline size?

Thank you!

bcamper:
Hi @donflopez -- @patriciogonzalezvivo and I were tuning this yesterday on this branch, which is also rebased against the latest tangram master: https://github.com/tangrams/tangram/compare/point-outline

Can you check that branch out? I'm going to close this PR and open a new one continuing from there. Things are mostly working and looking good (yay), but there are a couple outstanding issues to deal with.

bcamper:
Continued as #508.

src/builders/points.js
```diff
@@ -5,9 +5,18 @@ import { default_uvs } from './common';
...
-    { quad, quad_normalize, offset, offsets, pre_angles, angle, angles, shape_w, curve, texcoord_scale, texcoord_normalize, pre_angles_normalize, angles_normalize, offsets_normalize }) {
+    { texcoord_index, position_index, shape_index, offset_index, offsets_index, outline_edge_index, pre_angles_index, angles_index },
+    { quad, quad_normalize, offset, offsets, pre_angles, angle, angles, shape_w, outline_width, curve, texcoord_scale, texcoord_normalize, pre_angles_normalize, angles_normalize, offsets_normalize }) {
+
     quad_normalize = quad_normalize || 1;
+
+    if (outline_edge_index && outline_width) {
+        let half_outline_width = outline_width === 1.0 ? outline_width : outline_width / 2;
+
+        quad[0] += half_outline_width;
```
bcamper: @donflopez for reference, the issue with the wrong point sizes was caused by modifying `quad` here -- this variable is reused for other points, so it can't be modified.

donflopez: Ups, I didn't know it, but why does it not happen with points?

bcamper: In master, the `quad` property is never modified: https://github.com/tangrams/tangram/blob/master/src/builders/points.js#L11-L18

donflopez: I mean, with my modified branch, if I use points as the data source, they don't change his size, but if you use polygons, all change them size.

Could it be because, with points, it's calling that function one time and with polygons that function is called one time per coords in the polygon?

bcamper: Not sure, I wasn't seeing that behavior, but it's probably just chance based on order of execution (whether the variable was modified and then reused or not).

bcamper:

PR #488, opened 2017-01-26
Bundleable and Renameable Tangram

dmvaldman:
This PR allows Tangram to be imported/required in as a CommonJS module. In addition it allows the distribution files (`tangram.debug.js`, `tangram.min.js`) to be renamed by the end user. 

Previously this was not allowed because our Web Workers needed a pre-defined filename to load their source code from. Now this source code is generated at runtime by getting the source code from within our UMD build by calling the `Function.prototype.toString` method. We intercept Browserify's bundling step to do this with a rather fragile string replacement. Another way to do this is to wrap Tangram in a self-executing function and call `arguments.callee.toString()` to get the source code, but there is no way I've found to do this without leaking a global variable.

I've confirmed that this build works both for Tangram development (which uses `Tangram` on the `window` object), and on Mapzen.js and Tangram-play by importing Tangram (using a local CommonJS build).

We can now look to putting Tangram on the NPM registry, and having Mapzen.js and Tangram-play use this method of importing Tangram so they can be bundled with Tangram.

### DIFF
--- a/src/scene.js
+++ b/src/scene.js
@@ -19,6 +19,7 @@ import FeatureSelection from './selection';
 import RenderStateManager from './gl/render_state';
 import FontManager from './styles/text/font_manager';
 import MediaCapture from './utils/media_capture';
+import source from './source';
 
 // Load scene definition: pass an object directly, or a URL as string to load remotely
 export default class Scene {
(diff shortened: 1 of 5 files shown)


src/scene.js
```diff
@@ -244,7 +245,16 @@ export default class Scene {
...
-        let worker_url = this.worker_url || URLs.findCurrentURL('tangram.debug.js', 'tangram.min.js');
+        let worker_url;
+        if (source._worker_src){
+            // load from source object using browserify shim (preferred)
+            worker_url = URLs.createObjectURL(new Blob(['(' + source._worker_src + ')()'], { type: 'application/javascript' }));
+            delete source._worker_src;
+        }
+        else {
+            // hardcode distribution build names (not preferred)
+            worker_url = this.worker_url || URLs.findCurrentURL('tangram.debug.js', 'tangram.min.js');
```
bcamper: Should we just eliminate this path?

dmvaldman: 👍

bcamper:
This looks good except that source maps are currently not included in the web worker (which has been critical for our development/debugging process). I believe this is because the code to stringify the source is injected before the source map has been generated. Perhaps it is possible to capture the source map at the *end* of the browserify process, then append it to the variable that has the stringified source.

bcamper:

PR #499, opened 2017-02-21
Bundling

dmvaldman:
This PR improves a prior one allowing Tangram to be bundleable (imported as an es6 module) and also allows for the file to be renamed. Unlike the previous PR, this also supports source maps in web workers. 

If the debug file (`tangram.debug.js`) is renamed, the source map file also need to be renamed to match (by adding `.map` to whatever the debug build's filename is). For example if `tangram.debug.js` is renamed to `tangram.js`, then the source map needs to be renamed to `tangram.js.map`. For the minified build, this is not necessary. We should encourage use of the minified build for any purpose where debugging Tangram is not necessary.

Currently this PR fails a certain test:

```
  Scene
    .load()
      loading scene from an existing object
        ✖ "before each" hook
          Firefox 34.0.0 (Windows 8)
        timeout of 2000ms exceeded
```

For reasons that are not clear to me as of yet. I cannot reproduce this failing test on my machine, but will continue to investigate.


build/quine.js
```diff
@@ -0,0 +1,50 @@
...
+
+var source_variable = '__worker_src__'; // variable name used to refer to the Tangram source code
+var source_origin = '__worker_src_origin__'; // variable name used to refer to URL origin of source code (e.g., "http://localhost:8000/dist/tangram.debug.js")
+
+// prepend this code - wraps Tangram source in a function and later calls arguments.callee.toString to get source
+var prefix = '(function(){';
+    prefix += 'var target = (typeof module !== "undefined" && module.exports) || (typeof window !== "undefined");';
+    prefix += 'if (target) {';
+    prefix += 'var ' + source_variable + ' = arguments.callee.toString();';
+    prefix += 'var ' + source_origin + ' = document.currentScript.src;';
```
bcamper: `document.currentScript` is not available in IE (http://caniuse.com/#feat=document-currentscript), so we will need to guard against it being null here.

bcamper:
@dmvaldman so I'm not entirely clear why this fixes the issue, but loading the worker from a fixed URL, instead of from a blob URL (created from the same content) seems to consistently pass remote tests: https://github.com/tangrams/tangram/pull/499/commits/37bd895ab95cc35a42c56800a025fd6717ec9236. I've sometimes seen issues with heavy use of blob URLs when testing (a lot of resources being allocated and deallocated), but there's not reason it *shouldn't* work. Anyway, I'm good to use this fix and move on :)

It still makes the most sense to let the worker load via the quine variable for tests for consistency with the real-world environment, but I know we were having a different problem with that.

dmvaldman:
Oh nice! I'm fine with this as well as a circle-ci solution

package.json
```diff
@@ -65,6 +68,7 @@
     "brfs": "1.4.3",
     "csscolorparser": "1.0.3",
     "earcut": "2.1.1",
+    "exorcist": "^0.4.0",
```
bcamper: This line is duped below. I also prefer to peg the version numbers to exact ones (instead of using `~` or `^`). It's more of a pain to upgrade dependencies, but we've been bitten multiple times by auto-upgrades of packages that were supposed to be backwards compatible but included regressions -- so we switched to exact versions.

package.json
```diff
@@ -97,6 +102,7 @@
     "karma-sinon": "1.0.4",
     "mocha": "1.21.4",
     "sinon": "1.10.3",
+    "through2": "^2.0.3",
```
bcamper: Let's use exact version, see comment above.

bcamper:
To recap a conversation from last week: source maps appear to not be working for me with the budo live reload path (`npm start`). It looks like something about the separate `quine_browserify.js` plugin is not being triggered, and the source map in the served file includes the original inline source map, rather than the expected extracted one (to separate `tangram.debug.js.map` file).

package.json
dmvaldman: good catch. these things were just auto-generated from the install command, and i forgot to check on it

dmvaldman:
By using the npm module `mapstraction` instead of `exorcist` we're able to run `npm start` with no problems. This resolves the last remaining hurdle to this PR!

bcamper:
There appears to still be an occasional pernicious source map issue, where breakpoints can't be set (in Chrome) on some lines. For example, trying to set a breakpoint on [this line](https://github.com/tangrams/tangram/blob/1b065279a9fe3fd99c1aeb07495a206177f9cc52/src/scene.js#L447) causes the breakpoint to skip down to line 450. But this only happens occasionally, so it's not a consistent off-by-one-line issue in the source map. However, *something* in the modifications (maybe the fact that there are still column offsets at the beginning and end) is making it behave differently in Chrome than the source map generated on master. I tested Firefox and Safari and this issue *doesn't* appear to happen there... however, they both have less robust support for source maps in the first place (they don't do things like skip non-executable lines at all, and neither supports source maps in a worker, making them not very useful for us for practical development), so they may be missing the issue because they are less sophisticated rather than more.

Not sure we want to invest in this route (could be another rabbithole), but it may be possible to use a tool like [`combine-source-map`](https://github.com/thlorenz/combine-source-map) to properly incorporate the quine code into the source map (though now that I say that, I just realized the quine code is *not* in the worker thread code, so the source map would not be the same anymore, and that defeats the purpose of trying to share it between threads).

We might also be able to just live with this issue. It's annoying, but in many cases can be worked around by setting a breakpoint above the desired line (not a perfects substitute for more advanced things like conditional breakpoints).

bcamper:
Magic string (https://github.com/Rich-Harris/magic-string) does look promising for fixing the source map  (as does combine-source-map), though I think we will need separate source maps for the main thread and worker in that case, since the source between the two *is* slightly different. I think that's fine though, since source maps are a very specific development case.

bcamper:
Going to merge as is and live with the sometimes-off-by-one error (would still like to follow up on it, but we can deal with it in dev for now).

bcamper:
Well... this is merged now, but github UI is stuck in a weird state because the CI build got stuck when the S3 outage occurred a couple weeks ago. The commits are rebased and merged into master now though.

dmvaldman:
🗞

Issue #489, opened 2017-01-31
Allow style definition to include defaults


nvkelso:
In most the Mapzen basemap styles we create a `style` and then a special `layer` with a draw section that sets up required defaults for that style (and is usually linked up to data `source`). 

The downside is when someone wants to use the defaults in our house basemap styles, they must use exactly the data source / layer names to enjoy the sensible draw style defaults. Or they need to dip into the YAML source code and copy our defaults to a custom layer.

Many other mapping and graphics applications allow defaults to be stored with at the `style` level, I propose Tangram also enable this.

nvkelso:
**Example 1:**

To make use of Mapzen's icon library, a developer would need to restate many of the property settings from the `pois` layer. Critically this includes the sprite's `size`, `sprite` lookup function, and `sprite_default`. 

From Walkabout, but applies to all house styles:

```
styles:
    icons:
        base: points
        texture: pois
        blend_order: 1

layers:
    pois:
        data: { source: mapzen, layer: pois }
        #visible: global.label_visible_poi_landuse
        filter:
            all:
                - not: { kind: [building,address,farm,apron,residential,commercial,industrial] }
                - $geometry: point
                # White list a few kinds to "always" show per the tile zoom inclusion, else test feature's min_zoom
                #- function() { return feature.min_zoom <= $zoom }
                - function() { if( feature.kind == 'attraction' || feature.kind == 'landmark' || feature.kind == 'hunting' || feature.kind == 'dive_centre' || feature.kind == 'gas_canister' || feature.kind == 'hunting' || feature.kind == 'fishing_area' || feature.kind == 'playground' || feature.kind == 'toilets' ) { return true } else { return feature.min_zoom <= $zoom } }
        draw:
            icons:
                visible: global.icon_visible_poi_landuse
                size: [[13, 18px], [16, 18px], [18, 22px]]
                sprite: function() { return feature.kind; }
                sprite_default: generic
                interactive: global.interactive
                priority: 65 #function() { return (feature.min_zoom && Math.floor(feature.min_zoom * 1000)) || 65; }
                repeat_group: abc
                buffer: 7px
                text:
                    visible: global.text_visible_poi_landuse
                    text_source: global.ux_language_text_source
                    text_wrap: 18
                    max_lines: 3
                    font:
                        family: global.text_font_family
                        weight: 600
                        fill: [0.10,0.10,0.10]
                        size: [[13,10px],[14,10px],[18,11px]]
[...]

PR #504, opened 2017-02-28
Styles can specify default `draw` parameters

bcamper:
Allows a rendering style to include defaults for `draw` parameters (that can then be overridden or supplemented as normal when the style is used in a draw group under `layers`).  See #489 for background.

Example, setting a default `size` for icons:

```
styles:
  # setting up style with default draw size
  icons:
    base: points
    texture: icons
    draw:
      size: 16px # default size

layers:
  ...
  draw:
    # drawing style with default size, plus setting sprite
    icons:
      sprite: coffee
```

This new capability does raise an unusual syntax issue though, which is that in some cases, the style definition may now include *all* the required draw parameters, but a draw group referencing that style must still be added in `layers` to invoke the actual draw operation.

With current Tangram JS behavior (unsure about ES), a `null` draw group will fail validation, and nothing will be drawn. Therefore, adding an all-default draw group requires this syntax:

```
layers:
  ...
  draw:
    icons: {} # draw with defaults
```

This branch also **relaxes** the handling of `null` draw groups, so that they *are* registered as valid draw groups (assuming they satisfy all required parameters after draw group defaults and merging are applied). This makes the following alternative syntax also valid:

```
layers:
  ...
  draw:
    icons: # draw with defaults
```

While both the empty object and `null` syntax are a bit awkward, I thought it was slightly better to provide this more streamlined option, and simply treat `null` draw groups as empty objects.

A more complex example is included in the bundled demo, for colored transit lines (based on OSM tags in Mapzen tiles):

```
styles:
     transit-lines:
         base: lines
         blend: overlay
         blend_order: -2
         shaders:
             blocks:
                 filter: |
                     color.rgb *= 1.25; // pump up the colors
                     color.a = 0.5;     // translucent
         draw: # default draw parameters
             color: function() { return feature.colour || 'gray'; }
             width: 6px
             outline:
                 color: [.8, .8, .8]
                 width: 1px
             interactive: true
```

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -89,6 +89,13 @@ styles:
                 filter: |
                     color.rgb *= 1.25; // pump up the colors
                     color.a = 0.5;     // translucent
+        draw: # default draw parameters
+            color: function() { return feature.colour || 'gray'; }
+            width: 6px
+            outline:
+                color: [.8, .8, .8]
+                width: 1px
+            interactive: true
 
     rainbow:
         base: polygons
@@ -747,18 +754,6 @@ layers:
             lines:
                 style: transit-lines
                 order: 400
-                color: gray
-                width: 6px
-                outline:
-                    color: [.8, .8, .8]
-                    width: 1px
-                interactive: true
-
-        colored:
-            filter: { colour: true }
-            draw:
-                lines:
-                    color: function() { return feature.colour; }
 
     # schools:
     #     data: { source: schools }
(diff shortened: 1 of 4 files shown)


bcamper:
cc @matteblair @tallytalwar on handling of `null` draw groups as empty objects.

tallytalwar:
Just did a quick check on `null` draw group. Obviously at present nothing will be drawn as no rules are present post merging.

How I imagine this being implemented on ES, is before merging of draw rules, initial set of rules will be the default rules picked from the style, so post merging even if a drawgroup is empty (read `null` or `{}`) one will have some draw rules coming from the default style (if specified), and unless a required rule is missing, feature will be drawn using these rules.

Issue #471, opened 2016-12-15
Boundary labels, text and line orientation flips


sensescape:
Added boundary labels on branches for places refactor on all the house styles.

Text orientation looks like this:
<img width="491" alt="screen shot 2016-12-15 at 1 29 28 pm" src="https://cloud.githubusercontent.com/assets/466585/21242799/e2682826-c2ca-11e6-9f0f-5ca8bec90f12.png">
here, Utah is on both sides

Here's another one:
<img width="755" alt="screen shot 2016-12-15 at 1 33 08 pm" src="https://cloud.githubusercontent.com/assets/466585/21242844/1a2f8010-c2cb-11e6-9056-559212ac1511.png">

bcamper:
@sensescape can you send a link to the YAML you're using for this?

@nvkelso may be a data issue, since Tangram won't know which way the line is "supposed" to go - or it could just be that Tangram is flipping the text orientation to keep it upright... in which case some additional styling and/or data hints would be needed so that it knows what it can move where.

sensescape:
@bcamper @nvkelso 
another example:
<img width="408" alt="screen shot 2016-12-15 at 1 52 34 pm" src="https://cloud.githubusercontent.com/assets/466585/21243467/d95ce53e-c2cd-11e6-9b5c-2870b0fcb609.png">

https://github.com/tangrams/refill-style/blob/sensescape/places-refactor/refill-style.yaml#L4014-L4040

https://github.com/tangrams/refill-style/blob/sensescape/places-refactor/refill-style.yaml#L4068-L4095

meetar:
iirc at one point we discussed having a label option for "inside" vs
"outside" …

bcamper:
@sensescape a couple related notes on syntax issues in the snippets you linked (these should be covered in the Tangram docs, we can clarify if not):

`text_wrap: 100px`
This parameter is the **number of characters** to wrap on, not the number of pixels. 

`offset: 10px`
This parameter requires **two values**, for x and y: `offset: [0px, 10px]` (offsetting a line label in the x direction probably isn't super useful, but it's technically possible and provides consistency on the `offset` parameter for point labels).

sensescape:
@bcamper 
ok, thanks!
will remove px and rework offset

nvkelso:
@bcamper We're still seeing inverted text (where left labels right, right labels left). Holding off on releasing anything related to this in Mapzen house styles until there's a fix (not urgent).

[...]

PR #506, opened 2017-03-01
Oriented lines

dmvaldman:
This PR addresses https://github.com/tangrams/tangram/issues/471 where we were not previously using the orientation of a line to modify an offset of a label. This allows a line label to remain on a specific side (inside or outside) of a polygon as it is placed around the polygon's boundary.

Currently the offset will follow the geometry orientation only if both an offset is specified and the `text_source` is one of `name:left` or `name:right`. This `text_source` pattern is currently used in our tiles. However, there was a discussion in https://github.com/tangrams/tangram/issues/471 to use a more intentional scene file markup like

```
draw:
  text:
    text_source:
      left: 'name:left'
      right: ‘name:right'
```
This part is not yet implemented. Are we in agreement about this syntax @nvkelso @bcamper?

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -127,13 +127,6 @@ styles:
             emission:
                 texture: images/sunset.jpg
                 mapping: spheremap
-    uv:
-        base: lines
-        texcoords: true
-        shaders:
-            blocks:
-                color: |
-                    color.rgb = vec3(v_texcoord.x, v_texcoord.y, 0.);
 
 sources:
     mapzen:
@@ -184,15 +177,11 @@ layers:
 
     earth:
         data: { source: mapzen }
+        visible: false
         draw:
             polygons:
                 order: global.feature_order
                 color: '#f0ebeb'
-            lines:
-                order: 100000
-                color: black
-                width: 1px
-                tile_edges: true
 
         continents:
             filter: { kind: continent, $geometry: point }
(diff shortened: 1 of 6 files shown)


nvkelso:
@dmvaldman Can you add a screenshot, please? And example YAML code snippet so I can give it a whirl in Bubble Wrap?

> However, there was a discussion in #471 to use a more intentional scene file markup like

+1 for adding the new text_source `left`, `right` syntax suggested in #471:

```
draw:
  text:
    text_source:
      left: 'name:left'
      right: ‘name:right'
```

dmvaldman:
Check out an example here https://precog.mapzen.com/tangrams/tangram/oriented-lines-example/

nvkelso:
This is super rad! I've tested in the areas mentioned in #471 and they're all looking good.

@dmvaldman and I were talking in Slack about exposing an option to fail the pair of labels if one of them can't be placed (similar to what happens with requiring both sprite and name for points by default, but allowing that to fail optionally).

https://precog.mapzen.com/tangrams/tangram/oriented-lines-example/#40.65394574139022,-74.09729003906251,13

screen shot 2017-03-01 at 13 46 16

bcamper:
Sweet, looking great. I also like extending `text_source` to accept an object with `left` and `right` properties (it currently accepts string and function values, so this is a natural extension). Adding a parameter to optionally require both works for me too (assuming it can be done without introducing a lot of complexity).

dmvaldman:
I'll look at using the point/text linking to make sure both left and right labels appear together.

Otherwise, here's my fav. Who knew Kentucky was an island? :-)

tangram-1488407583324

nvkelso:
Kentucky Bend :)

dmvaldman:
Do we want to support oriented offsets for both markups:

```
draw:
  text:
    text_source:
      left: 'name:left'
    offset : [0, 10px]
```

and 

```
draw:
  text:
    text_source: 'name:left'
  offset : [0, 10px]
```

Or just the former?

nvkelso:
I prefer this one as it isn't tied to the Tilezen implementation of vector tiles, and generally seems a bit more sane anyhow.

```
draw:
  text:
    text_source:
      left: 'name:left'
    offset : [0, 10px]
```

PR #508, opened 2017-03-02
Add point outline (rebased)

bcamper:
Continuing from #472.

### DIFF
--- a/src/styles/points/points_fragment.glsl
+++ b/src/styles/points/points_fragment.glsl
@@ -20,6 +20,8 @@ varying vec4 v_color;
 varying vec2 v_texcoord;
 varying vec4 v_world_position;
 varying float v_alpha_factor;
+varying vec4 v_outline_color;
+varying float v_outline_edge;
 
 #define TANGRAM_NORMAL vec3(0., 0., 1.)
 
@@ -56,7 +58,14 @@ void main (void) {
             // Draw a point
             vec2 uv = v_texcoord * 2. - 1.; // fade alpha near circle edge
             float point_dist = length(uv);
-            color.a = clamp(color.a - (smoothstep(0., TANGRAM_FADE_RANGE, (point_dist - TANGRAM_FADE_START)) / TANGRAM_FADE_RANGE), 0., color.a);
+
+            #ifdef TANGRAM_POINT_OUTLINE
+                color.a = clamp(color.a - (smoothstep(0., TANGRAM_FADE_RANGE, (point_dist - TANGRAM_FADE_START)) / TANGRAM_FADE_RANGE), 0., color.a);
+            #else
+                color = mix(color, v_outline_color, smoothstep(v_outline_edge, v_outline_edge + ((1. - v_outline_edge) / 2.), point_dist));
+                color.a = color.a - step(1., point_dist);
+            #endif
+
         #endif
     #ifdef TANGRAM_MULTI_SAMPLER
     }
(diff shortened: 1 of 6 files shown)


donflopez:
It works perfect! Thank you guys, it is a really great feature for us.

One questions, do we have a limit at the point size?

donflopez:
Sure! Let me help with that @bcamper, can you show me an example of the existing syntax?

bcamper:
@donflopez sure, a line outline is specified like:

```
draw:
  lines:
    color: red
    width: 10px
    outline:
      color: white
      width: 2px
```

The style code will automatically make this available as an object, so you can access them like `outline.color` and `outline.width`:
https://github.com/tangrams/tangram/blob/master/src/styles/lines/lines.js#L243-L245

You can update these and other locations where the `outline_*` params are referenced:
https://github.com/tangrams/tangram/blob/point-outline/src/styles/points/points.js#L137-L138
https://github.com/tangrams/tangram/blob/point-outline/src/styles/points/points.js#L365-L366

donflopez:
Thank you @bcamper I'll do it!

donflopez:
Sorry @bcamper, I can't push to this branch, I've the changes for the outline, should I create a PR to this PR?

Now I'm investigating the collision box issue.

bcamper:
You can create a PR to this PR, sure.

bcamper:
@donflopez thanks, I've already started on the collision issue (can be a little tricky to trace, with the parallel geometry building and collision processing steps) - will commit that soon

bcamper:
@donflopez here check this commit: 0e0f4dedb975169b4420c618524a73ede2d84739

The outline calculation logic was moved out of the builder, upstream into the rendering style, which means the final `quad` box passed to the builder already has the outline incorporated.

The builder is actually intended to do work that varies *per-vertex*, so simple properties that are constant across the feature, like color, etc. -- and outline width -- can be populated one-time in the `makeVertexTemplate()` method.

You may need to update your syntax changes though since I moved the references to the style properties around :)

bcamper:
Should also enforce that outlines are wider that than the point size itself...

donflopez:
Great!! Thank you guys 🎉

bcamper:

src/styles/points/points_fragment.glsl
```diff
@@ -49,14 +49,20 @@ void main (void) {
...
             // Draw a point
             vec2 uv = v_texcoord * 2. - 1.; // fade alpha near circle edge
             float point_dist = length(uv);
-            color.a = clamp(color.a - (smoothstep(0., TANGRAM_FADE_RANGE, (point_dist - TANGRAM_FADE_START)) / TANGRAM_FADE_RANGE), 0., color.a);
+            color = mix(
+                color,
+                v_outline_color,
+                (1. - smoothstep(v_outline_edge - v_aa_factor, v_outline_edge + v_aa_factor, 1.-point_dist)) * step(.000001, v_outline_edge)
+            );
+            color.a = mix(color.a, 0., (smoothstep(1. - v_aa_factor, 1., point_dist)));
```
javisantana: Hey, sorry to open this conversation but I don't fully understand this.

`v_aa_factor` is calculated taking like `TANGRAM_PX_FADE_RANGE * 1.0/ length(a_shape.xy / 256.)`. 

This opens some questions:
- even `v_aa_factor` is a varying, does it vary? the v_aa_factor calculated for each vertex should be the same (I'm talking just about the quads used to render a point)
- when point size is <= 2px `v_aa_factor` is greater than 1 so the `smoothstep(1. - v_aa_factor, 1., point_dist)` is kind of misleading
- if `a_shape` is [+-128, +-128], then should n't be `TANGRAM_PX_FADE_RANGE * 1.0/ length(a_shape.xy / 128.)` or something like `max(1.0, TANGRAM_PX_FADE_RANGE * 1.0/ length(a_shape.xy / 256.))` to avoid aliasing on 1px points? 

Am I misunderstanding `v_aa_factor` usage?

thanks!

PR #516, opened 2017-03-13
Oriented offsets for boundary lines

dmvaldman:
(cleaned up and rebased version of previous PR https://github.com/tangrams/tangram/pull/506)

This PR addresses #471 where we were not previously using the orientation of a line to modify an offset of a label. This allows a line label to remain on a specific side (inside or outside) of a polygon as it is placed around the polygon's boundary.

The offset will follow the geometry orientation if both an offset is specified and the text_source is given as an object with a `left`, and/or `right` key. For example

```
draw:
  text:
    text_source:
      left: 'name:left'
      right: ‘name:right'
    offset : [0, 10px]
```

tangram-1489439469844

### DIFF
--- a/src/styles/text/text.js
+++ b/src/styles/text/text.js
@@ -92,15 +92,32 @@ Object.assign(TextStyle, {
             return;
         }
 
-        q.feature = feature;
-        q.context = context;
-        q.layout.vertex = false; // vertex placement option not applicable to standalone labels
+        // text can be an array if a `left` or `right` orientation key is defined for the text source
+        // in which case, push both text sources to the queue
+        if (q instanceof Array){
+            q.forEach(function(q){
+                q.feature = feature;
+                q.context = context;
+                q.layout.vertex = false; // vertex placement option not applicable to standalone labels
 
-        // Queue the feature for processing
-        if (!this.tile_data[tile.key] || !this.queues[tile.key]) {
-            this.startData(tile);
+                // Queue the feature for processing
+                if (!this.tile_data[tile.key]) {
+                    this.startData(tile);
+                }
+                this.queues[tile.key].push(q);
+            }.bind(this));
         }
-        this.queues[tile.key].push(q);
+        else {
+            q.feature = feature;
+            q.context = context;
+            q.layout.vertex = false; // vertex placement option not applicable to standalone labels
+
+            // Queue the feature for processing
+            if (!this.tile_data[tile.key] || !this.queues[tile.key]) {
+                this.startData(tile);
+            }
+            this.queues[tile.key].push(q);
+        }
 
         // Register with collision manager
         Collision.addStyle(this.name, tile.key);
@@ -186,7 +203,6 @@ Object.assign(TextStyle, {
             if (text_info.text_settings.can_articulate){
                 var sizes = text_info.size.map(function(size){ return size.collision_size; });
                 fq.layout.no_curving = text_info.no_curving;
-                fq.layout.text = fq.text;
                 feature_labels = this.buildLabels(sizes, fq.feature.geometry, fq.layout, text_info.total_size.collision_size);
             }
             else {
--- a/src/styles/text/text.js
+++ b/src/styles/text/text.js
@@ -186,6 +186,7 @@ Object.assign(TextStyle, {
             if (text_info.text_settings.can_articulate){
                 var sizes = text_info.size.map(function(size){ return size.collision_size; });
                 fq.layout.no_curving = text_info.no_curving;
+                fq.layout.text = fq.text;
                 feature_labels = this.buildLabels(sizes, fq.feature.geometry, fq.layout, text_info.total_size.collision_size);
             }
             else {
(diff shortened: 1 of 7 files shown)


nvkelso:
Nice! It's working 100% for me as I adapt Bubble Wrap to use the new functionality :) 

I filed a new issue for the Idaho label flipping problem we noticed, https://github.com/tangrams/tangram/issues/517. That shouldn't hold up release of this feature.

Can you also reply here with the new issue tracking colliding the left and right labels as a block? That is a feature I'd like as an add on after this release as even around Four Corners labels can look a little funky now.

9/37.1745/-108.8663

screen shot 2017-03-13 at 16 22 48

src/styles/text/text.js
```diff
@@ -92,15 +92,32 @@ Object.assign(TextStyle, {
             return;
         }
 
-        q.feature = feature;
-        q.context = context;
-        q.layout.vertex = false; // vertex placement option not applicable to standalone labels
+        // text can be an array if a `left` or `right` orientation key is defined for the text source
+        // in which case, push both text sources to the queue
+        if (q instanceof Array){
```
bcamper: Nitpick but I think `Array.isArray()` was added (in ES5) deal with edge cases around `instanceof Array`. While maybe unlikely they would affect us, I feel like it's better to be consistent (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#instanceof_vs_isArray).

src/styles/text/text_labels.js
```diff
@@ -31,31 +31,58 @@ export const TextLabels = {
...
         let sizes = this.texts[tile.key][text_settings_key] = this.texts[tile.key][text_settings_key] || {};
 
-        // unique text strings, grouped by text drawing style
-        if (!sizes[text]) {
-            // first label with this text/style/tile combination, make a new label entry
-            sizes[text] = {
-                text_settings,
-                ref: 0 // # of times this text/style combo appears in tile
-            };
+        if (text instanceof Object){
```
bcamper: Related to earlier comment, I've been using `typeof` for checking `object`/`function`/`string`/`number`, but I don't know that I have a good reason for that vs. `instanceof` (for the object and function cases anyway... can't check string vs. number with `instanceof`). Do you know of any reasons to use one vs. the other? https://jbkflex.wordpress.com/2013/05/09/javascript-typeof-and-instanceof-the-concept-and-differences/

src/styles/text/text.js
```diff
@@ -92,15 +92,32 @@ Object.assign(TextStyle, {
...
+                q.feature = feature;
+                q.context = context;
+                q.layout.vertex = false; // vertex placement option not applicable to standalone labels
+
+                // Queue the feature for processing
+                if (!this.tile_data[tile.key]) {
+                    this.startData(tile);
+                }
+                this.queues[tile.key].push(q);
+            }.bind(this));
```
bcamper: Let's use ES6 arrow functions instead of `bind(this)` for consistency please, e.g.
```
q => {
  ...
}
```

dmvaldman: can do. old habits die hard. what about arrow functions when `this` does not need to be bound?

src/styles/text/text_labels.js
dmvaldman: i just choose the one that is fastest and most memory efficient. these things don't have much of a performance impact in the grand scheme of things, but I still prefer them

dmvaldman: which boils down to `instanceof` for everything that it works for, and `typeof` for numbers and strings

src/styles/text/text.js
dmvaldman: If I were to choose, I'd choose `instanceof` for all array checking. The places where it doesn't work is for iframes... which, shouldn't be a problem for us?

src/styles/text/text.js
bcamper: I've been using arrow functions for any of these callbacks, but in the case where the binding is irrelevant, it's just to be concise, less important.

src/styles/text/text_labels.js
bcamper: That makes sense. I won't stress over making it consistent, we can swap out `instanceof` in future cases where it works.

src/styles/text/text.js
bcamper: Probably not. We do some weird things with iframes in tools like Differ, but I don't think that's relevant for the cases we're discussing here (and not part of the core library anyway).

src/styles/text/text_labels.js
```diff
@@ -319,8 +325,8 @@ export const TextLabels = {
 
         // repeat rules include the text
         if (layout.repeat_distance) {
-            if (text instanceof Array){
```
bcamper: @dmvaldman this was not getting triggered because the individual strings are passed into this function in separate calls, not as an array in a single call (that might have been true in an earlier version).

The change here is a bit messy but improves the repeat logic in some areas.

bcamper:

PR #521, opened 2017-03-15
update tile urls

meetar:


nvkelso:
Consider adding `url_params` with an `api_key` sample?

bcamper:
There's an example in the demo, which could also be re-appropriated for docs/readme:

https://github.com/tangrams/tangram/blob/master/demos/scene.yaml#L138-L144

meetar:
Yes, I'm not clear where in this PR `url_params` would be included –

 - the first example is a direct link to a json file on the tileserver, which should have an `&api_key` url param added
 - the second is the `setDataSource()` method, which takes a `config` object, into which  `url_params` could go, but I don't know how that should behave in situations where the `global.mapzen_sdk_api_key` is set
 - the third is a test fixture, which I presume could have one added?
 - the fourth is a sample yaml file, which definitely needs it
 - the last is another test fixture, which is only checking for the source url, and which I don't believe requires `url_params` to pass.

Were there any other places that need it?

PR #523, opened 2017-03-17
Rename layer `visible` -> `enabled` & don't allow descendant layers to override

bcamper:
As discussed in https://github.com/tangrams/tangram-docs/issues/198:

- Renames the layer-level `visible` property to `enabled`, to prevent confusion with the *`draw` group-level `visible` property*, which continues to exist as-is.
  - Provides backwards compatibility for the previous `visible` property name: if the `enabled` parameter is undefined for a given layer, any value for `visible` will be used instead.
- Changes the Tangram JS treatment of the layer `enabled` (nee `visible`) parameter to match that of Tangram ES: once set by a layer, the parameter *cannot* be overridden by any descendant layers.
  - This provides a mechanism to cull large portions of the layer tree, with the guarantee that they will stay hidden. This is especially useful as more optional overlay-style layers are added to the map (bike, transit, etc.), allowing the engine to efficiently cull these when they are off.
  - For cases where it's useful for sub-layers to control visibility, the `draw` group-level `visible` property continues to provide this fine-grained control; since all `draw` group properties are merged, and descendant layer `draw` groups can override any properties defined by ancestors, this behavior continues to work as-is.

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -212,10 +212,10 @@ Enjoy!
                         }
                     }
 
-                    scene.config.layers.earth.visible = true; // some custom shaders may need to render earth
+                    scene.config.layers.earth.fill.enabled = true; // some custom shaders may need to render earth
                 }
                 else {
-                    scene.config.layers.earth.visible = false; // don't need earth layer in default style
+                    scene.config.layers.earth.fill.enabled = false; // don't need earth layer in default style
                 }
             }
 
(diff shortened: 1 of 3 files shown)


bcamper:
@matteblair @nvkelso let's confirm this is the behavior we want? I'd like to include any such syntax changes in 0.12.0 rather than putting them off with further behavior divergence (and backwards compatibility transition period for previous `visible` property).

matteblair:
:+1: Looks good to me!

bcamper:

PR #527, opened 2017-03-24
Make source map filename configurable

bcamper:
This makes the source map filename used for workers a parameter passed to the `quine.js` plugin. This addresses two issues:

- Fixes an issue where the source maps would not be correct for worker threads if the Tangram bundle was renamed to a different filename, since it was relying on the origin (via `document.currentScript`) to construct the source map URL for workers.
- Allows the `budo` live-reload process to write to a temp source map name (which is in `.gitignore`), preventing it from conflicting with the official build filenames in `dist/`.

The downside is that the source map name has to be explicitly passed in to the plugin, but this was already true for the call to `mapstraction`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 
 # build & test artifacts
 node_modules
+dist/tangram.debug.temp.js.map
 
 # a make target for npm install
 .npm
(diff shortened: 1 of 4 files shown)


bcamper:
This has been bugging me all weekend while trying to switch branches, so I'm going to go ahead with it, even though it makes things a little more complicated.

Issue #517, opened 2017-03-13
Label orientation when almost vertical line should always be consistent


nvkelso:
This is a followup for https://github.com/tangrams/tangram/pull/516. I think this is a longer standing text bug not associated with the left/right labels work, so filing a new issue for it.

When adding the left/right labels in Bubble Wrap I noticed that some labels along the vertical Idaho boundary line are flipped when compared to neighbouring labels (boxed in red, with red arrow next to it). 

Labels on a (near) vertical line should all be oriented the same direction (the same as the blue arrows). Maybe a tiny JS math rounding bug?

9/48.6193/-115.9140
Using scene file in this [GIST](https://gist.github.com/nvkelso/8ef4520edc64ab1d6511bde8af7f9d97).

<img width="515" alt="idaho-label-orientation" src="https://cloud.githubusercontent.com/assets/853051/23879182/c28a78b4-0808-11e7-8cd8-cadeb9d9a0aa.png">

bcamper:
Makes sense. Vertical lines are the inflection point between left/right labels, and the smallest deviation (I'm guessing more likely in tile geometry, either because the geometry isn't perfectly straight in the source, or simplification/encoding rounds it) can cause it to go one way or the other.

I think a simple fix is to just pick a small tolerance range (few degrees?) within which nearly vertical lines will consistently use the same text direction. I guess we will need to apply the tolerance for lines pointing up and for those pointing down.

bcamper:
@dmvaldman can you take a look at this - it can be independent of any upcoming releases.

pnorman:
> I think a simple fix is to just pick a small tolerance range (few degrees?) within which nearly vertical lines will consistently use the same text direction. I guess we will need to apply the tolerance for lines pointing up and for those pointing down.

Say it's 5 degrees, and lines between 355 and 5 use the same orientation as a positive angle would normally result in. Won't this then lead to another problem with roads at approximately 355 degrees? 354.9 goes one way and 355.1 goes another.

I guess this is an improvement because there are a lot more almost north/south roads than almost 5 degree roads.

nvkelso:
Maybe it’s a number like 1 or 2 instead of 5?

pnorman:
> Maybe it’s a number like 1 or 2 instead of 5?

The same thing applies - there's still a point where it suddenly changes. I think it's probably an improvement since it'd be fewer streets, but if you happen to be in a city with a grid layout aligned to the direction switch point, it'll still look odd.

[...]

PR #532, opened 2017-03-29
Uniform vertical labels

dmvaldman:
Makes all vertical labels point up, instead of be a mix of up and down. Fix for https://github.com/tangrams/tangram/issues/517

www gifcreator me_djfoq9


bcamper:

PR #536, opened 2017-04-05
preRender and postRender are events instead of callbacks

dmvaldman:


bcamper:
This exposes an oddity that continues to plague Tangram naming...

Internally Tangram uses names `with_underscores`, but when instantiating the Leaflet layer (`Tangram.leafletLayer(...)`), I went with `camelCase`, to match the rest of Leaflet's options. You can see there are places where the naming is ["converted"](https://github.com/tangrams/tangram/blob/master/src/scene.js#L38) from the options on instantiation.

This is largely why we have the options `preUpdate` and `postUpdate` for function hooks... but when converting these to events (which makes sense), it would be more consistent to internally use the event names `pre_update` and `post_update` (for consistency with other events like `view_complete`).

Another thing we could do is transition the external options to `use_underscores`, and accept the existing `camelCase` variants for compatibility.

dmvaldman:
I just used the internal names that were previously there, but I could change them to be underscored instead of camelcased. I think for the surface area of Tangram that is accessible through Leaflet, we should match Leaflet's patterns.

PR #539, opened 2017-04-07
Meetar/demo refactor

meetar:
function scope: all the layers buttons were toggling "terrain"

enable earth: earth and bg were the same color, so the earth toggle didn't appear to do anything


dmvaldman:
Thanks!

PR #543, opened 2017-04-11
Fix #541 RTL curved labels when segments end on neutral character

### LINKED ISSUE #541: RTL curved label text flips order of punctuation/spaces
@mushon pointed this out to me on Twitter. I'll work on fixing this ASAP. Believe it's because we split words into 2-letter segments, and if one of the letters is RTL and the other neutral (like spaces, punctuation, etc) then the order between them is indeterminate (it would depend on what the next strongly-directioned character is). He provided the examples below:

Spaces
space_bad

Apostrophes
apostrophe

Quotes
quotes

Parenthesis
paranethesis

dmvaldman:
We detect if a broken up text segment ends on a neutral character. In this case we draw the characters individually (as separate glyphs) instead of clustering them. This should resolve the issue with curved labels for RTL text, but not for multi-line labels, which will require a different approach. This fixes part of the issue in #541.

Spaces:
www gifcreator me_joyyqh

Punctuation:
www gifcreator me_0tjsdn

### DIFF
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -596,12 +596,12 @@ function isTextShaped(s){
 
 // Right-to-left / bi-directional text handling
 // Taken from http://stackoverflow.com/questions/12006095/javascript-how-to-check-if-character-is-rtl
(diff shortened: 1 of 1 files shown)


dmvaldman:
@mushon If you'd like to test this, you can do so here: https://precog.mapzen.com/tangrams/tangram/rtl-fix/#16.966313008924754/32.07126714677083/34.7888861289828

(toggle the language to "default" for Hebrew labels)

It should be centered on the top image above. The multi-line labels are still not fixed, but we should have a different fix for that soon.

bcamper:

PR #544, opened 2017-04-11
Adds debug option to collect per-layer stats

bcamper:
Adds a `layer_stats` debug option, which will enable collection of feature and geometry stats per layer:

- Number of features that were built for each layer
- Number of GL geometries (triangles) that were built for each layer
  - GL geometry counts are further collected by rendering `style`, and grouped by style `base` (e.g. `polygons`, `lines` etc.)

These stats are collected in two formats:

- A flattened **list** that displays all of fully qualified layer names, e.g. `roads:path:bridge`
- A **tree** that aggregates counts at each level of the layer tree, with the ability to drill down into descendant levels

Counts for the tiles in the **current view** can be retrieved through `scene.debug.layerStats()` (a warning is logged if the `layer_stats` debug option is not enabled). This method returns an object with `list` and `tree` properties. Here's an example:

Flat `list` of layers:

screen shot 2017-04-11 at 1 18 06 pm

Opening stats for a specific layer:
screen shot 2017-04-11 at 1 17 51 pm

Top-level of `tree`:

screen shot 2017-04-11 at 1 20 20 pm

Drilling down into child `layers`:

screen shot 2017-04-11 at 1 22 49 pm

### DIFF
--- a/demos/main.js
+++ b/demos/main.js
@@ -46,6 +46,9 @@ Enjoy!
             // webGLContextOptions: { // explicitly add/override WebGL context options
             //     antialias: false
             // },
+            // debug: {
+            //     layer_stats: true // enable to collect detailed layer stats, access w/`scene.debug.layerStats()`
+            // },
             logLevel: 'debug',
             attribution: '<a href="https://mapzen.com/tangram" target="_blank">Tangram</a> | &copy; OSM contributors | <a href="https://mapzen.com/" target="_blank">Mapzen</a>'
         });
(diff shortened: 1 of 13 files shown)


bcamper:
@nvkelso this will help to understand hot spots and/or potential issues in the style layers

nvkelso:
:1st_place_medal:

bcamper:
Note that since features can match multiple layers (with the `draw` groups for all of them merged), a single feature may appear multiple times in these layer counts.

nvkelso:
But generally the top-level layers should be single count only?

bcamper:
That's a good question, I'm not sure, depending on how they are counted now... we have to infer some of these counts from lower-level data, so I need to look more closely to see what kind of double-counting might be happening.

Features will also be counted multiple times for each tile they appear in.

bcamper:

PR #547, opened 2017-04-17
Meetar/line offsets

meetar:


meetar:
This technique works pretty well for a constant offset, with one caveat – in places where multiple points create a curve with a radius _r_, if the offset is larger than _r_ the curve will become inverted:

angles2

So the style would have to ensure this doesn't happen by setting appropriate offsets at various zoom levels.

As I understand the situation there isn't really any way around this without some kind of smarter context-aware clipping algorithm à la [Clipper](http://www.angusj.com/delphi/clipper.php) or [this paper](https://hal.inria.fr/inria-00518005/document), neither of which would be ~easily compatible with our vertex shader… but I'm happy to be proven wrong :)

nvkelso:
We use Clipper for some geom validation tasks in Tilezen.

Love this illustration :)

screen shot 2017-04-17 at 12 16 14

demos/test.json
```diff
@@ -0,0 +1,30 @@
+{
```
bcamper: We should remove test files from final PR

demos/test2.json
```diff
@@ -0,0 +1,98 @@
+{
```
bcamper: Ditto let's not merge test files

src/gl/render_state.js
```diff
@@ -21,7 +21,7 @@ export default class RenderStateManager {
         this.defaults = {};
 
         // Culling
-        this.defaults.culling = true;
+        this.defaults.culling = false;
```
bcamper: Need to turn culling back on

src/styles/lines/lines.js
```diff
@@ -237,6 +241,7 @@ Object.assign(Lines, {
         draw.width = StyleParser.createPropertyCache(draw.width, StyleParser.parseUnits);
         draw.next_width = StyleParser.createPropertyCache(draw.width, StyleParser.parseUnits); // width will be computed for next zoom
         draw.z = StyleParser.createPropertyCache(draw.z, StyleParser.parseUnits);
+        draw.offset = StyleParser.createPropertyCache(draw.offset || 0, StyleParser.parseUnits);
```
bcamper: If we are going to only accept `px` units (which I think we decided on), we should enforce that, probably here (first time the parameter is processed). Will need to double-check, but I think we should be able to do that by checking (if `draw.offset` isn't null) that `draw.offset.value.units === 'px'`, and if not, log a warning and setting `draw.offset` to null. We may also want to add the ability to specify different default units for `parseUnits` though, so that offset values can have pixels as default units (if we decide we want that).

src/styles/polygons/polygons_vertex.glsl
```diff
@@ -30,6 +30,8 @@ attribute vec4 a_color;
     // z:  half-width of line (amount to extrude)
     // w:  scaling factor for interpolating width between zooms
     attribute vec4 a_extrude;
+    // xy: direction of line, for getting perpendicular offset
+    attribute vec4 a_offset;
```
bcamper: `offset` is only a `vec2`

src/styles/polygons/polygons_vertex.glsl
```diff
@@ -77,27 +79,36 @@ void main() {
...
     #ifdef TANGRAM_EXTRUDE_LINES
+        // vec2 extrude = a_extrude.xy / 256.; // values have an 8-bit fraction
         vec2 extrude = a_extrude.xy / 256.; // values have an 8-bit fraction
         float width = a_extrude.z;
         float dwdz = a_extrude.w;
-
-        // Adjust line width based on zoom level, to prevent proxied lines from being either too small or too big.
-        // "Flattens" the zoom between 1-2 to peg it to 1 (keeps lines from prematurely shrinking), then interpolate
-        // and clamp to 4 (keeps lines from becoming too small when far away).
+        vec2 offset = a_offset.xy; // values have an 8-bit fraction;
```
bcamper: I think 8-bit fraction comment is no longer accurate since we're not multiplying/un-multiplying this value anymore

src/styles/polygons/polygons_vertex.glsl
```diff
@@ -77,27 +79,36 @@ void main() {
     vec4 position = vec4(a_position.xy, a_position.z / TANGRAM_HEIGHT_SCALE, 1.); // convert height back to meters
 
     #ifdef TANGRAM_EXTRUDE_LINES
+        // vec2 extrude = a_extrude.xy / 256.; // values have an 8-bit fraction
```
bcamper: remove old line

src/builders/polylines.js
```diff
@@ -510,69 +563,81 @@ function addFan (coord, nA, nC, nB, uvA, uvC, uvB, isCap, context) {
...
-    addVertex(coord, nA, uA, context);
-    addVertex(coord, nB, uB, context);
-
-    var orientation = nA[0] * nB[1] - nA[1] * nB[0] > 0;
+    // ccw or cw bend?
+    var cross = eA[0] * eB[1] - eA[1] * eB[0] > 0;
+
+    if (cross) { // ccw bend
+        // pivot
+        addVertex(coord, Vector.neg(eC), false, uC, context, -1);
```
bcamper: Bevel joins are broken - expected third parameter of `addVertex()` should be the offset vector (line's 2d normal), but it's being set to `false` here (not sure what the correct vector value should be).

bcamper:
Follow-up work in:
https://github.com/tangrams/tangram/compare/meetar/line-offsets...premultiply-width
https://github.com/tangrams/tangram/compare/premultiply-width...mesh-variants

bcamper:
Merged to master via [`mesh-variants-rebase`](https://github.com/tangrams/tangram/compare/mesh-variants-rebase) branch

🎊

meetar:
> Merged to master via mesh-variants-rebase branch

Prepping the docs for this – is there an associated PR or commit I can refer to?

bcamper:
It's all internals beyond the line offsets themselves. (Separately there's related work for specifying texture and line dash patterns in `draw` blocks, but I think we already covered that for docs elsewhere.)

PR #549, opened 2017-04-18
Do not apply shader blocks on `points`-based styles to attached `text`

bcamper:
In `points`-based styles, Tangram JS has been applying any shader blocks to any `text` attachments for that style, e.g.:

```
styles:
  tinted:
    base: points
    shaders:
      blocks:
        filter: color.r += 0.5;

layers:
  ...
    draw:
      tinted:
        ... # shader blocks applied to these points
        text:
           ... # shader blocks ALSO applied to this text
```
This branch changes that behavior so that these shader blocks **will not** be applied to the attached `text`. This is for two reasons:

- While it can be useful to apply shader blocks to text for some effects (which is still possible with the standalone `text` style), it can be problematic to require the *same* blocks apply to both the point and text portions; this recently arose when working on adding tinted icons to Refill.
- Tangram ES already does *not* apply these shader blocks, so this is an existing incompatibility. This branch brings the two to parity.

The better longer-term solution is to allow any attached `text` to specify an explicit style to use instead, simile to the `outline` block in `lines`-based styles. However, the current Tangram JS pipeline for processing combined points/text complicates this, so that work will be deferred.

The larger code footprint here is because the fragment shaders for `points` and `text` styles were previously shared (with several conditional `#define`-based branches), but have been split into two separate files. This improves comprehension, especially for the standalone `text` shader. Separating the vertex shader, which is still shared, may also be worthwhile, but the benefit did not seem as clear at this time due to the amount of logic still shared between the two (tradeoff of comprehension vs. duplicated code). Minor shader/setup simplification was included in that change as well.

### DIFF
--- a/src/styles/lines/lines.js
+++ b/src/styles/lines/lines.js
@@ -102,7 +102,6 @@ Object.assign(Lines, {
         // Specify a line texture (either directly, or rendered dash pattern from above)
         if (this.texture) {
             this.defines.TANGRAM_LINE_TEXTURE = true;
-            this.defines.TANGRAM_ALPHA_TEST = 0.5; // pixels below this threshold are transparent
             this.shaders.uniforms = this.shaders.uniforms || {};
             this.shaders.uniforms.u_texture = this.texture;
             this.shaders.uniforms.u_texture_ratio = 1;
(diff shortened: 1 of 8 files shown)


bcamper:

PR #555, opened 2017-05-04
Make curved labels in more languages

mapmeld:
I saw Mapzen's [blog post](https://mapzen.com/blog/curved-labels) about curved labels awhile ago and was interested in how I could expand it to more languages

screen shot 2017-05-04 at 11 17 12 am

Burmese (pictured above) has some text-shaping - for example the word "ဟို" is three chars long. But it's similar to how "ý" can be two chars. Burmese doesn't connect all letters like Arabic or Devanagari. By counting grapheme clusters instead of chars, we can prevent both writing systems from being split up irregularly. That's what this pull request fixes.

The RegEx is long because it includes other writing systems (at least Latin alphabet, Devanagari and Tamil), but I have not enabled or tested them in curved labels yet.

In the future it would be possible to use [Intl.Segmenter or a polyfill](https://gist.github.com/inexorabletash/8c4d869a584bcaa18514729332300356).

Arabic is another level of complexity (choosing the right initial/medial/final/isolated form) but it might be possible based on previous work [in OSM iD](https://gist.github.com/mapmeld/556b09ddec07a2044c76e1ef45f01c60) and [circular-arabic](https://github.com/mapmeld/circular-arabic).

Edit: fixed screenshot, always preview

### DIFF
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -568,19 +568,19 @@ CanvasText.texcoord_cache = {};
 // Contextual Shaping Languages - Unicode ranges
 const context_langs = {
     Arabic: "\u0600-\u06FF",
-    Bengali: "\u0980-\u09FF",
-    Devanagari: "\u0900-\u097F",
-    Mongolian: "\u1800-\u18AF",
-    Tibetan: "\u0F00-\u0FFF"
+    Mongolian: "\u1800-\u18AF"
 };
 
 const accents_and_vowels = "[:\u0300-\u036F" + // Combining Diacritical Marks
+"\u0900-\u0903\u093A-\u094C\u094E\u094F\u0951-\u0957\u0962\u0963" + // Devanagari
+"\u0981-\u0983\u09BC\u09BE-\u09CC\u09D7\u09E2\u09E3" + // Bengali
 "\u0A01-\u0A03\u0A3C-\u0A4C\u0A51" + // Gurmukhi
 "\u0A81-\u0A83\u0ABC\u0ABE-\u0ACC\u0AE2\u0AE3" + // Gujarati
 "\u0B01-\u0B03\u0B3C\u0B3E-\u0B4C\u0B56\u0B57\u0B62\u0B63" + // Oriya
 "\u0B82\u0BBE-\u0BCD\u0BD7" + // Tamil
 "\u0C00-\u0C03\u0C3E-\u0C4C\u0C55\u0C56\u0C62\u0C63" + // Telugu
 "\u0D82\u0D83\u0DCA-\u0DDF\u0DF2\u0DF3" + // Sinhala
+"\u0F18\u0F19\u0F35\u0F37\u0F39\u0F3E\u0F3F\u0F71-\u0F83\u0F86\u0F87\u0F8D-\u0FBC\u0FC6" + // Tibetan
 "\u102B-\u1038\u103A-\u103E\u1056-\u1059\u105E-\u1060\u1062-\u1064\u1067-\u106D\u1071-\u1074\u1082-\u108D\u108F\u109A-\u109D" + // Burmese
 "\u1A55-\u1A5E\u1A61-\u1A7C" + // Tai Tham
 "\u1DC0-\u1DFF" + // Combining Diacritical Marks Supplement
@@ -589,7 +589,7 @@ const accents_and_vowels = "[:\u0300-\u036F" + // Combining Diacritical Marks
 "\u0C80-\u0C83\u0CBC\u0CBE-\u0CCC\u0CD5\u0CD6\u0CE2\u0CE3" + // Kannada
 "\u0EB1\u0EB4-\u0EBC\u0EC8-\u0ECD" + // Lao
 "]";
-const combo_characters = "[\u094D\u09CD\u1039\u17D2\u0A4D\u0ACD\u0C4D\u0CCD\u0B4D\u1A60\u1A7F\u0DCA]";
+const combo_characters = "[\u094D\u09CD\u0A4D\u0ACD\u0B4D\u0C4D\u0CCD\u0F84\u1039\u17D2\u1A60\u1A7F]";
 const graphemeRegex = new RegExp("^.(?:" + accents_and_vowels + "+)?" + "(" + combo_characters + "\\W(?:" + accents_and_vowels + "+)?)?");
 
 let reg_ex_shaping = '[';
@@ -641,7 +641,7 @@ function splitLabelText(text, rtl){
         let testText = text;
         let graphemeCount = 0;
 
-        for (graphemeCount; graphemeCount < codon_length && testText.length > 0; graphemeCount++) {
+        for (graphemeCount; graphemeCount < codon_length && testText.length; graphemeCount++) {
             let graphemeCluster = (graphemeRegex.exec(testText) || testText)[0];
             segment += graphemeCluster;
             testText = testText.substring(graphemeCluster.length);
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -580,7 +580,7 @@ const accents_and_vowels = "[:\u0300-\u036F" + // Combining Diacritical Marks
 "\u0B01-\u0B03\u0B3C\u0B3E-\u0B4C\u0B56\u0B57\u0B62\u0B63" + // Oriya
 "\u0B82\u0BBE-\u0BCD\u0BD7" + // Tamil
 "\u0C00-\u0C03\u0C3E-\u0C4C\u0C55\u0C56\u0C62\u0C63" + // Telugu
-"\u0D82-\u0D83\u0DCF-\u0DDF\u0DF2\u0DF3" + // Sinhala
+"\u0D82\u0D83\u0DCA-\u0DDF\u0DF2\u0DF3" + // Sinhala
 "\u102B-\u1038\u103A-\u103E\u1056-\u1059\u105E-\u1060\u1062-\u1064\u1067-\u106D\u1071-\u1074\u1082-\u108D\u108F\u109A-\u109D" + // Burmese
 "\u1A55-\u1A5E\u1A61-\u1A7C" + // Tai Tham
 "\u1DC0-\u1DFF" + // Combining Diacritical Marks Supplement
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -569,7 +569,6 @@ CanvasText.texcoord_cache = {};
 const context_langs = {
     Arabic: "\u0600-\u06FF",
     Bengali: "\u0980-\u09FF",
-    Burmese: "\u1000-\u109F",
     Devanagari: "\u0900-\u097F",
     Khmer: "\u1780-\u17FF",
     Gujarati: "\u0A80-\u0AFF",
@@ -582,6 +581,9 @@ const context_langs = {
     Telugu: "\u0C00-\u0C7F",
     Tibetan: "\u0F00-\u0FFF"
 };
+
+const accents_and_vowels = "[:\u0300-\u036F\u0902\u093E-\u0944\u0947\u0948\u094B\u094C\u0962\u0963\u0981\u09BC\u09BE-\u09C4\u09C7\u09C8\u09CB\u09CC\u09D7\u09E2\u09E3\u0BBE-\u0BC2\u0BC6-\u0BC8\u0BCA-\u0BCD\u0BD7\u102B-\u1032\u1036-\u1038\u103A-\u103E\u1056-\u1059\u1AB0-\u1AFF\u1DC0-\u1DFF\u20D0-\u20FF\uFE20-\uFE2F]";
+const combo_characters = "[\u094D\u09CD\u1039]";
 
 let reg_ex_shaping = '[';
 for (let key in context_langs){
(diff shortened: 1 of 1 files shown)


mapmeld:
New commit to add more languages and better explain where the language RegEx comes from.

Test coordinates:
- Burmese: 17.3/16.793161/96.169961
- Khmer: 16.3/11.544808/104.937004
- Lao: 19/19.88344/102.12752

India and Sri Lanka road data is overwhelmingly English, so I am having trouble finding roads on OSM named in the other languages: Gujarati, Gurmukhi, Kannada, Oriya, Sinhala, Tamil, Telugu.

As mentioned earlier, Devanagari and Arabic scripts are connected so they're beyond the scope of this fix

bcamper:
This is an awesome contribution!! I've been playing around with it and noticed a couple things:

1. Creating a new `RegExp` object for every character iteration is kind of intense and caused a noticeable slowdown (all this Canvas text measuring and rendering is already pushing it... :)

   I tried moving the grapheme characters check to a single compiled `RegExp`, with the first text character explicitly appended to the current segment text before the results of the regex match:

   ```
   // compiled regex to find grapheme cluster chars
   const grapheme_regex = new RegExp("(?:" + accents_and_vowels + "+)?" + "(" + combo_characters + "\\w(" + accents_and_vowels + ")?)?");
   ...

   // later, in function to find next grapheme cluster
   segment += text[0];
   let grapheme_cluster = text.substring(1).match(grapheme_regex);
   if (grapheme_cluster && grapheme_cluster[0] != null) {
      segment += grapheme_cluster[0];
   }
   ```
   I'm not yet familiar enough with these scripts and combining characters, but it seems like this should be OK since the original regex was just doing an explicit check for the first character, rather than any dependent match or lookahead/behind logic (`new RegExp(testText[0] + "(?:" + ...`)? After some fiddling and spot-checking, it seems to produce the same results as your original. Anything I'm missing here?

2. Had problems with labels that had special regex characters (like parentheses) inside them breaking the regex. This can be fixed by escaping them, but is eliminated by the changes above in any case.

mapmeld:
@bcamper  - agreed, this looks better and doesn't cause any language issues.  The pattern of a grapheme cluster can be: [character], [character][additional accents/signs], or the more complex one being [character][additional accents/signs][character-combining mark / virama][2nd character][2nd character's accents/signs].   If there were complex Emoji on OSM then we'd need more.

I was thinking more on this and would ask your input on two additional changes:

* capturing the first character of the segment by starting the RegEx with . instead of adding it on a separate line. In this case we can always expect to add at least one character to the segment string.
* changing from ```str.match(grapheme_regex)``` to ```grapheme_regex.exec(str)``` to only returns the first result, which is the only one we need

bcamper:
Thanks - the single regex looks good. I'm still debugging a case where it goes into an infinite loop, even though I'd expect it to always take at least one character from the `.` - fun :)  

By setting `scene.setIntrospection(true)`, you can get HTML tooltips (and thus theoretically corrected rendered by the browser) for everything with a `name`. This is handy for comparing to the Tangram rendering. This at least allows a layperson who doesn't read the script to do a visual confirmation.

I was spot-checking Khmer and came across a case that looked wrong (see last two characters):

Bad (curved label):
khmer bad

Good (straight label):
khmer good

I know nothing about Khmer but I'm definitely interested to understand what's going on here. I see that you had it defined as:
`"\u17B4-\u17D1\u17D3" + // Khmer`

There's a conspicuous single missing character in the sequence there... 
screen shot 2017-05-08 at 5 40 29 pm
via http://www.unicode.org/charts/PDF/U1780.pdf

I tried adding that character to the range (`"\u17B4-\u17D3" + // Khmer`), and voila, it seems to work! 

screen shot 2017-05-08 at 5 49 58 pm

Do you know why it wasn't present? Intentional/mistake?

I'm going to keep reviewing these but wanted to run that one by you in the meantime.

mapmeld:
[\u17D2](http://www.fileformat.info/info/unicode/char/17d2/index.htm) is not a visible accent by itself but a "combo character". The regex shouldn't be splitting on it unless there's a combo character and then another letter which goes subscript.

Do you have coordinates where this label is? I would like to look at the text char by char.

bcamper:
Aha I should have checked that other regex. It should be visible right in the middle of your test coords: 16.30000/11.54480/104.93701

The string is "វិថី កោះពេជ្រ".

bcamper:
Similar deal with these two, near middle of 17.63334/11.54948/104.92715:

screen shot 2017-05-08 at 6 17 52 pm

screen shot 2017-05-08 at 6 17 32 pm

mapmeld:
I think I see the problem and (fortunately? I think?) a regex issue

This sequence [letter][combo character][letter] is somehow failing because the regex includes an optional accents/vowel signs after each letter, which for some reason is not being treated as optional.

mapmeld:
I'm thinking this is the right regex (using a . for both 1st letter and optional 2nd combo letter, also allowing multiple accents on the 2nd letter)

new RegExp(".(?:" + accents_and_vowels + "+)?" + "(" + combo_characters + ".(" + accents_and_vowels + "+)?)?")

mapmeld:
ok here's my latest idea

```javascript
const graphemeRegex = new RegExp("^.(?:" + accents_and_vowels + "+)?" + "(" + combo_characters + "\\W(?:" + accents_and_vowels + "+)?)?");
...
for (graphemeCount; graphemeCount < codon_length && testText.length; graphemeCount++) {
    let graphemeCluster = (graphemeRegex.exec(testText) || testText)[0];
    segment += graphemeCluster;
    testText = testText.substring(graphemeCluster.length);
}
```

bcamper:
Thanks @mapmeld - this is looking good, plus I've found a few unrelated issues in the process of testing it :)  Is this ready to go as far as you're concerned?

Question as I test and learn more about this. Here's an example Sinhala feature:
http://www.openstreetmap.org/way/368378417

Full text (`name:si`) is:
"ඉඹුල්ගස්දෙනිය - රඹුක්කන පාර"

The current grapheme cluster regex will split this up into these segments (I changed the `codon_length` to 1 for this, which will give us only one grapheme cluster per segment, instead of up to 2 as is the usual Tangram default):
["ඉ", "ඹු", "ල්ග", "ස්දෙ", "නි", "ය", " ", "-", " ", "ර", "ඹු", "ක්ක", "න", " ", "පා", "ර"]

@meetar and I were looking at some examples of groups like "ල්ග" and "ස්දෙ", where it seems visually like the characters could theoretically be broken down further. Is this just an effect of the current regex being conservative enough to catch cases like the Khmer issue we saw earlier? e.g. maybe it would be possible to create even more detailed / script-specific regexes to get more granularity, but this is a good middle-ground? Anyway I'm happy to stop where we are for now since this is a great improvement, just curious.

It's looking nice here:
17.14000/7.31549/80.32824
screen shot 2017-05-10 at 3 54 43 pm

I believe I've spot-checked all of the scripts you included except for Gujarati and Gurmukhi (will try those too if I can). BTW, if you want to just hard-code some label text for Tangram to render, to verify the output (if we can't find suitable OSM features), you can use a JS function in the `text_source` parameter to return an explicit string to render, e.g. `text_source: function() { return 'ඉඹුල්ගස්දෙනිය' }`

bcamper:
I was testing some Gurmukhi/Punjabi (http://www.openstreetmap.org/way/248011769, at 12.41000/31.13213/75.06559), and I noticed this script does have the bar running across the top -- do you know if it's OK for the continuity of that to be broken across segments like this? In any case, Google also appears to do that with Punjabi labels...

screen shot 2017-05-10 at 5 17 16 pm

mapmeld:
You're right - that's wrong how the Sinhala text was blocking there. I looked at the Unicode chart and fixed two parts of my Sinhala text range, which appears to fix the problem.

I don't know enough about Gurmukhi/Punjabi to say whether this is good or bad.  I think I might remove it out of abundance of caution

(did a merge to update with master branch)

src/styles/text/canvas_text.js
```diff
@@ -580,7 +580,7 @@ const accents_and_vowels = "[:\u0300-\u036F" + // Combining Diacritical Marks
 "\u0B01-\u0B03\u0B3C\u0B3E-\u0B4C\u0B56\u0B57\u0B62\u0B63" + // Oriya
 "\u0B82\u0BBE-\u0BCD\u0BD7" + // Tamil
 "\u0C00-\u0C03\u0C3E-\u0C4C\u0C55\u0C56\u0C62\u0C63" + // Telugu
-"\u0D82-\u0D83\u0DCF-\u0DDF\u0DF2\u0DF3" + // Sinhala
+"\u0D82\u0D83\u0DCA-\u0DDF\u0DF2\u0DF3" + // Sinhala
```
bcamper: Should `\u0DCA` be removed from `combo_characters`?

bcamper:
@tallytalwar says that the Gurmukhi/Punjabi example is OK and that it's fine for the "bar" at top to have some breaks when segments bend on curved labels. 

I've been doing some reading up on grapheme clusters and Unicode, but I'm definitely still unclear on some things :) (which is OK to merge this, I would just like to have a general idea of what's going on). Is there a simple explanation for the criteria of when a character goes in `accents_and_vowels` vs. `combo_characters`?

Also, I understand why Arabic is beyond the scope here, but less so Devanagari (given other similar scripts in your set) -- does it have shaping rules that make it unsuitable for this kind of regex approach? @tallytalwar is a Devanagari reader and wasn't aware of why it wouldn't apply, but also isn't familiar with the encoding/grapheme rules.

Similarly, the only other remaining scripts in our "curved label blacklist" are Bengali, Mongolian, Tibetan. Have you left them out because you're not familiar with them, and/or do you know if this could be extended to cover them?

Otherwise, I think we are ready to merge on our side, if you are? Thanks again!

mapmeld:
ok - thanks @tallytalwar for checking the Gurmukhi/Punjabi example. By my original thought process I could have left it out. If breaking the bar is OK on these labels, yes we can include Devanagari and Bengali.

**Accents and vowel signs** modify one base character (y -> ý, က -> ကြ). The character does not belong on its own ( ́ or ြ). In this family of scripts that we're talking about, each base character is a syllable (Devanagari न  = "na", नन = "nana"). People still need double-consonant sounds ("Chennai" as example). To make it work you "break the leg" of the 1st one to form न्न. In Unicode it looks like न + ् (virama) + न. The combinations don't need to be the same letter (न + ् + त = न्त). In Burmese it works a little differently; letters are stacked on top of each other. In Tamil there is just a dot and no merging.

This is my first time looking closely at Tibetan... it looks like we could apply the same rule similar to Burmese.

Mongolian is typically written vertically and [its Wikipedia article](https://en.wikipedia.org/wiki/Mongolian_script) contains no examples of it being written horizontally.

Also, Sinhala point is right. I'll remove that from ```combo_characters``` and add the other scripts later today, I think.

nvkelso:
Will be nice to see Tibetan in there :)

pnorman:
> Mongolian is typically written vertically and its Wikipedia article contains no examples of it being written horizontally.

We looked at Mongolian in OpenStreetMap Carto, and although the [issue doesn't contain much useful](https://github.com/gravitystorm/openstreetmap-carto/issues/2387) my recollection is that @russdeffner, who has spent time there, said horizontal was okay. I would say that if you can correctly render horizontal on straight lines you're ahead of many map renderers, and along the line for a curved line is fine.

mapmeld:
OK, latest commit covers everything mentioned here, except Mongolian, which I would like to leave for now. In addition to the vertical/horizontal issue, Mongolian letters have text-shaping more similar to Arabic, actually, so the word-chunking process would have similar issues.

For future reference on Tibetan: http://rishida.net/scripts/tibetan/

bcamper:
@mapmeld just wanted to mention that @tallytalwar and I saw what we think is a Devanagari issue when testing - will get you some more details tomorrow.

mapmeld:
Might be good to add Malayalam for SW India: http://thottingal.in/blog/2017/05/21/a-formal-grammar-for-malayalam-conjunct/

mapmeld:
Added Malayalam and included a fix which may resolve your Devanagari issue (if it is related to the concept in the Malayalam link?)

bcamper:
I believe that does fix the Devanagari issue! Hadn't gotten around to writing it up yet, and it looks like you probably already figured this out, but here is what I was seeing.

First of all there was a problem with the regex. For example, for the word "स्ट्रिंग" ("string"), our previous regex was returning these 3 characters as the first grapheme cluster:
`["स", "्", "ट"]`.

I tried checking this against the results for Chrome's [`Intl.v8BreakIterator`](https://code.google.com/archive/p/v8-i18n/wikis/BreakIterator.wiki) (as I understand it, close to a passthrough API to ICU), and it returns just these **2 characters** for the first cluster:
`["स", "्"]`

I was wondering if our regex was too greedy (to account for some of the other script features you adjusted it to support?), and it looked like we were actually taking a character from what V8 considered the *next* cluster:

Tangram: `[["स", "्", "ट"], ["्"]]`
V8: `[["स", "्"], ["ट", "्"]`

But even using `Intl.v8BreakIterator` for grapheme cluster detection did not render properly. Here, all roads have the same text ("string string string"), but note the curved label character order is different from the (correct) straight order:

screen shot 2017-05-16 at 10 43 56 pm

It looks like Is what we are seeing here an issue with what the link you provided calls conjuncts, and what this link calls [consonant clusters](http://rishida.net/scripts/indic-overview/#variantglyph)? Does this mean that in cases like this, the grapheme cluster itself is **not** sufficient as an independent visual unit, but requires additional script knowledge?

In any case, your changes **do** fix the example above, which now renders correctly:

screen shot 2017-05-21 at 4 32 29 pm

Your updated regex takes these sets of characters as the first two clusters:

`[["स", "्", "ट", "्", "र", "ि", "ं"], ["ग"]]`

@tallytalwar doe this make sense to you?

@mapmeld do you know if there are any downsides to the more expansive regex for other scripts?

mapmeld:
In the past I had been led to believe that joining more than two consonants was extremely rare, and bringing it up I had been asked to provide examples.  This article got me thinking that maybe you could be seeing the same thing on the map. In any case the RegEx now supports unlimited consonants joined there.

tallytalwar:
@bcamper yes that does make sense and [indic language overview](http://rishida.net/scripts/indic-overview/) here does help explain the scenario.

The question we raised the other day was what happens for a character set with 2 or more consecutive placed (half sounding) consonants, as this introduces the `virama`/`halant` characters. And the logic mentioned in @mapmeld's link above does take care of these situations.

PR #564, opened 2017-05-31
Split Arabic text for curved labels

mapmeld:
Let's do this (Arabic, Hebrew, and Divehi (Maldives) labels)

In Arabic text, some letters join to the left, and others don't. (Australia =  أستراليا for a reference). These letters are common and consistent, so we can reliably split a word on them without splitting the baseline. The only additional concern is making sure that any vowel-marks for the letter don't get split off and put at the start of the next segment. 

TODO on this: bidirectional / bilingual text. Potentially replacing with a good RegEx if we can find it

Hebrew and Divehi won't use this splitter - instead I'm just adding their vowel-marks to the code from last curved labels code

### DIFF
--- a/src/styles/text/canvas_text.js
+++ b/src/styles/text/canvas_text.js
@@ -567,11 +567,16 @@ CanvasText.texcoord_cache = {};
 
 // Contextual Shaping Languages - Unicode ranges
 const context_langs = {
-    Arabic: "\u0600-\u06FF",
     Mongolian: "\u1800-\u18AF"
 };
 
+// test http://localhost:8000/#16.72917/30.08541/31.28466
+const arabic_range = new RegExp(/^[\u0600-\u06FF]+$/);
+const arabic_splitters = new RegExp("[\u0622-\u0625\u0627\u062F-\u0632\u0648\u0671-\u0677\u0688-\u0699\u06C4-\u06CB\u06CF\u06D2\u06D3\u06EE\u06EF]");
+const arabic_vowels = new RegExp("^[\u0610-\u061A\u064B-\u065F\u0670\u06D6-\u06DC\u06DF-\u06E4\u06E7\u06E8\u06EA-\u06ED]+");
 const accents_and_vowels = "[:\u0300-\u036F" + // Combining Diacritical Marks
+"\u0591-\u05BD\u05BF\u05C1\u05C2\u05C4\u05C5\u05C7" + // Hebrew
+"\u07A6-\u07B0" + // Thaana
 "\u0900-\u0903\u093A-\u094C\u094E\u094F\u0951-\u0957\u0962\u0963" + // Devanagari
 "\u0981-\u0983\u09BC\u09BE-\u09CC\u09D7\u09E2\u09E3" + // Bengali
 "\u0A01-\u0A03\u0A3C-\u0A4C\u0A51" + // Gurmukhi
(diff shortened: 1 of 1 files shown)


bcamper:
Awesome! Two issues I am seeing so far:

1. The Arabic code path isn't being triggered for some strings. An example is "شارع أحمد باشا كمال". It looks like the line start and end (`^` and `$`) symbols in `arabic_range` are causing it to only match strings that are *single Arabic words* (no spaces, neutral chars, etc.). We only need to know if there are *any* Arabic chars in the string (bidi detection is done earlier in the flow and currently skips `splitLabelText()` entirely), so I believe that just `/[\u0600-\u06FF]+/` should work fine.

I noticed this because this string the last (first RTL) character in the circled word is incorrect (this is the result of it going down the regular grapheme cluster path):
screen shot 2017-05-31 at 10 23 18 am

2. When I updated the regex to not remove line start/end matches, it looks like that problem is fixed, however a new problem is introduced:
screen shot 2017-05-31 at 11 13 05 am

I believe what we're seeing here is a new manifestation of a previously fixed issue for RTL text (#541). When the text segment starts or ends with a neutral character (space or otherwise), it gets flipped to the wrong side during Canvas text rendering, making it join to the wrong word. You can see this in the circled word above, and also in the word from the first example too (leftmost character in the string incorrectly has a space after it). 

The current code to fix this issue (for Hebrew) is [here](https://github.com/mapmeld/tangram/blob/92c2bb9f9d118c758edd49cfb4650e7de5ee0167/src/styles/text/canvas_text.js#L676-L705), but it's being skipped in the new Arabic path (which builds the segments a bit differently). The neutral char code fix still feels sort of clunky to me, it would be ideal to factor it out and have a common implementation for all RTL text.

bcamper:
@mapmeld I have a fix to deal with the second issue above, which also allows us to remove the existing special-case code for it. Will push it today or tomorrow.

bcamper:
@mapmeld I've pushed a few changes here, extending your branch: https://github.com/tangrams/tangram/compare/mapmeld-curvy-labels

This revamps a bit of how we handle RTL and neutral characters to avoid the issue I mentioned above. It also slightly changes your Arabic regex to handle strings with spaces or other neutral chars.

Things look good for the places where I'd previously seen issues, but I haven't tested extensively yet.

mapmeld:
I'm not familiar with how the old RTL code was working exactly, but if you feel good about this, this could be the way to do it

Does it flip any text if you have bidirectional label (Latin + Arabic)?

bcamper:
I'm fairly confident this is working (as much as I can be from visual inspection as a non-reader anyway!) for Arabic.

For your bidi question, we still explicitly blacklist bidi text from curving (for now):
https://github.com/tangrams/tangram/compare/mapmeld-curvy-labels#diff-56e91e9082620f32a1ac931d75dc9957R71

So anything that either has bidi, or characters with unsupported curved shaping (only Mongolian now!) will skip curving, but still render as a straight label.

Going to merge this, then apply my additional changes on top. Thanks again!!

bcamper:
See follow-up work in:
https://github.com/tangrams/tangram/commit/0ce6c045e0962f9996fcbde5b68b8bf59ce6a622
https://github.com/tangrams/tangram/commit/a114f98115a5276008b73749e2c6eda33ec88fba
https://github.com/tangrams/tangram/commit/02f1de28153bd0bd0a591b5e0dfa28a07b85cf64

mapmeld:
thanks for making it happen! 👍

bcamper:

PR #566, opened 2017-06-12
Adds the ability to query features from tile data

bcamper:
This PR adds a `Scene.queryFeatures()` method for directly querying tile data (separate from the existing [`Scene.getFeatureAt()`](https://mapzen.com/documentation/tangram/Javascript-API/#getfeatureatpixel) method, which returns a single feature at a given pixel location).

## Syntax

The query method has the following signature and default parameters:
`queryFeatures({ filter = null, visible = null, unique = true, group_by = null, geometry = false })`

- `queryFeatures()` will query the features from tiles that **currently intersect the map viewport**.
  - The method returns a promise with an array of matching features, e.g. `queryFeatures().then(features => console.log(features));` will print the matching features to the console. (An array is the default return type, see `group_by` below for an alternate format.)
  - Each returned result will have a `properties` field for that feature. See the `geometry` option below to include feature geometry with the results.
  - Optional parameters described below will further limit the set of features returned.
  - Because tiles usually include some area outside the viewport, the `queryFeatures()` method can be thought of as roughly querying the visible area, but results may include some nearby features as well. This effect can also be caused tile over-zooming, for data sources with a [`max_zoom`](https://mapzen.com/documentation/tangram/sources/#max_zoom).
- `filter`
  - An optional filter object with [the same syntax](https://mapzen.com/documentation/tangram/Filters-Overview/) used for selecting features for styling in `layers`. If no `filter` is provided, all features will be returned (subject to other parameters defined below).
  - Example: `scene.queryFeatures({ filter: { $layer: 'pois', kind: 'restaurant' } })` will return all restaurants in the `pois` layer. 
  - Filters used with this method support an additional parameter, `$source`, which can be used to specify a data source name to filter features by. For example, `filter: { $source: 'mapzen' }` will only return features from the `mapzen` data source. (This parameter is not relevant for filters in `layers` because the data source is already explicitly selected by the `data` block.)
- `visible`
  - By default, *all* features in the tile source data will be returned, regardless of whether they were rendered in the scene or not.
  - If `visible: true`, the query will be restricted only to *features that were rendered* in the scene. Note that this means the feature matched a visible `draw` group within `layers`, and was not culled by collision detection (in the case of points or labels). It does not guarantee, however, that the feature is visible from any given view position; it may be drawn but underneath another feature with a higher `order` value, or it may be behind another 3d object such as a building.
    - Example: `scene.queryFeatures({ filter: { $layer: 'landuse', $geometry: 'polygon' }, visible: true })` will return all visible landuse polygons.
  - If `visible: false`, the query will be restricted only to *features that were NOT rendered* in the scene. This is useful for providing feedback on data that is related to the scene but which you don't want to actually visualize, or for understanding which features were not drawn due to collision (lack of available space on screen).
    - Example: `scene.queryFeatures({ filter: { $layer: 'pois' }, visible: false })` will return all POIs that were not drawn. 
- `unique`
  - The `unique` parameter indicates whether (and how) duplicate features should be included in the results. Valid values are `true` (default), `false`/`null`, a string providing a single feature property (`id`), or an array of feature properties (`['kind', 'operator']`)
  - The default value, `true`, will limit features to those that are entirely unique, meaning it will exclude features with identical properties (and if the `geometry: true` option described below is also set, features with identical geometry will be excluded as well). This is useful for avoiding duplicate features that may be included in multiple tiles, such as building polygons in Mapzen's vector tiles.
  - A `false` or `null` value will return *all* features, without any regard to their properties or geometry.
  - A single string, or array of strings, will only return features that are unique with regards to the named feature properties (and geometry if `geometry: true`). For example, `unique: id` will avoid features with duplicate ids in the results, and `unique: ['kind', 'operator']` will only return a single feature for each unique combination of `kind` and `operator` property values.
- `group_by`
  - The `group_by` parameter can be used to group results by one or more unique property values. Valid values are `false`/`null` (the default), a string providing a single feature property (`kind`), or an array of feature properties (`['kind', 'kind_detail']`).
  - When grouping properties are specified, the `queryFeatures()` results will be an object (still returned via a promise), where each key is a unique value according to the grouping criteria, and the value of each key is an array of results for that key.
  - When a *single* property is provided for grouping, the results key will be the value of that property. 
    - Example: `queryFeatures({ filter: { $layer: 'pois' }, group_by: 'kind' })` will return results like:
      - `{ station: Array(8), jewelry: Array(19), bus_stop: Array(5)... }`
  - When *multiple* properties are provided the grouping, the results key will be a *stringified object for each unique combination of values*.
    - Example: `queryFeatures({ filter: { $layer: 'pois' }, group_by: ['kind', 'kind_detail'] })` will return results like:
      ```
      {
        '{"kind":"restaurant","kind_detail":"chinese"}': Array(8),
        '{"kind":"restaurant","kind_detail":"french"}': Array(4),
        '{"kind":"restaurant","kind_detail":"indian"}': Array(3),
        ...
      }
      ```
    - The caller can optionally use `JSON.parse()` to parse these stringified grouping keys for additional processing.
- `geometry`
  - By default, `queryFeatures()` does not return feature geometry, nor consider geometry when determining if a feature is unique (see `unique` parameter above). Only `type` and `properties` will be returned in the query results.
  - When `geometry: true`, an additional `geometry` property will also be returned, containing the feature's geometry (as a GeoJSON [`geometry`](https://tools.ietf.org/html/rfc7946#section-3.1) object).
  - Including feature geometry in the result can be useful for further visualizations outside of Tangram, such as with Leaflet markets or SVG paths, or even direct exports of raw GeoJSON.

## Use Cases

Here are a few examples of ways the `queryFeatures()` parameters can be used.

**Get a list of unique subway lines in tile data**
```
scene.queryFeatures({ filter: { $layer: 'transit', kind: 'subway' }, group_by: 'ref' }).then(results => {
  console.log(Object.keys(results));
});

--->
["1", "2", "3", "4", "5", "6", "W", "R", "J", "Z", "PATH", "E", "C", "A", "D", "B", "Q", "N"]
```

**Get a count of visible POIs by kind, each time new tiles are rendered**
```
scene.subscribe({
  view_complete: function() { // when new tiles are rendered
    scene.queryFeatures({ filter: { $layer: 'pois' }, visible: true, group_by: 'kind' }).then(results => {
      for (let key in results) { 
        results[key] = results[key].length;
      }
      console.log(results);
    });
  }
});

--->
{
  "station": 6,
  "cafe": 20,
  "bank": 5,
  "restaurant": 34,
  "convenience": 10,
  "place_of_worship": 5,
  "bus_stop": 5,
  "hotel": 10,
  "museum": 1,
  "pub": 3,
  "bar": 2,
  "hospital": 1
}
```

**Add Leaflet markers for visible restaurant POIs**
```
scene.queryFeatures({ filter: { $layer: 'pois', kind: 'restaurant' }, visible: true, geometry: true }).then(results => {
  results.forEach(feature => {
    L.marker([feature.geometry.coordinates[1], feature.geometry.coordinates[0]]).addTo(map);
  });
});
```
screen shot 2017-06-12 at 1 43 48 pm

**Add Leaflet polylines (SVG) for major roads**
```
scene.queryFeatures({ filter: { $layer: 'roads', kind: 'major_road' }, unique: false, visible: true, geometry: true }).then(results => {
  results.forEach(feature => L.geoJSON(feature, {
    style: function () {
        return { color: 'red' };
    }
  }).addTo(map))
});
```
screen shot 2017-06-12 at 1 48 01 pm

### DIFF
--- a/src/geo.js
+++ b/src/geo.js
@@ -97,6 +97,46 @@ Geo.latLngToMeters = function([x, y]) {
     return [x, y];
 };
 
+// Transform from local tile coordinats to lat lng
+Geo.tileSpaceToLatlng = function (geometry, z, min, max) {
+    const units_per_meter = Geo.unitsPerMeter(z);
+    Geo.transformGeometry(geometry, coord => {
+        coord[0] = (coord[0] / units_per_meter) + min.x;
+        coord[1] = (coord[1] / units_per_meter) + min.y;
+
+        let [x, y] = Geo.metersToLatLng(coord);
+        coord[0] = x;
+        coord[1] = y;
+    });
+    return geometry;
+};
+
+// Copy GeoJSON geometry
+Geo.copyGeometry = function (geometry) {
+    if (geometry == null) {
+        return; // skip if missing geometry (valid GeoJSON)
+    }
+
+    let copy = { type: geometry.type };
+
+    if (geometry.type === 'Point') {
+        copy.coordinates = [geometry.coordinates[0], geometry.coordinates[1]];
+    }
+    else if (geometry.type === 'LineString' || geometry.type === 'MultiPoint') {
+        copy.coordinates = geometry.coordinates.map(c => [c[0], c[1]]);
+    }
+    else if (geometry.type === 'Polygon' || geometry.type === 'MultiLineString') {
+        copy.coordinates = geometry.coordinates.map(ring => ring.map(c => [c[0], c[1]]));
+    }
+    else if (geometry.type === 'MultiPolygon') {
+        copy.coordinates = geometry.coordinates.map(polygon => {
+            return polygon.map(ring => ring.map(c => [c[0], c[1]]));
+        });
+    }
+    // TODO: support GeometryCollection
+    return copy;
+};
+
 // Run an in-place transform function on each cooordinate in a GeoJSON geometry
 Geo.transformGeometry = function (geometry, transform) {
     if (geometry == null) {
@@ -110,11 +150,11 @@ Geo.transformGeometry = function (geometry, transform) {
         geometry.coordinates.forEach(transform);
     }
     else if (geometry.type === 'Polygon' || geometry.type === 'MultiLineString') {
-        geometry.coordinates.forEach(coordinates => coordinates.forEach(transform));
+        geometry.coordinates.forEach(ring => ring.forEach(transform));
     }
     else if (geometry.type === 'MultiPolygon') {
         geometry.coordinates.forEach(polygon => {
-            polygon.forEach(coordinates => coordinates.forEach(transform));
+            polygon.forEach(ring => ring.forEach(transform));
         });
     }
     // TODO: support GeometryCollection
--- a/src/scene.js
+++ b/src/scene.js
@@ -4,6 +4,7 @@ import debugSettings from './utils/debug_settings';
 import * as URLs from './utils/urls';
 import WorkerBroker from './utils/worker_broker';
 import subscribeMixin from './utils/subscribe';
+import sliceObject from './utils/slice';
 import Context from './gl/context';
 import Texture from './gl/texture';
 import ShaderProgram from './gl/shader_program';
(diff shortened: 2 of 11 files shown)


mountainMath:
That's really cool and will fit my needs for data visualizations where map data interacts with accompanying non map-based visualizations like charts and graphs. For my purposes, I will always include the geometry so that I can exclude features that don't intersect the current map view.

bcamper:
@mbloch I believe this would also be useful for the kinds of additional metadata you added to the election maps, e.g. vote % in visible area.

bcamper:
@jschleuss this may also be of interest for GeoJSON or SVG export from the Map Maker.

bcamper:

PR #571, opened 2017-06-20
Add optional pixel radius for feature selection

bcamper:
This PR adds an optional pixel radius when requesting features at/near a pixel location. Previously, feature selection only returned a feature at the *exact* pixel location provided, but there are common use cases where a larger radius is useful, such as:

- Mobile devices where input is not always so precise
- Features such as thin lines may be difficult to select

The following methods are affected:

- Direct scene interface:
  - `scene.getFeatureAt(pixel, { radius })`
- Leaflet layer interface for setting selection handlers
  -  `layer.setSelectionEvents (events, { radius })`
- Leaflet layer instantiation: 
   ```
   layer = Tangram.leafletLayer({
      scene: 'path/to/scene.yaml',
      events: {
         click: clickHandler,
      },
      selectionRadius: 10
   };
   ```

In all cases, `radius` or `selectionRadius` is an optional value in pixels (default matches existing behavior, where radius is zero). When using a radius, **the feature closest to the center point** will be returned. (As with existing feature selection, only features marked as `interactive: true` will register.)

### DIFF
--- a/src/leaflet_layer.js
+++ b/src/leaflet_layer.js
@@ -143,7 +143,7 @@ function extendLeaflet(options) {
 
                 // Setup feature selection
                 this.setupSelectionEventHandlers(map);
-                this.setSelectionEvents(this.options.events);
+                this.setSelectionEvents(this.options.events, { radius: this.options.selectionRadius });
 
                 // Add GL canvas to layer container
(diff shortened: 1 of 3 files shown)


bcamper:

Issue #565, opened 2017-06-06
Loading scene files from data/blob urls, pre-loaded string


Climax777:
Hi this is a feature request.

I'm trying to embed tangram with leaflet inside a C# WinForms application. The problem is that loading the scene file from cross-domain (ajax) is not supported with the WebBrowser control, which is probably also true for other use cases.

If it could be done to support loading the scene file from a blob/data url or even a pre-loaded string it would be a workaround for the issue.

bcamper:
Hi @Climax777 - you should have a couple options for this:

1. You should be able to convert your string to a blob URL yourself with `window.createObjectURL()`, and then feed it to Tangram.

2. Likely easier, you can directly give Tangram a JS object (rather than a string) to load.

In either case, you may have to account for the lack of a base path to resolve any relative resources against. You could make any resource paths in the scene absolute, or, you could .zip the scene and resources, and use `window.createObjectURL()` to send the entire .zip to Tangram.

Climax777:
Hi @bcamper 

I have a plunker [here](http://plnkr.co/edit/0SpkRzxUdLwQJ7G080Ze?p=preview) which tests the scenario. I can't get it to load the zip from the base64 zip file. Somehow it doesn't accept it as zip, because it tries to parse it as yaml.

Is there some way I can force it to load it as a zip?

bcamper:
Ohh... right, sorry, forgot that it tries to detect .zip via extension (currently anyway) :)

Have you tried the option of just sending a JS object?

Climax777:
The problem with that is the remote referenced assets would still be broken
due to the cross origin issue. How would I convert the yaml to json
anyways?

Would it be possible to implement a type override option in the
constructor?

bcamper:
Yes it's not difficult to add a format override... just will have to wait for a release. We can go ahead and do that though, this case has come up before.

Climax777:
That would be excellent. Please let me know how I can help out.

bcamper:
This will be addressed in #572 and released in the upcoming v0.13.0 release.

Climax777:
Thanks a million!

bcamper:
Released in [v0.13.0](https://github.com/tangrams/tangram/releases/tag/v0.13.0)

PR #572, opened 2017-06-21
Add optional `file_type` for scene.load()

bcamper:
This adds an optional `file_type` for the `scene.load()` method (and `sceneFileType` for the Tangram Leaflet layer constructor). This allows for zip files to be loaded without .zip extension (previously, the filename had to end in '.zip' to trigger zip-loading).

This is useful for cases like loading a blob URL that contains a zipped scene file (see #565). For example:

`scene.load(zip_blob_url, { file_type: 'zip' });`

or 
```
layer = Tangram.leafletLayer({
   scene: zip_blob_url,
   sceneFileType: 'zip',
   ...
});
```

This also updates internal naming of `scene.config_path` to more intuitive `scene.base_path`. The existing `scene.config_path` accessor will be maintained for now, but eventually deprecated. Options to `scene.load()` should also be passed as an object, e.g.:

`scene.load(scene_url, { base_path: 'https://site.com/scene/resources/' });

(The previous [public documented API](https://mapzen.com/documentation/tangram/Javascript-API/#loadscene_url-base_path) only had a single second argument, which was used for base path - this is still backwards compatible but should be discouraged.)

### DIFF
--- a/src/leaflet_layer.js
+++ b/src/leaflet_layer.js
@@ -159,8 +159,13 @@ function extendLeaflet(options) {
                 });
 
                 // Use leaflet's existing event system as the callback mechanism
-                this.scene.load(this.options.scene,
-                    { config_path: this.options.sceneBasePath, blocking: false }).then(() => {
+                this.scene.load(
+                    this.options.scene,
+                    {
+                        base_path: this.options.sceneBasePath,
+                        file_type: this.options.sceneFileType,
+                        blocking: false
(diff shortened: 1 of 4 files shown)


tallytalwar:
We also check filetype based on the extension of the file url. Maybe make sense for ES to also have an optional `fileType` with our `sceneLoad` API.

bcamper:

Issue #577, opened 2017-08-12
Point outline width is not sharp


nvkelso:
Following up on a PR comment https://github.com/tangrams/tangram/pull/508#discussion_r131962936 after it was closed:

From @javisantana:

Hey, sorry to open this conversation but I don't fully understand this.

`v_aa_factor` is calculated taking like `TANGRAM_PX_FADE_RANGE * 1.0/ length(a_shape.xy / 256.)`. 

This opens some questions:
- even `v_aa_factor` is a varying, does it vary? the v_aa_factor calculated for each vertex should be the same (I'm talking just about the quads used to render a point)
- when point size is <= 2px `v_aa_factor` is greater than 1 so the `smoothstep(1. - v_aa_factor, 1., point_dist)` is kind of misleading
- if `a_shape` is [+-128, +-128], then should n't be `TANGRAM_PX_FADE_RANGE * 1.0/ length(a_shape.xy / 128.)` or something like `max(1.0, TANGRAM_PX_FADE_RANGE * 1.0/ length(a_shape.xy / 256.))` to avoid aliasing on 1px points? 

Am I misunderstanding `v_aa_factor` usage?

thanks!

nvkelso:
I'm having the same problem today as I experiment with https://mapzen.com/tangram/view/?api=506/875#7.623/37.463/-122.311.

I'm finding the widths are:

- Too big compared to the pixel sizes specified in the scene file
- Blurry (not pixel perfect crisp)

<img width="259" alt="screen shot 2017-08-12 at 10 50 26" src="https://user-images.githubusercontent.com/853051/29243101-2f9ebbbc-7f4c-11e7-8eca-1206b337f6a2.png">


Here's I'm overcompensating by choosing small values that are still showing up larger than expected:

```
                outline:
                    color: black
                    width: [[4,0.25px],[6,0.25px],[8,0.25px],[10,1px]]
```

bcamper:
The basic answer to both of your questions is that these factors were mostly tuned by hand (by @patriciogonzalezvivo and I) to get a balance between aliasing (too blocky) and antialiasing (too blurry). The fade is currently configured to show a *full width line* of the width specified in `outline.width`, with the antialiasing feathering out from that. They can certainly be further tuned and @javisantana you may be right that limiting the factor on smaller points will help (have you tried? ;) 

[...]

PR #583, opened 2017-08-22
Alpha fix in point shader

dv343:
Related to https://github.com/tangrams/tangram/issues/577 , I think there are some problems with the current point shader.

I hope to have fixed them, but it's possible that there are still some cases with rendering problems.

Mapnik rendering:
mapnik
Rendering with the new changes:
new
Previous behavior:
previous

The changes made affects the anti-aliasing and the mixing between outline color and fill-color.
They fix the "blurriness" of the border between the fill-colored region and the outline region.
They also fix a corner case that Tangram didn't consider, points with semi-transparent borders. These points were rendered without blending the outline color with the fill color, which is probably not what we want, and it is clearly not what Mapnik does (see images).

The idea behind the antialiasing is basically the same, the distance is used to compute the alpha factor. 
The `antialias` function returns this factor given the UV coordinates and the radius. It uses `smoothstep` to interpolate between the nearest corner of the pixel and the opposite corner of the pixel. I've been using supersampling antialiasing as a reference, and I think that results are improved with small points and in the outline-fill border.

`v_size` is similar to `aa_factor`, but its meaning is more clear, it keeps the width, in pixels, of the quad.

The middleAlpha and mixColor variables are used in order to work with semi-transparent outlines that need to be shown over the fill-colored region. You can see that I have manually blended mixColor with `vec4 mixColor=outlineColor.a*outlineColor.rgba + (1.-outlineColor.a)*color.rgba*color.a;`. I don't know if this trick will work correctly in more complex cases when points are rendered over/below other points.

Changes in the vertex shader have been made to avoid missing samples due to not-conservative rasterization (see https://developer.nvidia.com/content/dont-be-conservative-conservative-rasterization). Before this change, some required (they would have an alpha value > 0) pixels weren't considered. This is because WebGL renders a pixel if, and only if, its sample location (the center of the pixel without MSAA) is inside the primitive. This case wasn't considered before, and if a pixel center was outside the circle and the quad, and the circle intersected the pixel, then, the pixel wasn't rendered when it needed to be. Using the Nvidia extension is obviously a bad solution, so I just made the quad bigger, by one pixel in each direction, and changed the UV coordinates to match this change.

The only problem I've seen with this change (despite that, as I've already said, the blending could be still broken in some cases), its that the outline size could be a little bit wrong. Points of the second row of the provided images, have a bigger outline with the new rendering. I could try to fix it by tweaking it by hand, but I would prefer to fix it without magic numbers. The `v_outline_edge` variable is probably the key. It probably needs to be remapped like the UV due to the change of size, but I would like to know for sure the meaning of this variable, could you explain it?


bcamper:
Thank you for the detailed explanation! Will take a look soon.

dv343:
Hi,
I've fixed the computation of alpha for the mixColor. I'm still not sure about the edge size, so, I wouldn't accept this pull-request as-is now, but I would like some feedback before improving it more.
Thank you!
Mapnik:
mapnik
New (after last fix):
new
Previous (pre pull-request):
prev

bcamper:
Thanks. For the edge size, `v_outline_edge` is set here:

https://github.com/tangrams/tangram/blob/master/src/styles/points/points.js#L168-L174

```
        style.outline_edge_pct = 0;
        if (style.outline_width && style.outline_color) {
            let outline_width = style.outline_width + 1;
            style.size[0] += outline_width; // bump outline by 1px to balance out antialiasing
            style.size[1] += outline_width;
            style.outline_edge_pct = outline_width / Math.min(style.size[0], style.size[1]) * 2; // UV distance at which outline starts
        }
```

You can see from the comment that it is intended to be the **UV distance at which outline starts** (so I believe that means the UV for the inner edge of the outline). The outline width itself is not sent to the shader, but the point size is enlarged to include the outline. All of this may be interacting undesirably (?) with your own adjustments to point size within the shader (on the other hand, what you have now does seem to be working pretty well :)  While it was moved around in a refactor, the outline edge logic and technique originated with the initial PR for adding point outlines: #472.

Notice that there is also a ` +1` on the outline size ("bump outline by 1px to balance out antialiasing"), which I had forgotten about; I believe it was added to thicken outlines which were perceived as too thin in some cases. Just removing this may address the issue of outlines being too big mentioned in #577.

Please take a look at the above and see if you can incorporate this all into your size adjustment logic. The original idea with doing the edge "percent" calculation CPU side was to avoid redundant calculation in the shader; on the other hand, since your adjustment happens in the vertex shader and is relatively simple, it is probably not a performance concern.

dv343:
Thank you very much, that "+1" was the problem, removing it has fixed many things, and I could simplify the shader more.
The only change (apart from adapting for the removal of the "+1") is that I passed the pixelOffset (now called aa_offset) computation to the vertex shader to improve performance.

This is the result:
Mapnik:
mapnik
New (latest commit):
new
Old (pre PR):
old

I don't see any problems at all on this scene after the "+1" removal.

However, I've used `a_shape` here: https://github.com/CartoDB/tangram-1/blob/004690915486c1bbec751769cc4ea00d62c9daaf/src/styles/points/points_vertex.glsl#L96-L97
I've assumed that it is the offest relative to the center of the quad. Please, check that I haven't messed it up in those lines.

Moreover, I don't know how or when exactly the `TANGRAM_SHADER_POINT` is defined or undefined. You probably want to check those `#ifdef`s too.

Lastly, by using `a_shape`, I stopped using `a_texcoord`, if this is OK, `a_texcoord` could be removed, improving memory bandwidth.

Thanks!

dv343:
I didn't mentioned it before, but I've also assumed that all points are rendered with a square quad (width=height).

One last thing... I've implemented the **over** operator in the shader because Tangram only renders each point once, despite having an outline. Supporting the rest of the blending modes can be done in two ways:
 - Keep the "render once", and just add a series of "if"s or "ifdef"s (better performance) to switch between blending modes.
 - Render each point with outline twice, with a simpler fragment shader (without blending), and just use WebGL blending flags.

What's your opinion about these?

bcamper:
Regarding `a_texcoord`, it cannot be removed generally because it is used for sprite rendering, where the UVs are mapped to a sub-area of a larger texture. Actually, there are some issues with your branch that prevent sprites from rendering entirely as-is, and which circumvent the logic for snapping labels to the pixel grid (for increased sharpness), but they are easily fixed.

We do have an upcoming refactor that would allow us to split the points-only geometries and the sprite geometries into separate VBOs (though they would technically share a single vertex layout and shader program, we could use GL's ability to set a static value for vertex attributes for the shader points, and leave them out of the VBO).

Can you expand on the blend mode suggestion more? We've considered adding more of the "Photoshop blend modes", @meetar has looked at that a bit (unfortunately "blend" is already a somewhat overloaded term in Tangram). We would likely go the `#ifdefs` route for these, which means they need to be set at the rendering `styles` level; less flexible, but more compatible with our current pipeline.

dv343:
Regarding those issues, do you plan to solve them or could you give me more information to solve them?

Regarding the upcoming refactoring, have you thought about using GL_POINTS? We could reduce vertex shader execution (and vertex shader memory bandwidth usage) by 4x-6x if we did. I know that for big points this is not a problem (fragment shader executions would be more than vertex ones), but for small points, this could be an important optimization.

About supporting more blend modes... The problem is that if we don't render each point with outline twice (one for the fill, one for the outline), then, we need to do the blending between fill and outline in the shader (in top of the WebGL blending flags). This would be the result with `#ifdefs`:
```
  vec4 middleColor;
  #ifdef TANGRAM_BLEND_SRC_OVER
    middleColor.rgb=outlineColor.a*outlineColor.rgb + (1.-outlineColor.a)*color.rgb*color.a;
    middleColor.a=outlineColor.a + (1.-outlineColor.a)*color.a;
    middleColor.rgb/=middleColor.a;
  #endif
  #ifdef TANGRAM_BLEND_ADD
   //This is the idea, but code could be broken (this hasn't been tested)
   middleColor.rgb=outlineColor.a*outlineColor.rgb + color.a*color.rgb;
   middleColor.a=outlineColor.a +color.a;
   //I think this division could be omitted if we work with pre-multiplied alpha
   middleColor.rgb/=middleColor.a;
  #endif
  #ifdef TANGRAM_BLEND_X
      ...
  #endif
```
This means that we will need one final shader program per blending mode, but I don't see a problem in this, and it will be certainly faster than using an uniform. 

You may be interested in this SVG related spec: https://www.w3.org/TR/2009/WD-SVGCompositing-20090430/
It is also the base for Mapnik's blending modes.

However, we won't be able to implement all of them for 2 reasons:
- WebGL blending flags are limited and they cannot support some of them. This is very hard to overcome and all my ideas rely on tricks that will degrade performance considerably.
- In-shader blending relies on the associativity of the blending. Some of the SVG blending modes are not associative. This is easy to solve with the already mentioned method of rendering each point twice.

bcamper:
I can take care of the sprite fixes, thanks (I'll cc you so you can see the changes, they are minor).

The PR looks good but before merging I want to take a closer look today at the effects of the enlarged point size for fragments at the edges of the quad. What's your expectation of the visual effect on the output?

We could consider `GL_POINTS` in the future (we've briefly discussed it before). My understanding is that they will drop the point once it goes out of the clip space, which is probably not OK for sprites (we could try it but I think it will be too conspicuous). This wouldn't be as much of an issue for smaller and denser points though, which sounds like the use case you are more concerned with.

The original use case for points in Tangram has been for base maps, where (at least in Mapzen's case), polygons and lines (in particular small road segments) have been the overwhelming majority of geometry. Point counts have predominantly been in the hundreds to thousand range, and haven't yet registered as a performance concern. I can see the possible advantage of other approaches though for much higher volume point displays, and we're interested in supporting that for dataviz needs.
 
We should open a separate issue for discussion of blend modes, where you and @meetar can compare notes. We had also looked briefly at some existing implementations like https://github.com/jamieowen/glsl-blend. I think the SVG/comp-op reference is definitely a good one, and we try to follow similar specs for things like line joins/caps. That said, our goals is not 100% support of compatibility with SVG or mapnik, we have to choose how far to go based on a balance of needs.

bcamper:
PS to get some debug info on geometry type count and buffer sizes, you can do this, which is useful for getting a sense of real-world scenarios:

```
scene.debug.geometryCountByBaseStyle()
scene.debug.geometrySizeByBaseStyle()
```

bcamper:
Another tool you should check out is our [Differ](https://github.com/tangrams/differ), which can run a series of scenes against two different Tangram builds and provide a visual comparison. (@meetar can also help with Differ questions.)

There are a bunch of differ test files here: https://github.com/tangrams/differ-tests

For example, here is a series of point-drawing tests (it's trying to point to a localhost build here, but you can plug in your own to test it out):
http://tangrams.github.io/differ/?1=https://tangrams.github.io/differ-tests/unit-tests/points/points01.json&2=&lib1=0.13.1&lib2=http://localhost:8000/dist/tangram.debug.js&go

bcamper:
@matteblair can you please comment on the outline antialiasing logic and/or anything else relevant to ES compatibility?

matteblair:
Took me a while to get my head around all the pieces of the problem here, but I've spent some time with the GLSL now and I have some suggestions!

1. I would suggest to inline `drawPoint` at its one call site - mainly to prevent potential name collisions with a user-defined function. To keep the local vars from leaking, you can put it in `{}`.

2. I took the body of `drawPoint` and made it as small and fast as I could, keeping the same results (I think*):
```glsl
#define TANGRAM_SMOOTH(edge, value) smoothstep(edge - v_aa_offset, edge + v_aa_offset, value)
// Distance into the point from the edge.
float dist_in = 1. - length(uv);
// Mask of outermost circle, either outline or point boundary.
float outer_alpha = TANGRAM_SMOOTH(0., dist_in);
// Combine the alpha values of the point and outline colors with masks for their coverage.
float fill_alpha = TANGRAM_SMOOTH(outline_edge * .5, dist_in) * color.a;
float stroke_alpha = (outer_alpha - TANGRAM_SMOOTH(outline_edge, dist_in)) * outlineColor.a;
// Apply alpha compositing with stroke 'over' fill.
color.a = stroke_alpha + fill_alpha * (1. - stroke_alpha);
color.rgb = mix(color.rgb * fill_alpha, outlineColor.rgb, stroke_alpha) / color.a;
```
There's less arithmetic here in general, but importantly there's only one `length()` and no branching!

(* = I tested this in Tangram ES with a pipeline that's similar, but not exactly the same.)

dv343:
Hi @matteblair ,

I've added most of you suggestions:
- I've changed the blending to your way, it saves us from doing a second `mix()`
- I've removed the redundant calls to `length()`
- I've inlined the `drawPoint` function
- I've removed the branch

However, I've left 2 things, hope you are Ok with this:
- I left the `antiliasias` function, but I changed the name to avoid polluting the namespace. I see your point, and the previous name could have been problematic, but the usage of macros instead of functions it's a little bit too extreme in my opinion, since macros have many problems.
- I haven't changed the `1.-smoothstep(...)` substraction. Your version has 2 less scalar substractions, but I think that the code is a little bit more complicated to understand that way. I would left that micro-optimization unless I had benchmark numbers about its usefulness.

Another thing, although I've removed the branch, it is not clear that this will increase performance. In fact, I think that it could actually improved it.
Branch disadvantages:

-  Not completely uniform. The `v_outline_edge` can change from one point to another of the same draw group. If this happens, then some wavefronts (AMD terminology) or warps (Nvidia terminology) will have to execute the branch on all threads of the thread group. This means that some threads that don't need to execute the branch will have to wait until the other threads complete the branch execution.
- Cost of dynamically uniform branching. There is a small cost even if all threads takes the same branch.

Branch advantages:
- Less work. Computing all the outline-related information is not very cheap, we have 2 smoothsteps and a few arithmetic ops.
- No `else` branch. This means that the cost of executing the branch in non active threads due to the way warps/wavefronts work is the same as executing it anyway without the branch. And in many cases, the branch won't be taken for all the threads on the group, giving a probably important performance advantage.

I think that the branch will improve performance considerably unless all or almost all points have an outline, in which case I would expect a minor (<5%) performance overhead. However, even though I know a little bit about some GPU architectures, it's very difficult to correctly predict performance, and I don't know much about mobile GPU architectures. Therefore, I would leave the branch for now, but benchmarking both approaches would be the best option.

matteblair:
@dv343 The changes look good to me, I appreciate the thought you've put into this! Your comments are fair, it's hard to determine further optimizations without more rigorous profiling. I think it would be ok with or without the branch since there doesn't seem to be an obvious performance difference.

dv343:
You're welcome!

I didn't saw the `#pragmas` below, so yes, we needed to put the code inside a block (`{}`). I've already made a commit with this.

However, @bcamper mentioned another problem with sprites that should be fixed too before merging. I don't know much about this, therefore, I hope you can make that last fix.

bcamper:
@dv343 yep I'll take care of those sprite issues, just moving a couple lines around I believe. thanks!

bcamper:
Hi @rochoa -- please don't include new Tangram dist builds in a pull request -- these are only updated when a new release is cut. I have marked this issue as scheduled for the v0.14.0 release (which is likely to be the next one barring any major bugs in v0.13.2).

Could you please rebase and remove those files? Thanks.

rochoa:
Sorry @bcamper, I was using that branch to test an internal integration branch and didn't realise there was an open PR using this branch.

I'm gonna revert the changes in this PR and use another branch for my purpose.

rochoa:
Is enough to revert those files? I mean, as I resolved the conflicts this PR now looks clean, but that's not the reality.

Sorry about the mess I just created.

bcamper:
I would rather remove those commits entirely. I can also do that when merging by cherry-picking your commits or rebasing, rather than simply merging the whole PR.

dv343:
I think I've removed them entirely from here. :slightly_smiling_face:

bcamper:
For reference, this branch as-is had two issues with sprite rendering:

- The `v_texcoord` varying had been removed. This is needed for sprites, where the UVs are not implicit.
- The quad enlargement for shader points was being applied to sprites, which broke the code which snaps them to the pixel grid (when the viewport is not moving).

Both are fixed in https://github.com/tangrams/tangram/commit/ac1c2be632c0a605ca944aedfc6f291b2ec0c824.

rochoa:
Thanks a lot, and sorry for the mess I created. 😅

PR #592, opened 2017-09-11
Points additive blend fix 

dv343:
Following https://github.com/tangrams/tangram/pull/583

This PR fixes additive blending for points in various ways:

- Antialias points with the same code that antialias `overlay` blended points
- Fixes outline-fill region
- Removes `discard` for additive blended points

These are screenshots of Mapnik, post PR and pre PR:
mapnik
post
prev

Basically it uses a `#if` to switch the blending and avoid the usage of `discard`.

I don't know much about that `discard`, but if it tries to be a "lower-quality substitute" for performance reasons, keep in mind that using `discard` degrades performance considerably, in a similar way to traditional blending, and for similar reasons (but with alpha blending anti-aliasing can be implemented).


bcamper:
Regarding the `discard`, my recollection is that it was there less for performance reasons, and more for the case where you want to use Tangram's `order` pipeline to interleave the points with the rest of the 3D space (different from `overlay` blend more behavior). That said, it's very old code and it is a good time to re-evaluate that behavior (and possible divergence from Tangram ES), so we will do that.

Issue #598, opened 2017-09-26
404 raster tiles turn black and cover anything underneath it in the sort order


burritojustice:
**TANGRAM VERSION:**

0.13.3

**ENVIRONMENT:**

10.12.4 (16E195), Chrome, Safari

**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**

define a limited (non-global) set raster tileset as a source in Tangram

test case:

https://meetar.github.io/raster-overlay/#15/40.7159/-73.9976

this came about trying to display tiled Sanborn maps:

https://mapzen.com/tangram/play/?#16.65417/37.80467/-122.29981

individual Sanborns generally cover a few city blocks, approximately 3x3 z17 tiles. even when lat/lon `bounds:` are included as a `url:` argument, empty tiles just outside the bounds are requested.

note that some tile renderers will generate a 'sparse' set of raster tiles that do not include empty tiles (common when the street grid is not north/south).

**RESULT:**

a 404 raster tile turns black and covers anything underneath it in the sort order, no matter what the raster style's blend mode is.

**EXPECTED RESULT:**

any 404 raster tiles should not impact the rest of the map.

bcamper:
Will be fixed in #599, for v0.14.0

bcamper:
Released in [v0.14.0](https://github.com/tangrams/tangram/releases/tag/v0.14.0).

PR #599, opened 2017-09-28
Skip style when raster textures fail to load

bcamper:
When one or more of a style's attached raster texture fail to load (404 or other error), throw out the style for that tile. This fixes the issue in #598, where unavailable raster tiles would cover up areas of the map underneath.


bcamper:
@matteblair @tallytalwar what is the current behavior for failed rasters in ES?

tallytalwar:
Hmm, ES also does a default black texture and will also exhibit the same behavior. Will file a corresponding issue on ES for this, once I confirm the exact behavior.

PR #606, opened 2017-10-09
Use @mapbox/vector-tile

matsprea:
npm WARN deprecated vector-tile@1.3.0: This module has moved: please install @mapbox/vector-tile instead

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "pbf": "1.3.7",
     "strip-comments": "0.3.2",
     "topojson-client": "tangrams/topojson-client#read-only",
-    "vector-tile": "1.3.0"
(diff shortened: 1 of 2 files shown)


pnorman:
Does this bump the minimum npm version required?

matsprea:
No, the @mapbox/vector-tile version is the same of vector-tile, the package has beed renamed.

pnorman:
> No, the version is the same of vector-tile, the package has beed renamed.

Not all versions of npm support the `@org` syntax, that's what I was asking about.

bcamper:
@pnorman thanks for pointing that out.

It looks like scoped packages are supported since npm 2.x, so that seems like a reasonable minimum version (considering the current version is 5.x).

PR #615, opened 2017-10-30
Set the minimum NPM version

pnorman:
#606 set an implicit NPM version requirement, this makes it explicit.

I'm not sure if NPM 1.x will error with this or not, but it at least sets the acceptable version range.


bcamper:
Thanks!

PR #626, opened 2017-11-29
currentScript should be truthy, not !== undefined

### LINKED ISSUE #625: document.currentScript can be null
**TANGRAM VERSION:**

* 0.14.1 (also master)

**ENVIRONMENT:**

*  firefox 59.0a1 (2017-11-29) (64-bit)
* chrome 62.0.3202.94 (Official Build) (64-bit)
* requirejs

**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**

* include latest mapzen from requirejs's require

**RESULT:**

Load fails with error:

    Uncaught TypeError: Cannot read property 'src' of null

**EXPECTED RESULT:**

Library loads

The problem is here: https://github.com/tangrams/tangram/blob/5df4e7edcc7953d4080566b941c148a163360eca/build/quine.js#L13

specifically this condition:

    document.currentScript !== undefined

in my case document.currentScript is null, not undefined.

the fix is simple, just change the condition to:

    !!document.currentScript

this works in my case.

marianoguerra:
in some cases document.currentScript can be null, on that case the
condition resolves to true and fails to get the src attribute from null
ending with an exception.

this change checks that currentScript is truthy (not null nor undefined)
in the condition

closes #625

### DIFF
--- a/build/quine.js
+++ b/build/quine.js
@@ -10,7 +10,7 @@ function makePrefix(sourceMapFile) {
         prefix += 'var target = (typeof self === "undefined" || !(typeof WorkerGlobalScope !== "undefined" && self instanceof WorkerGlobalScope)) && ((typeof module !== "undefined" && module.exports) || (typeof window !== "undefined"));';
         prefix += 'if (target) {';
         prefix += 'var ' + source_var + ' = arguments.callee.toString();';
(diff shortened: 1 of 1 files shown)


bcamper:
Sorry for the delay, this will go into the upcoming release.

Issue #646, opened 2018-05-15
sprites failing to draw intermittently


meetar:
[heavily edited as my understanding of the problem evolved]

Since 0.15.0, sprite drawing intermittently fails in a set of Differ tests, I believe because they are very simple scenes containing only a single sprite, which apparently allows `view_complete` to be triggered before the final collision pass finishes drawing.

According to git-bisect the trouble started with https://github.com/tangrams/tangram/commit/882d4827d48be68c4796be13d89609acaea6b077 – although the Differ can't run the tests at all between that commit and https://github.com/tangrams/tangram/commit/6480e1f41622553f542f71c27e856546054172ef, when this issue becomes visible.

**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**

https://tangrams.github.io/differ/?1=https://tangrams.github.io/differ-tests/unit-tests/textures/textures.json&2=&lib1=0.14.1&lib2=0.15.1&go

**RESULT:**

About 5/6 of the time (tested by hitting a single test's "refresh" button 300 times and counting 247 failures), where sprites in the left column are drawn, the sprites in the right column will fail. (Other tests with no visible sprites are for a situation fixed by another branch.)

I have a hacky fix for this issue here: https://github.com/tangrams/tangram/tree/debug-view_complete which runs a check on updateViewComplete() for a specific object and rebuilds rather than trigger `view_complete`.

More specifically: the hack looks for a non-empty object such as `{labels: Array(1), containers: Array(1)}`, returned by mainThreadLabelCollisionPass() and passed through updateLabels().

Another workaround is to comment out the check for `!this.collision.task` at https://github.com/tangrams/tangram/blob/7faa575b2dbe934d5287628878fe5460c1dac3e3/src/tile_manager.js#L159 and run the code inside that block whether there's an existing task or not.

meetar:
I've narrowed down and recreated the issue using a single scene file, to reproduce, load http://tangrams.github.io/tangram-frame/?url=http://tangrams.github.io/differ-tests/unit-tests/textures/sprite-defined-auto.yaml&lib=0.15.1#1/0/0

Then in the console, execute: `scene.load(scene.config_source)`

The point appears about half the time, and in a debugging branch I determined that in those cases the collision task never starts: https://github.com/tangrams/tangram/compare/debug-view_complete

[...]

PR #649, opened 2018-06-05
fix task time tracking

meetar:
@bcamper Suggesting these changes in response to https://github.com/tangrams/tangram/issues/646 – it seems as though there are fewer problems with missing sprites in the Differ with this change (though it doesn't solve all the problems) … Is it possible these variables were set incorrectly?

### DIFF
--- a/src/tile_manager.js
+++ b/src/tile_manager.js
@@ -142,20 +142,26 @@ export default class TileManager {
         // from disappearing, e.g. due to a newly loaded repeat label nearby
         tiles.sort((a, b) => a.build_id < b.build_id ? -1 : (a.build_id > b.build_id ? 1 : 0));
 
-        // check if tile set has changed (in ways that affect collision)
+        // check whether tile set has changed since the task was created (in ways that affect collision)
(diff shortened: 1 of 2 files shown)


bcamper:
Hmm, I don't think those lines are wrong (though it's possible something else with task time tracking is): the `this.` vars track the total task execution time for all tasks this frame. The `task.` ones track time on individual tasks. A new task is only started if the total frame time is under the allowed max.

I suspect some other sync issue with the Differ not knowing the proper frame completion state. Could be that tracking of completion is wrong, or the logic for firing the event isn't right, or how it's being measure/interacted with by Differ is wrong. I guess more examples of how to reproduce and, if possible, a reliable one that always produces the same results would be of great help (I don't think I could ever get that for this  and I gave up and did the release anyway)

meetar:
It looks like the task is just never starting, I narrowed it down here: https://github.com/tangrams/tangram/issues/646#issuecomment-395192407

bcamper:
Root cause resolved in https://github.com/tangrams/tangram/issues/646#issuecomment-441479823 🎉

PR #651, opened 2018-06-19
Quick fix for Polygons with interior (holes)

ThFischer:
**Steps to reproduce the bug:**
Open the [Tangram Play](https://tangram.city/play/?#2/0/0)
and set this YAML definition:
```
sources:
  geojson:
    type: GeoJSON
    url: https://gist.githubusercontent.com/germanz/e0fc172fa458bd50c17bbac9b24e17e0/raw/a1f9bb86265c0ae1c72fc37b99b2967fb302faaa/polygon_with_hole

scene:
  background:
    color: black

layers:
  layer:
    data: {source: geojson}
    draw:
      polygons:
        color: red
        order: 1
```
A red rectangle with a rectangular hole is rendered as expected.
Zoom in until you see render artefacts like this:

screen shot 2018-06-19 at 13 57 11

### DIFF
--- a/src/builders/polygons.js
+++ b/src/builders/polygons.js
@@ -60,9 +60,8 @@ export function buildPolygons (
 
                     // Add UVs:
                     if (texcoord_index) {
-                        vertex_template.splice(texcoord_index, 2,
-                                ((vertex[0] - min_x) * scale_u + min_u) * texcoord_normalize,
-                                ((vertex[1] - min_y) * scale_v + min_v) * texcoord_normalize);
+                        vertex_template[texcoord_index + 0] = ((vertex[0] - min_x) * scale_u + min_u) * texcoord_normalize;
+                        vertex_template[texcoord_index + 1] = ((vertex[1] - min_y) * scale_v + min_v) * texcoord_normalize;
                     }
 
                     vertex_data.addVertex(vertex_template);
--- a/src/builders/polygons.js
+++ b/src/builders/polygons.js
@@ -23,12 +23,13 @@ export function buildPolygons (
     let vertex_elements = vertex_data.vertex_elements,
         element_offset = vertex_data.vertex_count,
         num_polygons = polygons.length,
-        geom_count = 0;
+        geom_count = 0,
+        min_u, min_v, max_u, max_v,
+        min_x, min_y, max_x, max_y;
 
     if (texcoord_index) {
         texcoord_normalize = texcoord_normalize || 1;
-        // Usage of "var" for function wide visibility:
-        var [min_u, min_v, max_u, max_v] = texcoord_scale || default_uvs;
+        [min_u, min_v, max_u, max_v] = texcoord_scale || default_uvs;
     }
 
     for (let p = 0; p < num_polygons; p++) {
@@ -37,13 +38,12 @@ export function buildPolygons (
             indices = triangulatePolygon(earcut.flatten(polygon)),
             num_indices = indices.length;
 
-        // The verices and vertex-elements must not be added if earcut returns no indicies:
+        // The vertices and vertex-elements must not be added if earcut returns no indices:
         if (num_indices) {
 
             // Find polygon extents to calculate UVs, fit them to the axis-aligned bounding box:
             if (texcoord_index) {
-                // Usage of "var" for function wide visibility:
-                var [min_x, min_y, max_x, max_y] = Geo.findBoundingBox(polygon),
+                [min_x, min_y, max_x, max_y] = Geo.findBoundingBox(polygon),
                     span_x = max_x - min_x,
                     span_y = max_y - min_y,
                     scale_u = (max_u - min_u) / span_x,
@@ -69,7 +69,7 @@ export function buildPolygons (
             }
 
             // Add element indices:
-            for (let i = 0; i < num_indices; i++){
+            for (let i = 0; i < num_indices; i++) {
                 vertex_elements.push(element_offset + indices[i]);
             }
             geom_count += num_indices / 3;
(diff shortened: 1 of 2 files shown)


ThFischer:
Be aware, that my other pull request [Fix return value of buildStraightLabel()](https://github.com/tangrams/tangram/pull/650) is needed!

meetar:
Hi @ThFischer, thanks very much for looking into this! I've confirmed the bug and the fix, and successfully tested it with our internal style-testing tool. Will merge both at the end of the week unless @bcamper has any objections before then.

bcamper:
@ThFischer thank you for the investigation here! I'm trying to understand a bit more about exactly what's happening.

Is the issue that [this call to earcut](https://github.com/tangrams/tangram/blob/master/src/builders/polygons.js#L59) is returning no triangles for some cases where part of the polygon becomes degenerate (maybe an interior ring is skipped)? But this call happens after vertices are added [in the loop above](https://github.com/tangrams/tangram/blob/master/src/builders/polygons.js#L54) -- is that right?

If this is the case, could se solve this just by flipping the order of these, so we check if any geometry will be added first ([here](https://github.com/tangrams/tangram/blob/master/src/builders/polygons.js#L58-L64)) and then only proceed to add vertices [here](https://github.com/tangrams/tangram/blob/master/src/builders/polygons.js#L40-L56) if `geom_count` is greater than 0?

I haven't checked if this works, but it could be a slightly simpler solution than having to track the original vertex data positions and reset them (basically avoid the problem in the first place if we can).

Separate question: I merged the fix in #650, which makes sense, but I'm not 100% clear of the relation to this issue, since that deals with simple quads vs. the polygon ring issue here, no? Is it a different manifestation of a similar issue?

Thanks again for these contributions!

ThFischer:
@bcamper  
Hi Brett,
It would be a much better solution to prevent adding of vertices already in case of `triangulatePolygon` returns 0. Thats why I called it a "Quick fix"  ;-)

The dependency of this fix and #650 is caused by the fact, that this fix resets the mesh data if `buildGeometry` doesn't return a value > 0. Exactly this would happen if `buildStraightLabel` doesn't return the expected geom_count (undefined > 0 == false).

bcamper:
Ah thanks I see the link between the two now.

Would you want to try the alternate fix I suggested, or do you prefer we do that?

ThFischer:
I can give it a try in the next days …

ThFischer:
I committed a "proper" fix as suggested by @bcamper.
I also cleaned up the buildPolygons and annotated it.

The absence of annotation and type definitions in Tangram is quite nasty. With annotations you could do signature and type checks during build time (e.g. with Closure Compiler). These would increase the stability (e.g. the #650 issue wouldn't be possible) and readability of the code enormously.

bcamper:
Thanks @ThFischer. I don't disagree at all about the benefit of annotations... though unfortunately there's a fairly large amount of code to make up at this point. Doesn't hurt to start small though! :)

src/builders/polygons.js
```diff
@@ -7,64 +7,80 @@ import earcut from 'earcut';
...
+                // Add vertex data:
+                let polygon_ring = polygon[ring_index];
+                for (let i = 0; i < polygon_ring.length; i++) {
+                    let vertex = polygon_ring[i];
+                    vertex_template[0] = vertex[0];
+                    vertex_template[1] = vertex[1];
+
+                    // Add UVs:
+                    if (texcoord_index) {
+                        vertex_template.splice(texcoord_index, 2,
```
bcamper: A note here, which may or may not be relevant any longer: back when this code was originally written, I did a lot of benchmarks to minimize geometry build time. One of the items that was important at the time was to maintain a fixed-length `vertex_template` array, and only overwrite specific values within it. By using `splice`, you're semantically removing and replacing those elements, rather than just overwriting them. This may not have an actual negative performance effect depending on how the browser handles (and could conceivably be a benefit as one mutation instead of two), but I think it should be benchmarked with a complex scene that uses UV coords to measure the average build time before/after. @meetar is that something you could do?

src/builders/polygons.js
```diff
@@ -7,64 +7,80 @@ import earcut from 'earcut';
...
 
-    var vertex_elements = vertex_data.vertex_elements;
+    let vertex_elements = vertex_data.vertex_elements,
+        element_offset = vertex_data.vertex_count,
+        num_polygons = polygons.length,
+        geom_count = 0;
 
     if (texcoord_index) {
         texcoord_normalize = texcoord_normalize || 1;
+        // Usage of "var" for function wide visibility:
```
bcamper: If I recall, the use of `var` was to get around a limitation in how babel was parsing destructuring, where it was not allowing me to define them once and then reassign with just `[min_u, min_v, max_u, max_v] = texcoord_scale || default_uvs;`. But I think this may be fixed, in which case we could use `let`.

src/builders/polygons.js
meetar: :+1:

ThFischer: You are right, splice is a bad choice because it creates a new array for the return value. I'll change it back to the previous approach.

src/builders/polygons.js
ThFischer: We can't use let here, if the variable is used outside the if block. Of course, as long as babel is used it will work, but not without babel.

bcamper: That makes sense. I was thinking define it above with the others (e.g. [here](https://github.com/tangrams/tangram/pull/651/files/c4652f00818fde764867f23ad369a5ef45eb6eed#diff-052edb1da57e539c22cf5308cd97bf7eR23)). They are only then used if the geometry has UV coords.

ThFischer: Ahh, I just mistaken it, will do …

bcamper:
Squashed and merged, assigned to v0.15.2 for release.

PR #654, opened 2018-07-02
try/catch on evalProperty

meetar:


bcamper:
I expanded the scope of this to include other cases where properties can be JS functions, and also included some additional error context (only "[Object object]" was being output before anyway).

nvkelso:
Rad! Once this is merged could we cut a v0.15.2 release?

bcamper:
Ya. Will do.

Issue #668, opened 2018-08-30
Allow Bearer authentication 


pakore:
Allow OAuth2 Bearer authentication ( http://self-issued.info/docs/draft-ietf-oauth-v2-bearer.html ) for the backend.

Or at least, allow sending custom headers to the backend.

Use Case:
I am a customer and I have a backend that provides tiles but it needs an OAuth2 token in the header ( "Authorization: Bearer <token>" header). I would like to use tangram to render the tiles, so I need a way to give tangram my Oauth2 credentials or at least be able to tell tangram to pass this custom header with a provided token.

bcamper:
Yep, I remember I looked at this a couple times. If I recall, one issue is that for CORS requests, custom headers need to be whitelisted by the server with [`Access-Control-Allow-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers). And some more info/example [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests).

The Netzen tiles do not directly support this concept (though I suppose they could, and could mimic back the list of headers requested by the client, for example). But if the client is serving their own tiles (or maybe proxying another source), I believe it would work if configured properly. Worth trying a prototype.

pakore:
Allowing sending custom headers as @matteblair  proposes would be the more generic solution, that will solve Auth problems, CORS problems, etc...

For the usecase: the user could manage getting a proper token and refresh it in custom code meanwhile.

bcamper:
@pakore re: my comments about `Access-Control-Allow-Headers`, that makes sense and fits into your workflow?

pakore:
I currently don't have any problems with CORS because my server accepts all, but I think definitely is a use case that needs to be solved for other users also and allowing custom headers could be a solution.

bcamper:
See #670 for an implementation of this.

bcamper:
Released in v0.16.0

PR #670, opened 2018-09-16
Data sources can specify optional request headers

bcamper:
Following up on #668: to enable cases where a client would like to send additional, custom HTTP headers with each tile request.

A data source can specify key/value pairs for header names/values, with the `request_headers` parameter. For example, if the tile backend is authorized with and OAuth token:

```
sources:
    tiles:
        type: TopoJSON
        url: ...
        request_headers:
            Authorization: Bearer XXXXXX
```

While `request_headers` *can* be specified statically in the scene file, it's more likely they would be added programmatically at run-time (they could either be added after a scene `load` event, or by directly modifying `scene.config` and calling `scene.updateConfig()`). They would also generally be expected to be used with custom tile sources.

CORS
------
If the tile source is being made via CORS requests (the tile source's origin is different from the origin the the web page was loaded from), then adding custom `request_headers` will cause an additional 
["preflight"](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) `OPTIONS` request for each tile (before the `GET` request for tile data). This extra request is required so that the client can verify that the server will accept the desired headers, which the server must authorize with the `Access-Control-Allow-Headers` header. See more info on [CORS preflight requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests).

### DIFF
--- a/demos/scene.yaml
+++ b/demos/scene.yaml
@@ -97,6 +97,8 @@ sources:
         url: https://tile.nextzen.org/tilezen/vector/v1/512/all/{z}/{x}/{y}.topojson
         tile_size: 512
         max_zoom: 15
+        # request_headers: # send custom headers with tile requests
+        #     Authorization: Bearer xxxxxxxxx
         # Data filtering demo with 'scripts' and 'transform' properties:
         # Tile data is passed through a 'transform' pre-processing function before Tangram geometry is built.
         # 'transform' adds an 'intersects_park' property to each road that intersects a park feature.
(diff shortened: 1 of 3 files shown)


pakore:
This is awesome! Thanks a lot.

bcamper:
@matteblair @tallytalwar are you good with this from a future ES perspective? (Should follow the syntax suggested by @matteblair.)

matteblair:
Yeah I don't see any problems implementing this in Tangram ES :+1:

Issue #672, opened 2018-09-18
`text_source` array fails for points labels


meetar:
**TANGRAM VERSION:**

Tested from v0.12.0 on – works as expected in v0.11.

**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**

Visit https://tangram.city/play/?#10.4708/37.7991/-122.5005 using this scene file:

```
sources:
    nextzen:
        type: TopoJSON
        url: https://tile.nextzen.org/tilezen/vector/v1/all/{z}/{x}/{y}.topojson

layers:
    water:
        data: { source: nextzen }
        draw:
            polygons:
                order: 1
                color: lightblue
            points:
                visible: false
                order: 2
                size: 10px
                color: red
                text:
                    text_source: [name, kind]
                    font:
                        size: 10px
                        fill: white
            text:
                text_source: [name, kind]
                font:
                    size: 10px
                    fill: white
```

…this shows `text_source: [name, kind]` working as expected for a `text` style. Then change `visible: false` to `visible:true`.

**RESULT:**

No text labels are drawn for the points style, and the console throws:

> Layer 'water': cannot use boundary labels (e.g. 'text_source: { left: ..., right: ... }') for 'text' labels attached to 'points'; provided 'text_source' value was ["name","kind"]

**EXPECTED RESULT:**

I expect the `points` style to use the array in `text_source` as a list of prioritized feature properties to use for the text label.

PR #673, opened 2018-09-18
handle multi-source labels

meetar:
Handles cases where the `text_source` for a point label should be the first available property in an array of prioritized feature properties, as in `[name, kind]`.

Fixes https://github.com/tangrams/tangram/issues/672.

### DIFF
--- a/src/styles/points/points.js
+++ b/src/styles/points/points.js
@@ -236,11 +236,9 @@ Object.assign(Points, {
             draw.text.visible !== false && // explicitly handle `visible` property for nested text
             this.parseTextFeature(feature, draw.text, context, tile);
 
+        // if the text feature is an array, there were multiple options: use the first entry
         if (Array.isArray(tf)) {
-            tf = null; // NB: boundary labels not supported for point label attachments, should log warning
-            log({ level: 'warn', once: true }, `Layer '${draw.layers[draw.layers.length-1]}': ` +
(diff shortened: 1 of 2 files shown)


bcamper:
I have not tested this but I believe the root cause is that arrays are improperly falling through this line:

[`if (source != null && typeof source === 'object') `](https://github.com/tangrams/tangram/blob/0fea0c09f38e23cfecd58c311bee75b21aea3581/src/styles/text/text_labels.js#L107
)

`parseTextSource()` should already properly account for "multi-source" labels, and return only a single value for them (it will still return an object for left/right boundary labels).

So for example I don't think we should need any changes [here](https://github.com/tangrams/tangram/pull/673/files#diff-495004c1ba85e44fb8e478cac90b13d6R48), [here](https://github.com/tangrams/tangram/pull/673/files#diff-5865521f4fdb3091818c71745ae6a4edR241) (it's a logic error for an array to ever get to this point), or otherwise downstream.

bcamper:
Sorry, meant to put the fix in that comment too. I think that all that is needed to fix the bug is that

`if (source != null && typeof source === 'object')`

should be

`if (source != null && !Array.isArray(source) && typeof source === 'object')`

Again this is a guess so there might be a nuance I'm missing but that looks like an issue with the current code :)

meetar:
Confirmed :)

bcamper:
Thanks, I'll squash it on merge to avoid having the extra changes added and then removed.

Issue #354, opened 2016-07-12
Debug flag to visualize tile boundaries


bcamper:
It would be useful to be able to visualize tile boundaries consistently. We often use the `tile_edges: true` flag for this when drawing land or water layers, but this both requires polygon geometry to be present in the source data covering the tile boundaries, and for additional `draw` blocks to add `lines`. We've also looked at highlighting Leaflet `<div>`s, but they do not follow Tangram's zoom conventions and so they are not an accurate depiction of the currently visible Tangram tiles.

This debug should just be built into the library. The most straightforward way to do this is probably to generate a simple GL quad geometry for each tile using `GL_LINES`, and include it in the tile rendering process. I believe this is similar to the technique used in ES - @karimnaaji? Note that we aren't actually using `GL_LINES` anywhere else in our render process, though the `VBOMesh` class was written with this use case in mind; it' possible some clean-up is needed to get it up to speed. Another option would be to actually tesselate polylines for the tile boundaries, like we do for other lines.

Because tiles are rendered with meshes grouped *by style* rather than all in one render call, we may need to add an additional call for this type of tile-wide drawing/debugging, somewhere in the main [tile draw loop](https://github.com/tangrams/tangram/blob/master/src/scene.js#L518).

nvkelso:
+1

karimnaaji:
We have two ways to draw debug on -es:

1. Screen space polygons called [primitives](https://github.com/tangrams/tangram-es/blob/master/core/src/gl/primitives.cpp) using immediate rendering mode available with `glVertexAttribPointer` available on Opengl ES. It uses client side buffer data instead of a buffered vbo, it's faster to implement (no need of setting up vbos/vaos) but much slower on rendering, so it's a good fit for debug. I doubt that WebGL has something similar since client data is more protected but would be worth looking at. Otherwise everything is drawn using `GL_LINES` like mentioned. Those are not rendered in the tile drawing process and are used for labels, debug stats.
2. Regular meshes drawn with `GL_LINES` for tile boundaries where each geometry is the 4 corners of a tile. Those are drawn in the tile drawing process.

bcamper:
Ah, yeah there's no "client data" in WebGL. Making regular meshes with `GL_LINES` is probably going to be the simplest (will use existing code and fit into the rendering pipeline cleanly).

bcamper:
Here are some notes on a couple possible paths:

**1. Hook into the existing tile building pipeline**
[...]

PR #675, opened 2018-09-19
make tile available in source transform function

meetar:
This makes per-tile data transformations possible, including debugging data. For example, this creates a polygon at the tile boundaries, named for its source tile:

```javascript
transform: |
    function(data) {
        t = this.tile;
        c = Tangram.debug.Geo.metersToLatLng;
        min = c([t.min.x, t.min.y]);
        max = c([t.max.x, t.max.y]);
        return {
            type: 'FeatureCollection',
            features: [{
                type: 'Feature',
                geometry: {
                    type: 'Polygon',
                    coordinates: [[
                        [min[0], min[1]], [max[0], min[1]],
                        [max[0], max[1]], [min[0], max[1]],
                        [min[0], min[1]]
                    ]]
                },
                properties: {
                    name: this.tile.coords.key
                }
            }]
        };
    }
```

Addresses https://github.com/tangrams/tangram/issues/354.

### DIFF
--- a/src/sources/geojson.js
+++ b/src/sources/geojson.js
@@ -114,10 +114,10 @@ export class GeoJSONSource extends NetworkSource {
     parseSourceData (tile, source, response) {
         let data = typeof response === 'string' ? JSON.parse(response) : response;
         let layers = this.getLayers(data);
-        source.layers = this.preprocessLayers(layers);
+        source.layers = this.preprocessLayers(layers, tile);
     }
 
-    preprocessLayers (layers){
+    preprocessLayers (layers, tile){
         for (let key in layers) {
             let layer = layers[key];
             layer.features = this.preprocessFeatures(layer.features);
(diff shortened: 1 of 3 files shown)


nvkelso:
> name: this.tile.coords.key

👍

src/sources/geojson.js
```diff
@@ -235,6 +235,7 @@ export class GeoJSONTileSource extends NetworkTileSource {
     prepareGeoJSON (data, tile, source) {
         // Apply optional data transform
         if (typeof this.transform === 'function') {
+            this.tile = tile;
```
bcamper: We should avoiding modifying `this` with transient data. But we also should avoid leaking `this` into these functions in the first place (an existing problem I hadn't realized!). It could open up some pretty crazy behavior, e.g. if modifications are then made in the function. I don't think we have this problem except in data source `transform` and `preprocess` functions; I can take care of these and will look for other possible cases.

For this case: to be more consistent with the existing syntax, I suggest making the `tile` an additional argument to the `transform` function:
`data = this.transform(data, this.extra_data, tile);`

If the calling code wants to access `tile` but doesn't have any `extra_data`, that's fine, it will just be an undefined value:

```
transform: |
    function(data, extra_data, tile) {
        t = tile;
        ...
```

bcamper:
@meetar did you want to follow up on the (relatively small) updates described above?

PR #685, opened 2018-12-06
Meetar/overzoom

meetar:
This PR adds an `overzoom` parameter to data source specifications, to allow manual adjustment of the internal `zoom_bias` property.

In circumstances when the default tiles are either too large or detailed for the current application, this parameter allows the user to request lower zooms for each zoom level, reducing tile size and visual detail.

Examples:

```yaml
sources:
  tilezen:
    type: TopoJSON
    url: https://tile.nextzen.org/tilezen/vector/v1/512/all/{z}/{x}/{y}.topojson
    tile_size: 512
    overzoom: 3
    url_params:
        api_key: global.api_key
```

```yaml
sources:
  stamen:
    type: Raster
    overzoom: 2
```

### DIFF
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -48,6 +48,11 @@ export default class DataSource {
         // set a custom extra overzoom adjustment factor to load consistently lower zoom levels
         // than the current map zoom level – eg a zoom_offset of 1 would load z3 data at z4
         this.zoom_offset = (config.zoom_offset != null) ? config.zoom_offset : 0;
+        if (this.zoom_offset < 0) {
+            let msg = `Data source '${this.name}' zoom_offset must not be negative – setting to 0. `;
+            log({ level: 'warn', once: true }, msg);
+            this.zoom_offset = 0;
+        }
 
         this.setTileSize(config.tile_size);
         // factor in zoom_bias to account for tile sizes > than 256px
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -45,6 +45,9 @@ export default class DataSource {
         // overzoom will apply for zooms higher than this
         this.max_zoom = (config.max_zoom != null) ? config.max_zoom : Geo.default_source_max_zoom;
 
+        // set a custom extra overzoom factor to load consistently lower zoom levels
+        this.overzoom = (config.overzoom != null) ? config.overzoom : 0;
+
         this.setTileSize(config.tile_size);
         this.max_coord_zoom = this.max_zoom + this.zoom_bias;
 
@@ -159,7 +162,10 @@ export default class DataSource {
         }
 
         // # of zoom levels bigger than 256px tiles - 8 in place of log2(256)
-        this.zoom_bias = Math.log2(this.tile_size) - 8;
+        // Many Tangram functions assume 256px tiles, this factor adjusts for the
+        // case of bigger tile sizes - eg 512px tiles are 1 zoom level bigger,
+        // 1024px tiles are 2 levels bigger
+        this.zoom_bias = Math.log2(this.tile_size) - 8 + this.overzoom;
     }
 
     // Infer winding for data source from first ring of provided geometry
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -45,8 +45,9 @@ export default class DataSource {
         // overzoom will apply for zooms higher than this
         this.max_zoom = (config.max_zoom != null) ? config.max_zoom : Geo.default_source_max_zoom;
 
-        // set a custom extra overzoom factor to load consistently lower zoom levels
-        this.overzoom = (config.overzoom != null) ? config.overzoom : 0;
+        // set a custom extra overzoom adjustment factor to load consistently lower zoom levels
+        // than the current map zoom level – eg a zoom_offset of 1 would load z3 data at z4
+        this.zoom_offset = (config.zoom_offset != null) ? config.zoom_offset : 0;
 
         this.setTileSize(config.tile_size);
         // factor in zoom_bias to account for tile sizes > than 256px
@@ -166,7 +167,7 @@ export default class DataSource {
         // Many Tangram functions assume 256px tiles, this factor adjusts for the
         // case of bigger tile sizes - eg 512px tiles are 1 zoom level bigger,
         // 1024px tiles are 2 levels bigger
-        this.zoom_bias = Math.log2(this.tile_size) - 8 + this.overzoom;
+        this.zoom_bias = Math.log2(this.tile_size) - 8 + this.tile_adjust;
     }
 
     // Infer winding for data source from first ring of provided geometry
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -167,7 +167,7 @@ export default class DataSource {
         // Many Tangram functions assume 256px tiles, this factor adjusts for the
         // case of bigger tile sizes - eg 512px tiles are 1 zoom level bigger,
         // 1024px tiles are 2 levels bigger
-        this.zoom_bias = Math.log2(this.tile_size) - 8 + this.tile_adjust;
+        this.zoom_bias = Math.log2(this.tile_size) - 8 + this.zoom_offset;
     }
 
     // Infer winding for data source from first ring of provided geometry
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -49,14 +49,12 @@ export default class DataSource {
         // than the current map zoom level – eg a zoom_offset of 1 would load z3 data at z4
         this.zoom_offset = (config.zoom_offset != null) ? config.zoom_offset : 0;
         if (this.zoom_offset < 0) {
-            let msg = `Data source '${this.name}' zoom_offset must not be negative – setting to 0. `;
+            let msg = `Data source '${this.name}' zoom_offset must not be negative – setting to 0.`;
             log({ level: 'warn', once: true }, msg);
             this.zoom_offset = 0;
         }
 
         this.setTileSize(config.tile_size);
-        // factor in zoom_bias to account for tile sizes > than 256px
-        this.max_coord_zoom = this.max_zoom + this.zoom_bias;
 
         // no tiles will be requested or displayed outside of these min/max values
         this.min_display_zoom = (config.min_display_zoom != null) ? config.min_display_zoom : 0;
@@ -173,6 +171,12 @@ export default class DataSource {
         // case of bigger tile sizes - eg 512px tiles are 1 zoom level bigger,
         // 1024px tiles are 2 levels bigger
         this.zoom_bias = Math.log2(this.tile_size) - 8 + this.zoom_offset;
+
+        // the max/min coordinate zoom level at which tiles will be loaded from server
+        // (but tiles can be styled at other zooms)
+        // zoom_bias adjusts for tile sizes > than 256px, and manual zoom_offset
+        this.max_coord_zoom = this.max_zoom + this.zoom_bias;
+        this.min_coord_zoom = this.zoom_bias;
     }
 
     // Infer winding for data source from first ring of provided geometry
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -49,6 +49,7 @@ export default class DataSource {
         this.overzoom = (config.overzoom != null) ? config.overzoom : 0;
 
         this.setTileSize(config.tile_size);
+        // factor in zoom_bias to account for tile sizes > than 256px
         this.max_coord_zoom = this.max_zoom + this.zoom_bias;
 
         // no tiles will be requested or displayed outside of these min/max values
--- a/src/sources/raster.js
+++ b/src/sources/raster.js
@@ -1,5 +1,5 @@
 import DataSource, {NetworkTileSource} from './data_source';
-import Tile from '../tile';
+import {TileID} from '../tile_id';
 import Geo from '../geo';
 import log from '../utils/log';
 
@@ -70,7 +70,7 @@ export class RasterTileSource extends NetworkTileSource {
             let zdiff = this.zoom_bias - tile_source.zoom_bias; // difference in zoom detail between the sources
             if (zdiff > 0) { // raster source is less detailed
                 // do extra zoom adjustment and apply this raster source's max zoom
-                coords = Tile.normalizedCoordinate(tile.coords, {
+                coords = TileID.normalizedCoord(tile.coords, {
                     zoom_bias: zdiff,
                     max_zoom: this.max_zoom
                 });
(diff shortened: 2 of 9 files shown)


nvkelso:
Neat!

We already use `max_zoom` in the sources – is `overzoom` or `over_zoom` more consistent? Or should we use another term like `zoom_offset` or?

What would the new documentation read like for overzoom? I think of it more as a zoom_offset where I'm at zoom X but I'm going to request zoom X-y tiles to display at X zoom (as overzoomed). Is that how you're thinking about it?

Here's the reference for existing `max_zoom`:

- https://tangrams.readthedocs.io/en/latest/Syntax-Reference/sources/

```
max_zoom
Optional integer. Default is 18.

Sets the highest zoom level which will be requested from the datasource. At higher zoom levels, the data from this zoom level will continue to be displayed, a condition called "overzoom".

There is no corresponding min_zoom parameter, for reasons of performance.
```

tallytalwar:
Is this the same zoom offset we were discussing @nvkelso? If so then yes, I prefer `zoom_offset`.

meetar:
Yep, agreed re: `zoom_offset` 👍

meetar:
If the map is zoomed out such that the current zoom - `zoom_offset` < the source's `min_zoom`, the `min_zoom` tiles are reloaded and rebuilt every time the zoom level changes – I'll add a check to prevent this.

src/sources/raster.js
```diff
@@ -47,9 +47,14 @@ export class RasterTileSource extends NetworkTileSource {
 
     // Return texture info for a raster tile
     tileTexture (tile) {
-        let key = tile.coords.key;
+        let new_coords = tile.coords;
+        if (this.zoom_offset !== 0) {
```
bcamper: I think you can achieve this section (the calls to `coordinateAtZoom` and `coordinateWithMaxZoom`) with one call to `normalizedCoordinate`:

https://github.com/tangrams/tangram/blob/master/src/tile.js#L86

e.g.

`let coords = Tile.normalizedCoordinate(tile.coords, { max_zoom: this.max_zoom, zoom_bias: this.zoom_offset });`

(Side note, it looks like the third parameter for `normalizedCoordinate`, `style_zoom`, is vestigial/unused.)

bcamper: I'm still a bit confused thinking through the use of `zoom_bias` vs. `zoom_offset` here (sources can now have both). I think it's because we determined that attached raster sources ignore their `tile_size` and are required to use the `tile_size` of their parent vector source. This could be revisited. I think the true restriction is that attached rasters need a `tile_size` that is the **same or larger** than their parent, but **not smaller**. This is because of the way we sample rasters within vector parent source shaders, where the raster is mapped to a single GL texture sampler. The vector source can cover **all or part of** a single raster texture (in the case it's partially covered, UVs are automatically adjusted to cover the appropriate sub-rectangle for each vector tile), but cannot span multiple raster textures.

So:
- We could relax this so rasters allow an equal or larger `tile_size` to the parent vector (parent vector source throws an error when trying to attach a raster source with a smaller `tile_size`)
- Docs reflect this (I don't think we mention anything about the "locking" of `tile_size` for attached rasters right now.

These are notes for related revisions, but don't need to be part of this PR.

src/sources/data_source.js
```diff
@@ -45,7 +45,11 @@ export default class DataSource {
         // overzoom will apply for zooms higher than this
         this.max_zoom = (config.max_zoom != null) ? config.max_zoom : Geo.default_source_max_zoom;
 
+        // set a custom extra overzoom factor to load consistently lower zoom levels
+        this.overzoom = (config.overzoom != null) ? config.overzoom : 0;
```
bcamper: `overzoom` needs to be re-named.

src/sources/data_source.js
```diff
@@ -159,7 +163,10 @@ export default class DataSource {
         }
 
         // # of zoom levels bigger than 256px tiles - 8 in place of log2(256)
-        this.zoom_bias = Math.log2(this.tile_size) - 8;
+        // Many Tangram functions assume 256px tiles, this factor adjusts for the
+        // case of bigger tile sizes - eg 512px tiles are 1 zoom level bigger,
+        // 1024px tiles are 2 levels bigger
+        this.zoom_bias = Math.log2(this.tile_size) - 8 + this.overzoom;
```
bcamper: `overzoom` needs to be re-named.

src/sources/raster.js
meetar: I believe we're agreed here. As I understand it, `zoom_bias` is strictly internal, and tracks cases of overzoom past `max_zoom` limits. `zoom_offset` would be an additional external control, in cases where the user wants to load 1 or more zoom levels below the current map zoom level. `zoom_offset` would be applied if its result is a lower zoom than the result of applying`zoom_bias`.

Examples, assuming a map zoom of 11:

- `max_zoom` is 10 – `zoom_bias` is 1, requested zoom is 10.
- `max_zoom` is 10, `zoom_adjust` is 1 – `zoom_bias` and `zoom_adjust` are equal, requested zoom is 10.
- `max_zoom` is 11, `zoom_adjust` is 1 – `zoom_bias` of 0 is less than `zoom_adjust`, requested zoom is 10.
- `max_zoom` is 10, `zoom_adjust` is 2 – `zoom_bias` of 1 is less than `zoom_adjust`, requested zoom is 9.

meetar: … this is slightly complicated by the fact that `zoom_bias` is also adjusted in cases where a datasource uses a non-256px tile, but I believe even then these manipulations should still work. https://github.com/tangrams/tangram/blob/e3cd42dad5a8e6416ceca69a10a1908bf46ad852/src/sources/data_source.js#L170

bcamper: I think there's still some confusion here.

`zoom_bias` arises from `tile_size` and the introduction of 512px tiles, and does something very similar to `zoom_offset` -- causes logically sized 256px tiles to load one or more levels "up". When we added 512px support, we opted to have an explicit pixel size param like `tile_size: 512` instead of making the user think in terms of zoom levels. But, `zoom_bias` isn't directly related to overzooming past `max_zoom`, though overzooming code does need to take it into account.

In fact, `zoom_offset` is incorporated into `zoom_bias` already (though in this current branch seems to have been mistakenly (?) referred to as `tile_adjust` [here](https://github.com/tangrams/tangram/pull/685/files#diff-06e3f25a9a00a1a8840d2720f9ab37e1R170)... and is also referred to as `zoom_adjust` in your comments above?). It's basically an extension of `zoom_bias`, and mostly doesn't need to be used again once the data source is created.

But we ran into a specific divergence from that here with attached rasters, because of their unique relationship to their parent vector source...

**For normal vector source and standalone raster source (not attached to a vector source) tile resolution**: `zoom_bias` incorporates `zoom_offset` already, and I think everything works as desired.

**For raster sources attached to a vector source**: `zoom_bias` of the raster source is (intentionally, but undocumented as far as I see) not being applied, because we have been forcing the `tile_size` (from which `zoom_bias` is derived) to inherit from the vector source (which is passing in the tile coords here, which will already have been adjusted for the vector source's `zoom_bias`). The additional code here then applies just the `zoom_offset` (instead of the whole `zoom_bias`), which otherwise would be skipped.

Does that all make sense? :)

My original comment was just that we can reuse existing code by using the existing `Tile.normalizedCoordinate` function, which works very similarly to the code in this block.

src/sources/data_source.js
```diff
@@ -159,7 +164,10 @@ export default class DataSource {
         }
 
         // # of zoom levels bigger than 256px tiles - 8 in place of log2(256)
-        this.zoom_bias = Math.log2(this.tile_size) - 8;
+        // Many Tangram functions assume 256px tiles, this factor adjusts for the
+        // case of bigger tile sizes - eg 512px tiles are 1 zoom level bigger,
+        // 1024px tiles are 2 levels bigger
+        this.zoom_bias = Math.log2(this.tile_size) - 8 + this.tile_adjust;
```
bcamper: `tile_adjust` should be `zoom_offset`?

meetar: yeah I just noticed that too! not sure how that got in there…

src/sources/raster.js
meetar: Yep, thanks for catching that. I've addressed those misnomers.

Re: attached rasters: yes, I'm seeing this as well – in my experiments, the practical effect is that an attached raster effectively inherits the `zoom_offset` of its 'parent' vector datasource. Given the technical limitations you described, would it be possible  to allow an attached raster to load at a higher zoom level than its parent?

I'll take a look at that `normalizedCoordinate` function after I unwrap my head from around this wrinkle :)

bcamper:
@meetar @nvkelso 
This is an interesting feature for reducing feature/raster resolution. If the primary use case is reducing network bandwidth for terrain shading, here are some interesting findings that suggest we could also just down-sample to use 256px normal tiles (while keeping the tile size 512px, meaning it is loading tiles at half-pixel resolution), which does not need any additional functionality.

Some Walkabout numbers, 1300x650 viewport:

**Pacific NW at 6.666666666666669/46.608586122368365/-117.93289320122385**
- Baseline (existing Walkabout):
  - `tile_size: 512`, url `/tilezen/terrain/v1/512/normal/`, `zoom_offset: 0`
  - 6 tiles / 1.6mb
- Zoom offset:
  - `tile_size: 512`, url `/tilezen/terrain/v1/512/normal/`, `zoom_offset: 1`
  - **2 tiles** / 488k
- 256px tiles scaled up to 512px:
  - `tile_size: 512`, url `/tilezen/terrain/v1/256/normal/`, `zoom_offset: 0`
  - 6 tiles / **407k**

**Washington state at 9.166666666666666/48.545903372270175/-120.71586691652305**
- Baseline (existing Walkabout):
  - `tile_size: 512`, url `/tilezen/terrain/v1/512/normal/`, `zoom_offset: 0`
  - 8 tiles / 3.5mb
- Zoom offset:
  - `tile_size: 512`, url `/tilezen/terrain/v1/512/normal/`, `zoom_offset: 1`
  - **4 tiles** / 2.0mb
- 256px tiles scaled up to 512px:
  - `tile_size: 512`, url `/tilezen/terrain/v1/256/normal/`, `zoom_offset: 0`
  - 8 tiles / **1.0MB**

The second two options are pixel identical. They load the same normal data but in two different ways. The 256px option doesn't save any tile requests, but loads less (in some cases considerably less) data overall, because it's keeping closer to the size of the viewport.

That technique could be extended by creating smaller tiles, e.g. 128px or 64px normal tiles, and loading those to cover the 512px tile space (the actual texture loaded can be any size, it just gets scaled to fit the expected raster tile size). This is more intensive compared to just changing a parameter of course, but also can achieve impressive results with current functionality.

It's also not mutually exclusive with adding this feature though, which is more flexible, if we want that control of have other use cases in mind (down-sampling vector data).

nvkelso:
Clever to mix and match the 512 and 256 raster tile endpoints in this way!

I think it lessens the need to add support for `zoom_offset`, but `zoom_offset` is still the more generic way of solving the problem (as not all raster sources will support this server side trick).

bcamper:
Not really. I may rename a couple related things to reflect the new usage of `Tile. normalizedCoordinate()` here. Otherwise I will tag this for a 0.17 release. My work on untiled rasters is the other current candidate for that.

src/sources/data_source.js
```diff
@@ -45,7 +45,12 @@ export default class DataSource {
         // overzoom will apply for zooms higher than this
         this.max_zoom = (config.max_zoom != null) ? config.max_zoom : Geo.default_source_max_zoom;
 
+        // set a custom extra overzoom adjustment factor to load consistently lower zoom levels
+        // than the current map zoom level – eg a zoom_offset of 1 would load z3 data at z4
+        this.zoom_offset = (config.zoom_offset != null) ? config.zoom_offset : 0;
```
bcamper: We should enforce this is 0 or positive (no negative values), and log a warning with the `once` option ([example](https://github.com/tangrams/tangram/blob/548ebf28916fa0e77b67bb6204c555e6300bd8ca/src/styles/style.js#L272))

src/tile.js
```diff
@@ -80,7 +80,7 @@ export default class Tile {
     }
 
     static normalizedKey (coords, source, style_zoom) {
-        return Tile.key(Tile.normalizedCoordinate(coords, source, style_zoom), source);
+        return Tile.key(Tile.normalizedCoordinate(coords, source), source, style_zoom);
```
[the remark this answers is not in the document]
meetar: Is this resolved or does it still need reverting?

meetar:
Still seeing unnecessary tile unloading and rebuilding when the current zoom is less than the `zoom_offset` – I assume this has something to do with the overzoom capability not recognizing the currently-displayed tile as the requested zoom.

meetar:
```yaml
sources:
    stamen-terrain:
        type: Raster
    nextzen-osm:
        type: TopoJSON
        url: https://tile.nextzen.org/tilezen/vector/v1/256/all/{z}/{x}/{y}.topojson
        rasters: [stamen-terrain] # attach stamen terrain
        zoom_offset: 3

layers:
    terrain:
        data: { source: nextzen-osm, layer: earth } # render earth layer from vector data source
        draw:
            raster:
                order: 0 # draw on bottom
```
Result:

offset-proxy

bcamper:
Yep, good catch. Looks like this is actually an existing bug which can be seen at z0 when using 512px tiles (since that's the same as a `zoom_offset` of 1 with 256px tiles). But the higher zoom offsets are making it more obvious. I know the area that needs to be adjusted in the proxy finding logic (in `TilePyramid`). It won't look past the z0 coordinate tile, but in these cases, there are multiple tiles at the z0 coord level (one per style zoom, e.g. a z0 data tile styled for display at zoom 1). There is similar logic on the other end to handle style zooms past `max_zoom`. I'll take a look.

bcamper:
This has ended up being significantly more complicated, to ensure that the relative zoom "downsampling" (effected by both `tile_size` and `zoom_offset` parameters) is applied correctly when a raster source is attached to another source. (The prior solution was downsampling the raster `zoom_offset` 2x, because it had already been applied once upstream when the geometry was created... but when the raster is attached to another source, it is the attaching source's zoom params that are applied.) This should be addressed here, but it's a fair amount of new logic so should be tested carefully (and increases the weight of this feature, though also improves logging/warning for existing cases where raster sources would be silently downsampled to match a vector source, e.g. if they had different `tile_size` values): https://github.com/tangrams/tangram/pull/685/commits/d0b352d8a6c0dab425afdd6b5914298da219624e

bcamper:
I should clarify , my last comment is unrelated to the proxy tile bug mentioned in https://github.com/tangrams/tangram/pull/685#issuecomment-446839813, that was resolved in https://github.com/tangrams/tangram/pull/685/commits/b8a77e90b415969c22115d0563b1ef3ad76c4dbe.

But I discovered another bug with raster `zoom_offset` being applied twice when addressing that issue.

PR #686, opened 2018-12-24
Fix repeat culling on main thread for non-collision labels

bcamper:
This fixes repeat behavior across tiles for labels that don't have regular collision applied (e.g. `collide: false`). These labels had repeat culling applied *within* the tile, during the initial tile build on the worker thread, but did not have an additional repeat culling pass applied later on the main thread, to catch repeats from nearby tiles (labels with `collide: true` *did* already have this additional check; `collide: false` labels skip part of the collision pipeline for speed, and were inadvertently missing this piece).

This is solved with an additional flag that tracks whether the label is less than its repeat distance away from the tile edge. If so, it may need to be culled with labels from nearby tiles; if not, it's far enough away to not be affected by them.

Here's an example showing labels with a repeat distance of `50px`. Note the extra labels that are too close one another, which are now removed. For dense point sets, this could also cause an artifact that showed a "ghost" tile edge, where points were noticeable denser.

webp net-gifmaker

This branch also removes the deprecated `moveIntoTile()` function.

### DIFF
--- a/src/labels/collision.js
+++ b/src/labels/collision.js
@@ -120,6 +120,12 @@ export default Collision = {
                                 object.linked.label.breach = true;
                             }
 
+                            // Similarly for labels that need main thread repeat culling, keep linked labels in sync
+                            if (object.label.may_repeat_across_tiles || object.linked.label.may_repeat_across_tiles) {
+                                object.label.may_repeat_across_tiles = true;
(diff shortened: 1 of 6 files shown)


bcamper:
@burritojustice this is relevant to some of your dense point data overlays.

nvkelso:
Neat!

PR #687, opened 2018-12-25
Fix label default repeat groups

bcamper:
This fixes an issue with label repeat group logic. The intended behavior is for labels to apply repeat rules **within their unique scene layer group**. For example, labels for a major and minor road with the same name *will* be allowed to repeat near each other. This was intended to catch uncommon but still existent cases where labels of different feature classes are desirable to show.

However, now that the bug is fixed, I notice that there are some suboptimal cases in Mapzen styles where extra labels are being shown nearby to each other and a bit crowded -- often 2 but sometimes even 3 -- because they are different road types, including ramps, etc. See the extra Market St and Park Row labels below:

sf16
sf15
nyc15

This can be addressed in the style by putting all of the road labels into one `repeat_group`, or at least into fewer, to consolidate them. That's what I've done in the default demo [here as an example](https://github.com/tangrams/tangram/compare/line-label-repeats?expand=1#diff-7f549883b996e161107e531e5a5db1cb).

This leads me to wonder if we should consider changing the default instead. A couple proposals:

- Use a global namespace for `repeat_group` by default (this is what the bug was doing!)
- Use only the top-level scene layer name instead (e.g. `roads`, instead of getting as granular as each sub-layer). This would still allow cases like roads and parks with the same name to repeat.

In either case, the default could still be overridden with a custom `repeat_group`.

@nvkelso please review the above and give your thoughts on restoring the original behavior vs. using a new default.

### DIFF
--- a/src/styles/points/points.js
+++ b/src/styles/points/points.js
@@ -419,9 +419,6 @@ Object.assign(Points, {
 
         // Repeat rules - no repeat limitation for points by default
         draw.repeat_distance = StyleParser.createPropertyCache(draw.repeat_distance, StyleParser.parseNumber);
-        if (draw.repeat_group == null) {
-            draw.repeat_group = draw.layers.join('-');
-        }
 
         // Placement strategies
         draw.placement = PLACEMENT[draw.placement && draw.placement.toUpperCase()];
@@ -494,7 +491,9 @@ Object.assign(Points, {
                 layout.repeat_group = draw.repeat_group(context); // dynamic repeat group
             }
             else {
-                layout.repeat_group = draw.repeat_group; // pre-computed repeat group
+                // default to top-level layer name
+                // (e.g. all labels under `roads` layer, including sub-layers, are in one repeat group)
+                layout.repeat_group = draw.repeat_group || context.layer;
             }
         }
 
--- a/src/styles/text/text_labels.js
+++ b/src/styles/text/text_labels.js
@@ -304,6 +304,12 @@ export const TextLabels = {
             StyleParser.parsePositiveNumber
         );
 
+        // Repeat group - if not defined, use layer names
+        // (but only if available - will be null for text attached to a point, and will inherit point's repeat group)
+        if (draw.repeat_group == null && draw.layers != null) {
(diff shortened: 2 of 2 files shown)


nvkelso:
Generally speaking, this makes the most sense to me as a default:

> Use only the top-level scene layer name instead (e.g. roads, instead of getting as granular as each sub-layer). This would still allow cases like roads and parks with the same name to repeat.

I'll need to review the examples a bit more... @bcamper is there a dev URL I can point at to play with this new code against the Mapzen scenes?

bcamper:
@nvkelso there is a preview build at:
https://line-label-repeats--tangram.netlify.com/dist/tangram.debug.js

And more generally, thanks to Netlify, you can always get a live demo of any branch at:
`https://branch-name--tangram.netlify.com/`

And library build at:
`https://branch-name--tangram.netlify.com/dist/tangram.debug.js`

🙌

Agree that top-level layer is a good middle-ground default.

bcamper:
@nvkelso will you be able to take a look at this soon? I'd like to get this fixed -- and any behavioral changes made -- for upcoming v0.17.0 release. I think this is now the only thing outstanding. Let me know if you need more assistance!

nvkelso:
👀

nvkelso:
I _do_ like the overall balance of labels better in the new branch – less minor road labels that cross major streets. But often there's duplicate major street labels introduced in that process. I think it's okay to knock some of the minor labels out, but not to make room for duplicate labels).

Here's zoom 15 in SF, with circle annotations this time!

sf_z15_label_change

nvkelso:
Zoom 14 is also arguably better overall balance, but with the dup's added in. Fell and Oak are noticeably better in the middle of the image (though Oak is newly duplicated).

image

bcamper:
OK what you're seeing makes sense, because this branch simply "fixes" the behavior to be the original intent, where sub-layers repeat independently, hence additional duplicates that are from different road layers. I think this just confirms that this is probably not the actual behavior we want.

I will update this branch to implement the suggestion to default the repeat logic to the top-level style layer, and then let's compare again.

nvkelso:
New setup looks good, ship it!

PR #694, opened 2019-01-22
File re-org

bcamper:
This PR moves many files, and extracts some functionality to a few new ones, to better organize them thematically, particularly:

- Group all scene-related files in `src/scene`.
- Group all lighting-related files in `src/lights`, including moving several GLSL files that were previously in `gl/shaders`, so that they are closer to the JS that actually uses them.
- Group selection-related files in `src/selection`.
- Moves other "orphaned" GLSL files to be closer to their relevant JS, including for styles more generally, and for raster shader globals specifically. (As a result, the entire `gl/shaders` folder is eliminated.)
- Extract logic for text segmentation (label splitting, RTL, bidi) and line wrapping to separate files.
- Similar move for tile-related files to `src/tile` was done recently in 2915abb21dab3deeffe88f530b6187a66f018279.

This started as a smaller move of just scene files, but quickly grew :) Better to get it all done at once I think. A downside is that by default, `git log` does not show history continuing through to the old files, however, you can do this when needed with `git log --follow`; though unfortunately, github's UI does not currently support this option. Ultimately decided it was better to have a more logical structure though, and the locations of several of these have been bugging me (and making it harder to trace how files are related, particularly JS + GLSL), for... years.


bcamper:
Forgot to note, generally speaking, git should correctly apply merges of existing branches/stashes from the old structure to the new one. Some manual adjustments are needed in a few cases (as with any complex merges).

bcamper:
discussed w/@meetar

PR #698, opened 2019-01-31
Fix raster sampling at explicit pixel locations with mixed vector/raster tile sizes

bcamper:
In doing some raster tests, @meetar found that the built-in `sampleRasterAtPixel()` raster shader method was not working properly when using mixed tile sizes, for example a 256px vector tile source with an attached 512px raster tile source. While the default / more common `sampleRaster()` method (used in the default raster style implementation) was working properly in this scenario, the explicit pixel look-ups were scrambled, with incorrect UV offsets. An example of the correct vs. actual results:

webp net-gifmaker 4

Digging in, it looks like the `sampleRasterAtPixel()` method was incorrectly applying a call to `adjustRasterUV()`, which is used to correctly offset tile UVs from smaller (e.g. 256px) tiles inside of a larger (e.g. 512px) raster tile texture. This adjustment is used in `currentRasterUV()`, which uses the tile's vertex positions. But it is incorrect to apply in the case of `sampleRasterAtPixel()`, where the final pixel location is already explicitly provided.

Removing this call fixes the problem. It appears this bug has always existed, but we have not previously caught it because we have never had a case where we are using `sampleRasterAtPixel()` with mixed tile size sources. `sampleRasterAtPixel()` is usually used to sample nearby pixels, such as for a kernel look-up. We have several such examples in https://github.com/tangrams/terrain-demos/, but these all operate on a single data source of terrain normal tiles. Mapzen styles such as Walkabout and Refill vector-masked raster tiles, but always of the same tile size.

To prevent a regression, I've added two Differ tests, for mixed vector/raster styles, with `sampleRaster()`, and `sampleRasterAtPixel()`:

https://github.com/tangrams/differ-tests/commit/e1a2e0e3d4c16ae6703e47ed5cf51366a1eabb57

screen shot 2019-01-31 at 3 08 40 pm

### DIFF
--- a/src/styles/raster/raster_globals.glsl
+++ b/src/styles/raster/raster_globals.glsl
@@ -33,7 +33,7 @@ uniform bool u_raster_mask_alpha;
 
 // Samples a raster tile texture for a given pixel
 #define sampleRasterAtPixel(raster_index, pixel) \
-    (texture2D(u_rasters[raster_index], adjustRasterUV(raster_index, (pixel) / rasterPixelSize(raster_index))))
+    (texture2D(u_rasters[raster_index], (pixel) / rasterPixelSize(raster_index)))
 
 // Returns size of raster sampler in pixels
 #define rasterPixelSize(raster_index) \


bcamper:
@matteblair @tallytalwar I believe this same issue exists in ES - could you please check?

tallytalwar:
The logic makes sense, will confirm if this is also visible in tangram-es
soon.

tallytalwar:
Yup seems to be the exact case in ES.

https://github.com/tangrams/tangram-es/blob/c3ac5438b23b87fd37b5f9030554168413cebd3c/core/shaders/rasters.glsl#L13

I will add the same fix, along with options to add `zoom_offset`. (https://github.com/tangrams/tangram-es/issues/2019).

@bcamper thanks for the heads up.

tallytalwar:
Yup I was able to reproduce the exact behavior and same fix resolves the issue.:
screen shot 2019-02-02 at 3 51 29 pm

Thanks @bcamper

PR #705, opened 2019-02-14
More layer matching control with `priority` and `exclusive`

bcamper:
This PR introduces two new keywords -- `priority` and `exclusive` -- for scene `layers`, to enable greater control over feature filter matching and styling composition.

**Background**
Tangram layer matching has always been inclusive, meaning a feature can match multiple layers (at multiple levels of the layer style tree), which are then merged to determine the final rendering parameters. Composing complex sub-layer trees can be used to capture both feature and zoom-dependent styling along orthogonal dimensions, for example varying road line widths and colors by separate criteria.

However, two in many ways simpler behaviors have *not* been easily expressed with the current layer syntax:

- The priority with which layers are matched has always been implicit. While layer depth takes priority (with an inheritance model), when layers at the same level match (siblings), the tie-breaker has been alphabetical, with the last lexically sorted layer name winning. This is cryptic at best and a significant hindrance to easy authoring where layer priority is important (see Bubble Wrap [shield matching](https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap-style.yaml#L996)).
- Since multiple sibling layers will match by default, there has been no way to express simple *exclusive* relationships, for example an ordered list of filter conditions, where the first matching filter should be used, and the others discarded. While it has often been possible to work around this by carefully building filters to be functionally exclusive (for example, defining a filter A, then defining a filter B that negates all of A's conditions, in addition to including its own), this is also an unwieldy approach, and does not serve all cases (such as when the filters have natural overlaps that can't easily be separated in a way that avoids a feature from matching more than once).

**New Syntax**
The `priority` and `exclusive` keywords address these cases, and can be used both separately and together.

- `priority`: allows for the explicit definition of the matching order for sibling layers (alphabetical by layer name is still used as the default / fallback).
- `exclusive`: marking a layer as `exclusive` ensures that when that layer matches, *no other* sibling layers (and their cascading sub-layers, etc.) will; when multiple `exclusive` layers match, the higher priority layer wins.

For example, `exclusive` could express a single `if-else` filter condition:

```
layers:
  if:
    filter: ...
    exclusive: true
  else:
    ...
```

When used together, `priority` and `exclusive` can be used for chained `if-elseif-else` filter cases, to ensure that only one of an ordered set of conditions matches:

```
layers:
  layerA:                 # if matches layerA...
    filter: ...
    priority: 1
    exclusive: true
    draw: ...
  layerB:                 # else if matches layerB...
    filter: ...
    priority: 2
    exclusive: true
    draw: ...
  layerC:                 # else default to layerC...
    priority: 3
    exclusive: true
    draw: ...
```

Separate from "flow control" cases, `priority` can simply be used on its own to disambiguate the precedence of multiple layers without reliance on naming conventions.

**Implementation**
The implementation is a fairly narrow impact on the layer matching logic:

- When determining layer matching and merging order:
  - Layers with an explicit `priority` value are matched first, with values sorted from lowest to highest (e.g. `priority: 1` is matched before `priority: 2`).
  - Layers without an explicit `priority` are matched after, and continue to be sorted lexically.
- Matching operation:
  - If a layer marked `exclusive` matches, all further matching at that level in the layer tree ceases and the matching layer is used (matching continues for that `exclusive` layer's child/descendant layers).
  - In the case that multiple layers are marked `exclusive`, the first one (according to the order described above) is used.

### DIFF
--- a/src/styles/layer.js
+++ b/src/styles/layer.js
@@ -6,7 +6,7 @@ import Geo from '../utils/geo';
 import { buildFilter } from './filter';
 
 // N.B.: 'visible' is legacy compatibility for 'enabled'
-const reserved = ['filter', 'draw', 'visible', 'enabled', 'data'];
+const reserved = ['filter', 'draw', 'visible', 'enabled', 'data', 'exclusive', 'priority'];
 
 let layer_cache = {};
 export function layerCache () {
@@ -59,16 +59,8 @@ export function mergeTrees(matchingTrees, group) {
             continue;
         }
 
-        // Sort by layer name before merging, so layers are applied deterministically
-        // when multiple layers modify the same properties
-        draws.sort((a, b) => (a && a.layer_name) > (b && b.layer_name) ? 1 : -1);
-
         // Merge draw objects
         mergeObjects(draw, ...draws);
-
-        // Remove layer names, they were only used transiently to sort and calculate final layer
-        // (final merged names will not be accurate since only one tree can win)
-        delete draw.layer_name;
     }
 
     // Short-circuit if not visible
@@ -83,7 +75,7 @@ const blacklist = ['any', 'all', 'not', 'none'];
 
 class Layer {
 
-    constructor({ layer, name, parent, draw, visible, enabled, filter, styles }) {
+    constructor({ layer, name, parent, draw, visible, enabled, filter, exclusive, priority, styles }) {
         this.id = Layer.id++;
         this.config_data = layer.data;
         this.parent = parent;
@@ -91,6 +83,8 @@ class Layer {
         this.full_name = this.parent ? (this.parent.full_name + ':' + this.name) : this.name;
         this.draw = draw;
         this.filter = filter;
+        this.exclusive = (exclusive === true);
+        this.priority = (priority != null ? priority : Number.MAX_SAFE_INTEGER);
         this.styles = styles;
         this.is_built = false;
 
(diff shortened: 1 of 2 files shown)


bcamper:
@nvkelso @meetar @matteblair @burritojustice this need for control over matching order and exclusivity recently came to a head for me. Having tried a few different approaches, this is where I've settled. I'm confident the use cases are real, and fairly satisfied with the balance of syntax and implementation footprint here (both less than I feared). But of course interested in opinions, possible alternatives, etc.!

nvkelso:
I dig it. Like you mentioned, we've had to do some silly gymnastics to accomplish layer filters in the Mapzen house styles and this new syntax would simplify that for the basemaps – and make it easier to style data viz. 2x win!

nvkelso:
Added Tangram ES sister issue in https://github.com/tangrams/tangram/issues/706.

matteblair:
Nice! We had discussed logic like this for Tangram ES, but actually for performance benefits rather than styling control. For layers with many siblings where the filters are effectively exclusive, matching can be much faster if sibling filter evaluation is "short circuited" like this. I can't immediately think of any reason we couldn't implement this in Tangram ES but I'll take a closer look later.

bcamper:
Performance is a good point! Yeah, in some ways `exclusive: true` might have been a better default, but would be problematic for backwards compatibility to change, so will probably end up just adding a lot of these lines to existing styles...

PR #712, opened 2019-04-09
Default to data source's only layer

bcamper:
Many data sources only have a single layer. With GeoJSON sources, the layer is implicit and un-named, because the contents of the file are a single `FeatureCollection` object. This layer gets automatically matched when a `data` block in `layers` leaves the source `layer` name unspecified, e.g.

```
layers:
  geojson:
    data: { source: geojson } # uses the default layer in the geojson source
    ...
```

Many MVT sources typically have a single layer as well, but there's no obvious or standard "default" layer name. Popular tools like Tippecanoe may [reuse the source file name as the layer name](https://github.com/mapbox/tippecanoe#input-files-and-layer-names) (you can explicitly name those layers if you want instead), or similar behavior. Other tools may just use a generic layer name like `features`.

This has come up repeatedly as a gotcha when using single-layer MVT sources in Tangram, because it's not always natural to realize you have to specify the layer in these cases.

This PR changes the data source layer matching behavior such that, for any data source that has a single layer, that layer will be matched in cases where no source `layer` name is specified in the `data` block. (If a `layer` name *is* specified, Tangram will still try to match on that, and return no data if the layer names don't match.)

### DIFF
--- a/src/tile/tile.js
+++ b/src/tile/tile.js
@@ -324,36 +324,31 @@ export default class Tile {
             geom: GeoJSON FeatureCollection
     */
     static getDataForSource (source_data, source_config, scene_layer) {
-        var layers = [];
+        const layers = [];
+        const single_layer = (Object.keys(source_data.layers).length === 1 && Object.keys(source_data.layers)[0]);
 
         if (source_config != null && source_data != null && source_data.layers != null) {
-            // If no layer specified, and a default source layer exists
-            if (!source_config.layer && source_data.layers._default) {
-                layers.push({
-                    geom: source_data.layers._default
-                });
-            }
-            // If no layer specified, and a layer for the scene layer name exists
-            else if (!source_config.layer && scene_layer) {
+            // If no source layer specified:
+            // 1. Look for a source layer that matches the scene layer name
+            // 2. If not, but the source only has one layer, use that as the default
+            if (!source_config.layer && (scene_layer || single_layer)) {
                 layers.push({
                     layer: scene_layer,
-                    geom: source_data.layers[scene_layer]
+                    geom: source_data.layers[scene_layer] || source_data.layers[single_layer]
                 });
             }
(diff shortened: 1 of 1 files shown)


nvkelso:
I've also come up against this a few times and I think this is a useful change. 

Both this and the other PR need Tangram ES changes, too.

bcamper:
@nvkelso yep, waiting to create ES issues until we decide to move ahead, but do need to do that! Thank you for the ones you made to track other recent additions :)

matteblair:
Seems useful!

I feel like this makes the data layer syntax harder to understand, the mental model of this option now requires more invisible variables and rules. However I would also believe that this "does the right thing" in practice for most situations.

How do you determine whether a data source has only a single layer? As I understand it, we don't know what layers are in a source until we get tiles from it. So how would we reason about an ocean tile that contains only the water layer? (I'm not sure if any known tile sources actually do that).

pnorman:
> How do you determine whether a data source has only a single layer?

You can't.

> As I understand it, we don't know what layers are in a source until we get tiles from it. So how would we reason about an ocean tile that contains only the water layer? (I'm not sure if any known tile sources actually do that).

I think that's fairly common.

bcamper:
> > How do you determine whether a data source has only a single layer?
> 
> You can't.
> 
> > As I understand it, we don't know what layers are in a source until we get tiles from it. So how would we reason about an ocean tile that contains only the water layer? (I'm not sure if any known tile sources actually do that).
> 
> I think that's fairly common.

Hmm yes that's a good point. I'd been thinking of data sources that are inherently or known in advance to be single-layer, but you are right that this could be a problem with basemap tiles. That would create inconsistent behavior here so we'll probably have to abandon this change, absent support for something with more formal source definition like TileJSON.

I think I will close this for now, though if others have suggested changes to make something like this work, please drop them here.

PR #715, opened 2019-04-21
Add support to parse some or all MVT feature properties as JSON

bcamper:
Sometimes source data includes properties with a richer object format than just strings or numbers -- e.g. arrays, nested objects, etc. The MVT spec [doesn't prescribe explicit behavior](https://docs.mapbox.com/vector-tiles/specification/#how-to-encode-attributes-that-arent-strings-or-numbers) for these cases, but notes that common tools such as [Tippecanoe](https://github.com/mapbox/tippecanoe) and Mapnik will encode these properties as stringified JSON.

This requires client-side support for parsing these fields. This PR adds support for this on MVT data sources through a new `parse_json` property:

- If `parse_json` is `true`, then *each* property will be checked to see if it "looks like" stringified JSON (defined as a string with first character being `{` or `[`); if so, it is parsed as JSON (with `JSON.parse`).
- If `parse_json` is an array of property names, only those specific properties are checked for JSON parsing. This is preferred to the above, because it limits the parsing impact to only the fields that need it.
- If `parse_json` is undefined/null/false, no special parsing of properties is done (e.g. the current behavior).

Example usage:

```
sources:
  tiles:
    type: MVT
    url: https://...
    parse_json: [prop_a, prop_b] # treat feature properties 'prop_a' and 'prop_b' as stringified JSON
```

These parsed properties will now be properly returned as full JS objects through functions such as `scene.getFeatureAt()` or `scene.queryFeatures()`.

To make use of these properties in filtering and styling, you will need to write a JS function to access the complex properties, e.g. if `prop_a` is an array of values, you could filter on it containing the value `x` like so:

`filter: function() { return feature.prop_a.indexOf('x') > -1; }`

### DIFF
--- a/src/sources/mvt.js
+++ b/src/sources/mvt.js
@@ -1,8 +1,17 @@
 import DataSource, {NetworkTileSource} from './data_source';
 import Geo from '../utils/geo';
+import log from '../utils/log';
 
 import Pbf from 'pbf';
 import {VectorTile, VectorTileFeature} from '@mapbox/vector-tile';
+
+const PARSE_JSON_TYPE = {
+    NONE: 0,
+    ALL: 1,
+    SOME: 2
+};
+
+const PARSE_JSON_TEST = ['{', '[']; // one-time allocated array/strings
 
 /**
  Mapbox Vector Tile format
@@ -13,6 +22,27 @@ export class MVTSource extends NetworkTileSource {
     constructor (source, sources) {
         super(source, sources);
         this.response_type = 'arraybuffer'; // binary data
+
+        // Optionally parse some or all properties from JSON strings
+        if (source.parse_json === true) {
+            // try to parse all properties (least efficient)
+            this.parse_json_type = PARSE_JSON_TYPE.ALL;
+        }
+        else if (Array.isArray(source.parse_json)) {
+            // try to parse a specific list of property names (more efficient)
+            this.parse_json_type = PARSE_JSON_TYPE.SOME;
+            this.parse_json_prop_list = source.parse_json;
+        }
+        else {
+            if (source.parse_json != null) {
+                let msg = `Data source '${this.name}': 'parse_json' parameter should be 'true', or an array of ` +
+                    `property names (was '${JSON.stringify(source.parse_json)}')`;
+                log({ level: 'warn', once: true }, msg);
+            }
+
+            // skip parsing entirely (default behavior)
+            this.parse_json_type = PARSE_JSON_TYPE.NONE;
+        }
     }
 
     parseSourceData (tile, source, response) {
@@ -55,6 +85,8 @@ export class MVTSource extends NetworkTileSource {
                     properties: feature.properties
                 };
 
+                this.parseJSONProperties(feature_geojson);
+
                 var geometry = feature_geojson.geometry;
                 var coordinates = feature.loadGeometry();
                 for (var r=0; r < coordinates.length; r++) {
@@ -92,6 +124,29 @@ export class MVTSource extends NetworkTileSource {
         return layers;
     }
 
+    // Optionally parse some or all feature properties from JSON strings
+    parseJSONProperties (feature) {
+        if (this.parse_json_type !== PARSE_JSON_TYPE.NONE) {
+            const props = feature.properties;
+            for (const p in props) {
+                // if specified, check list of explicit properties to parse
+                // (otherwise try to parse all properties)
+                if (this.parse_json_type === PARSE_JSON_TYPE.SOME &&
+                    this.parse_json_prop_list.indexOf(p) === -1) {
+                    continue; // skip this property
+                }
+
+                // check if this property looks like JSON, and parse if so
+                if (PARSE_JSON_TEST.indexOf(props[p][0]) > -1) {
+                    try {
+                        props[p] = JSON.parse(props[p]);
+                    } catch (e) {
+                        // continue with original value if couldn't parse as JSON
+                    }
+                }
+            }
+        }
+    }
 }
 
 // Decode multipolygons, which are encoded as a single set of rings
--- a/src/sources/mvt.js
+++ b/src/sources/mvt.js
@@ -128,20 +128,27 @@ export class MVTSource extends NetworkTileSource {
     parseJSONProperties (feature) {
         if (this.parse_json_type !== PARSE_JSON_TYPE.NONE) {
             const props = feature.properties;
-            for (const p in props) {
-                // if specified, check list of explicit properties to parse
-                // (otherwise try to parse all properties)
-                if (this.parse_json_type === PARSE_JSON_TYPE.SOME &&
-                    this.parse_json_prop_list.indexOf(p) === -1) {
-                    continue; // skip this property
-                }
 
-                // check if this property looks like JSON, and parse if so
-                if (PARSE_JSON_TEST.indexOf(props[p][0]) > -1) {
+            // if specified, check list of explicit properties to parse
+            if (this.parse_json_type === PARSE_JSON_TYPE.SOME) {
+                this.parse_json_prop_list.forEach(p => {
                     try {
                         props[p] = JSON.parse(props[p]);
                     } catch (e) {
                         // continue with original value if couldn't parse as JSON
+                    }
+                });
+            }
+            // otherwise try to parse all properties
+            else {
+                for (const p in props) {
+                    // check if this property looks like JSON, and parse if so
+                    if (PARSE_JSON_TEST.indexOf(props[p][0]) > -1) {
+                        try {
+                            props[p] = JSON.parse(props[p]);
+                        } catch (e) {
+                            // continue with original value if couldn't parse as JSON
+                        }
                     }
                 }
             }
--- a/src/styles/filter.js
+++ b/src/styles/filter.js
@@ -90,26 +90,18 @@ function rangeMatch(key, value, options) {
 function includesMatch(key, value) {
     let expressions = [];
 
-    // the array includes ONE OE MORE of the provided values
-    if (value.includes) {
-        if (Array.isArray(value.includes)) {
-            let arr = '['+ value.includes.map(maybeQuote).join(',') + ']';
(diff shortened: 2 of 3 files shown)


bcamper:
@nvkelso I think we also had some values similar to this in TIlezen for awhile (for multiple shields?), but removed them (for unrelated reasons).

nvkelso:
Tilezen still exports the array properties in GeoJSON and TopoJSON, but currently strips out in MVT. We can start exporting them in MVT when Tangram grows support. Primarily these are multishields (road, transit, walk, bike).

bcamper:
> Tilezen still exports the array properties in GeoJSON and TopoJSON, but currently strips out in MVT. We can start exporting them in MVT when Tangram grows support. Primarily these are multishields (road, transit, walk, bike).

Oh great! Bring it on :)

bcamper:
@matteblair thoughts on this on the ES side?

matteblair:
This would be a bigger change for Tangram ES since we currently model feature properties explicitly as a mapping of strings to scalar values. However I can see the value in supporting more structured data for properties.

This is also a topic of discussion for the next version of the MVT spec (https://github.com/mapbox/vector-tile-spec/issues/75). It does not seem like the proposed solution in MVT would be mutually exclusive with the one here, but it would make this one redundant. The MVT proposal also has the advantage that it's a much more compact representation of structured data: values can be de-duplicated in MVT but in stringified JSON they must be duplicated.

bcamper:
> This would be a bigger change for Tangram ES since we currently model feature properties explicitly as a mapping of strings to scalar values. However I can see the value in supporting more structured data for properties.
> 
> This is also a topic of discussion for the next version of the MVT spec ([mapbox/vector-tile-spec#75](https://github.com/mapbox/vector-tile-spec/issues/75)). It does not seem like the proposed solution in MVT would be mutually exclusive with the one here, but it would make this one redundant. The MVT proposal also has the advantage that it's a much more compact representation of structured data: values can be de-duplicated in MVT but in stringified JSON they must be duplicated.

I agree the eventual MVT v3 solution will be better, however after some time (that ticket was opened in 2016...) I decided to finally move ahead with this because: 1) there are popular tools that implement this JSON stringify behavior, particularly Tippecanoe, so it's a reality for a lot of existing data, and 2) MVT v3 is going to be a substantial upgrade (tons of other things in the spec) that will require widespread updates to both encoding and decoding libs and I think a lot of MVT v2 data will persist.

matteblair:
That makes sense. We probably shouldn't count on the MVT v3 spec being finished any time soon, and even with that approach we would have to update the feature data model in Tangram ES. This feature sounds good to me.

matteblair:
I'm also interested in finding ways to make these properties available to filters in a "native" way, since using a JS function in a filter makes it an order of magnitude slower in Tangram ES. Having built-in accesors for structured properties could also implicitly define which ones need to be parsed, and allow us to do it lazily.

bcamper:
Agree on finding adding support for more native syntax. I'm mostly concerned about how to do it in a way that doesn't impact performance, since the assumption of a 1:1 mapping from YAML string to feature property key is fast. But I'm sure we can use some clues from data source definition and/or caching strategy to mitigate this.

bcamper:
OK @matteblair inspired me to take this further. I've added support both for accessing nested properties with native syntax in filters, and for doing array-to-array filter operations. I'm not 100% sure of the syntax/naming decisions here, so feedback/alternatives welcome.

## Nested properties in filters
Dot notation with `.` can be used to access nested feature properties (these properties could have been encoded in a GeoJSON/TopoJSON source, or parsed from MVT stringified JSON properties as covered by this PR):

Given a feature property `a: { b: { c: 'test' } }`, this filter will match:

`filter: { a.b.c: test }`

Feature property names that include a `.` can be escaped with `\.`, e.g. a feature property named `'d.e.f': 'escaped'` will match with:

`filter: { d\.e\.f: escaped }`

These could be mixed, e.g. a property `{ 'a.b': { c: 'mixed' }` would match with:

`filter: { a\.b.c: mixed }`

## Array-to-array filters
We have existing syntax for checking if a *single value* matches *one of* a list of values, e.g.:

Does value `x` match `a`, `b`, or `c`?

`filter: { x: [a, b, c] }`

As part of this new support for "complex" property values that may be arrays, we need to expand this syntax. This branch adds two keywords for this, by extending the same filter object pattern we use for `min/max` range filter syntax:

To check if an array `a` contains **one or more of** the values `p`, `q`, `r`:

`filter: { a: { includes: [p, q, r] } }`

Note: while it would be possible to check the type value of `a` to see if it's an array, and apply this logic without additional syntax, I was weary of introducing an additional run-time check that can occur on many many features (vs. the other additions in this branch, which only occur once per filter, lazily when the filter is first parsed/built), especially for a case that is only needed for a small subset of feature properties. On the other hand, [jsperf](https://jsperf.com/array-isarray-vs-instanceof-array) says my browser can perform 500 million array type checks per second, so this may be unfounded.

There are also cases where you'd like to test that an array contains **all of** a list of values. An `includes_all` variant was added for this. For instance, to check if an array `a` contains the values `p`, `q`, **AND** `r`:

`filter: { a: { includes_all: [p, q, r] } }`

One could check for multiple array values with multiple filters, but supporting direct syntax for them would be much more efficient than this, which will execute 3 separate array tests, e.g.:

```
filter:
  all:
    - a: { includes: p }
    - a: { includes: q }
    - a: { includes: r }
```

Note: `includes` (implicit "or"/"one of", like existing single-value-to-array filters) vs. `includes_all` doesn't feel entirely right, but I couldn't come up with something I liked better, suggestions welcome!

matteblair:
I like the property accesors! Should be both more performant and easier to use for authoring filters. I'd suggest `includes_any` instead of `includes` just to make the meaning more obvious.

matteblair:
An alternative way of accessing nested properties in YAML would be to use arrays for the filter keys, like:
```
filter: { [a, b, c]: test }
```
But I don't think this works in the JS YAML model - plus it's not very natural to read.

The `\.` escaping just seems a little awkward so I'm trying to think of other syntax options. Hopefully it should be very rare that we encounter data with property keys including `.`, so it's probably not worth worrying about.

nvkelso:
I dig it, and @matteblair suggestion for `includes_any`.

> filter: { d\.e\.f: escaped }

For the complicated property names, could we allow that be be quoted like `"d.e.f"`, too?

nvkelso:
@bcamper can you add an example on how to access the multishields in the `all_shields` properties in Tilezen's GeoJSON output, which are arrays (so first element in the array, or for each element in the array)? 

Will Tangram abstract that GeoJSON and MVT store it in different ways, but I think Tangram will expand the MVT shorthand to the GeoJSON internally?

bcamper:
Agree on `includes_any`, will update.

Yep, property names can be quoted or unquoted (e.g. `d.e.f` or `"d.e.f"`), as per standard YAML syntax (it's just personal preference).

@matteblair you're following my thinking too -- I considered an array syntax, but you're correct JS YAML parser doesn't support it, and them it seemed just as complicated or more to parse it out manually from a string key. And this type of dot notation would be valid JS. Agree `\.` is a bit awkward, but I concluded it was common enough as a string escaping pattern in other cases. It could also use JS-style bracket notation like `a[b][c]`, but this similarly felt a little more involved to parse and still requires some form of escaping (to support `[` and `[`) if you're really trying to be complete.

bcamper:
@nvkelso can you provide an example tile/location where that property shows up? Not finding it in a quick search.

nvkelso:
> "all_networks":["US:US","US:CA"],"all_shield_texts":["101","128"]

- [12-648-1568.topojson.json.zip](https://github.com/tangrams/tangram/files/3150049/12-648-1568.topojson.json.zip)


image

bcamper:
@nvkelso 

> "all_networks":["US:US","US:CA"],"all_shield_texts":["101","128"]

Can you elaborate on how you think we would want to filter on and use these array fields? For example, are the index positions 0, 1, etc. going to reliably reflect a fixed network "level" (national highway, state highway, etc.)? It seems like we couldn't necessarily guarantee that for all data?

It's easy enough to match on the values, like "does `all_networks` contain `US:CA`, but I don't have an immediate answer for how we would then match that index to `all_shield_texts` to get the desired shield text out. We could put the whole object into one array (e.g. something like `[{ network: 'US:US', text: '101' }, { network: 'US:CA', text: '128' }]`), but that also doesn't fit simple filter matching syntax (it's more like the complex destructuring matching offered JS and other languages... which is cool but likely overkill here!); so we'd probably want a JS filter function for it as currently stored.

Let's figure out how we want to style them, then determine what additional scene syntax and/or data format changes are needed (since they aren't in use anyway, right?).

@matteblair other thoughts?

nvkelso:
Mostly I'm interested in multishield support, for roads, walking paths, bicycle, and transit. As long as this gets us part way there, and it's possible to step all the way there with an extension then I'm good.

matteblair:
I don't know the full range of data for shields, but for the case above it we might be able to use an object property like:
```json
"networks": { "US:US": "101", "US:CA": "128" }
```
Then do filtering and styling like:
```yaml
filter: { networks.US:US : true }
draw:
  text:
    text_source: networks.US:US
```
This implies that property accessors work in styling expressions as well as filtering expressions - not sure if that was planned or not.
Encoding this way is also not very compact if we're stringifying the whole JSON object :\

nvkelso:
😂 I think this is taking a complex turn when I'm just interested in multi-shields ;)

bcamper:
It's the tangram way. I feel comfortable we can prototype multi-shields
with this data and some of the new syntax, and then go from there.

matteblair:
@nvkelso I don't think I understand what you mean when you say "multi-shields". Can you explain?

nvkelso:
The section of road given as an example above has data to support 2
shields, a US route and a California route. On many maps both routes would
have a shield drawn, so 2 shields total in this example. Often this is zoom
13+ as mid zooms can only accommodate 1 shield for density issues. But both
routes will be signed in real life, on the ground.

Issue #576, opened 2017-08-09
override colors for all `roads` sublayers, at once


burritojustice:
I'd like to be able to assign colors to roads in any sublayer in a Mapzen basemap at the top level of the layer, without having to dive into each sublayer.

Doing so now is possible, but it assumes a detailed knowledge of the layer hierarchy which is a challenge for one Mapzen basemap style, never mind all of them.

Right now, it has to be done this way. (This assumes I have a function defining the color.

```
global:
    my_colors: ...
    
layers:
    roads:
        filter: {...}
        z-my_colors: global.my_colors
        major_road:
            z-my_colors: global.my_colors
            trunk_primary:
                z-my_colors: global.my_colors
                routes:
                    z-my_colors: global.my_colors
            secondary:
                z-my_colors: global.my_colors
            tertiary:
                z-my_colors: global.my_colors
                routes:
                    z-my_colors: global.my_colors
        	...
```
an example:

https://mapzen.com/tangram/play/?api=22%2F872#15.1500/37.7737/-122.4122

It would be much simpler and less error prone to do something like this:

```
layers:
    roads:
        my_stuff_is_more_important:
            filter: {...}
            z-johns_colors: global.johns_colors
```
Note that I inadvertently got this to work -- in the example below, while the sublayer is `major_road`, the colors apply to *all* roads, regardless of the sublayer name.

https://mapzen.com/tangram/play/?api=22/872#14.4667/37.7773/-122.4108

```
        major_road:
            z:
                early:
                    draw:
                        lines:
                            color: |
                                function(){
                                    var name = feature.name
                                    if (name.match(/Harrison/i)) {
                                       return 'blue'
                                        }
                                    if (name.match(/Howard/i)) {
                                       return 'lightgreen'
                                        }
                                    if (name.match(/Folsom/i)) {
                                       return 'orange'
                                        } 
                                    if (name.match(/Mission/i)) {
                                       return 'red'
                                        }  
                                    else {return 'black'}
                                }
                            # color: lightgreen
                            width: 2px
                            order: 500
```

I don't exactly know _why_ this works. Note also that if you change `z` to `a`, some of the roads disappear, so this seems to be parsed in alphabetical order?

Anyway, a more stable method of doing this would be useful.

bcamper:
cc @matteblair @meetar @nvkelso 

(Side note, the example links in this issue aren't working.)

So yeah you've come up against a style precedence issue.

Tangram `draw` group specificity
===
[...]

PR #717, opened 2019-05-03
Allow draw groups to `freeze` parameter values through sub-layers

bcamper:
OK a couple years later I'm returning to this override issue #576. Here's my latest attempt...

This introduces a `freeze` keyword to `draw` groups. When a draw group contains the value `freeze: true`, any values in that group (only those values defined "locally" in the same immediate block as the `freeze` keyword) will be "frozen", meaning they will be unaffected by other draw group values, either from child/descendants in the same layer tree, or from parallel/sibling layer trees.

The simplest way to think of this is that it will create a draw group with infinite layer depth, and that is how it is implemented: after all draw groups are merged across both "across" (parallel layer trees) and "down" (layer -> sub-layers) with existing logic, any frozen groups are then additionally applied, as if they were in the next sub-layer down.

On balance, I think this is a helpful solution, but still has potential issues and is just a tricky problem to solve... suggestions welcome! The implementation footprint is small, it's more a matter of getting the logic right (in a way that keeps the implementation simple/performant :) 

## Basemap override

Returning to one of the common cases where you want to "globally" override one or more values, here is an example that re-colors all roads in an imported Refill:

```
import:
    - https://www.nextzen.org/carto/refill-style/12/refill-style.zip
    - https://www.nextzen.org/carto/refill-style/12/themes/label-10.zip

layers:
    roads:
        override-colors:
            draw:
                lines:
                    freeze: true
                    color: red
                    outline:
                        color: darkred
```

No matter what other draw groups are defined, at the end of the draw group merge process, an additional draw group that looks like this will be applied:

```
draw:
  lines:
    color: red
    outline:
      color: darkred
```

This yields the following:
tangram-1556902003464

## Priority
These draw groups are applied with the same priority/specificity as any other layer, so if you have two or more conflicting "frozen" groups...

`override-colors2` will win over `override-colors`, because it has a higher priority due to last-alphabetical-layer-name-wins logic:
```
layers:
    roads:
        override-colors:
            draw:
                lines:
                    freeze: true
                    color: red
                    outline:
                        color: darkred

        override-colors2:
            draw:
                lines:
                    freeze: true
                    color: blue
                    outline:
                        color: darkblue

```

`override-colors` will win over `override-colors2`, because it has a higher explicit `priority` defined:
```
layers:
    roads:
        override-colors:
            priority: 1
            draw:
                lines:
                    freeze: true
                    color: red
                    outline:
                        color: darkred

        override-colors2:
            priority: 2
            draw:
                lines:
                    freeze: true
                    color: blue
                    outline:
                        color: darkblue
```

Similarly, if a layer marked `exclusive` matches, it will take precedence.

## Freeze at multiple levels

A child layer won't modify values of its frozen ancestor, e.g. in this case, the `child` layer won't change the color, which will remain frozen as `red`:

```
layers:
    roads:
        override-colors:
            draw:
                lines:
                    freeze: true
                    color: red
                    outline:
                        color: darkred

            child:
                draw:
                    lines:
                        color: yellow # has no effect because ancestor layer is frozen
```

If multiple levels of the tree contain `freeze: true`, they are merged, but with **inverse** priority from the normal deepest-level-wins logic: instead, the higher-up/ancestor groups will win (**does this make sense, or is it backwards...? I'm trying to think through the use cases**). Also as a result, marking a descendant draw group as `freeze: false` has no effect (**though I suppose this could be useful, as a way to say "you definitely cannot override this value"?**).

## Imports

There's a potential gotcha with imports (which would be likely to be mixed with this feature due to its nature), in that if you are importing an existing draw group and then applying `freeze` to it, you're also going to freeze the **whole** draw group, including any properties that were imported -- because import merges happen at the beginning of scene parsing, long before layers are parsed and merged (and it's likely not really practical to change that).

That's why in the example above, a top-level draw group with no filter and adjacent to others was created, called ` override-colors`, rather than just applying `freeze` to the existing top-level `draw` group under the `roads` layer.

### DIFF
--- a/src/styles/layer.js
+++ b/src/styles/layer.js
@@ -27,9 +27,8 @@ function cacheKey (layers) {
 
 // Merge matching layer trees into a final draw group
 export function mergeTrees(matchingTrees, group) {
-    let draws, treeDepth = 0;
-
     // Find deepest tree
+    let treeDepth = 0;
     for (let t=0; t < matchingTrees.length; t++) {
         if (matchingTrees[t].length > treeDepth) {
             treeDepth = matchingTrees[t].length;
@@ -42,26 +41,38 @@ export function mergeTrees(matchingTrees, group) {
     }
 
     // Merged draw group object
-    let draw = {
-        visible: true, // visible by default
+    const draw = {
+        visible: true // visible by default
     };
 
-    // Iterate trees in parallel
+    // Iterate layer trees in parallel
+    const draws = []; // matching draw groups in priority order
+    const frozen = []; // any frozen draw groups, in priority order
     for (let x=0; x < treeDepth; x++) {
         // Pull out the requested draw group, for each tree, at this depth (avoiding duplicates at the same level in tree)
-        draws = [];
-        matchingTrees.forEach(tree => {
+        matchingTrees.forEach((tree, i) => {
             if (tree[x] && tree[x][group] && draws.indexOf(tree[x][group]) === -1) {
                 draws.push(tree[x][group]);
+
+                // Set or unset frozen draw group for this layer tree
+                if (tree[x][group].freeze === true) {
+                    frozen[i] = tree[x][group];
+                }
+                else if (tree[x][group].freeze === false) {
+                    frozen[i] = null;
+                }
             }
         });
         if (draws.length === 0) {
             continue;
         }
-
-        // Merge draw objects
-        mergeObjects(draw, ...draws);
     }
+
+    // Add frozen draw groups (in layer priority order)
+    draws.push(...frozen.filter(x => x));
+
+    // Merge draw objects
+    mergeObjects(draw, ...draws);
 
     // Short-circuit if not visible
     if (draw.visible === false) {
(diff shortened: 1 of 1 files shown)


bcamper:
Call for comments... @matteblair @nvkelso @meetar @burritojustice

nvkelso:
Seems useful, though I find the name `freeze` confusing. Is there a more common term for this?

bcamper:
Ha, yeah I've always struggled with naming this. Freeze was just the closest conceptual match I could find to what's happening. In CSS there is `important`, but that is... worse IMO. Happy for alternatives.

matteblair:
From my understanding of the feature maybe `final: true` is a good name? Maybe too OOP-ish though. Still trying to grok all the cases here.

nvkelso:
Yeah, CSS `!important`, ftw ;) What about? 

- `prevail`
- `evaluate_last`

Others: 

- `freeze`
- `final`

bcamper:
`prevail` lol

`winner: true`

bcamper:
Decided not to pursue this.

PR #718, opened 2019-06-05
Change default `repeat_distance` for point feature labels

bcamper:
The default `repeat_distance` for labels has always been `256px` across the board. This is a sensible default for cases such as street labels (the original use case for repeat label culling) and road shields, but can generate unexpected results particularly when applied to point labels.

For example, when labelling data with a property that has a finite set of values, such as category names ("bar", "restaurant", "hardware store", etc.) or numeric values, it's often surprising/undesirable that only a few labels are plotted.

This branch changes this behavior such that:

- For labels generated **from point features**, the default `repeat_distance` is `0px`.
- For **all other labels**, including those generated **from line or polygon features** (e.g. street names, road shields, park names, etc.), the default `repeat_distance` continues to be `256px`. Note that in these examples, the *labels themselves* may still be point labels -- it is the underlying geometry type that matters, e.g. road shields are point labels generated from line geometries.

This example illustrates the difference in point geometry behavior:

```
        draw:
            points:
                color: red
                size: 4px
                collide: false
                text:
                    text_source: function(){ return 1 }
                    font:
                        color: white
                        stroke: { color: black, width: 4px }
```

In the prior behavior, only a few point labels are shown, and in the new behavior, all points are rendered:

Webp net-gifmaker

### DIFF
--- a/src/styles/text/text_labels.js
+++ b/src/styles/text/text_labels.js
@@ -300,7 +300,7 @@ export const TextLabels = {
 
         // Repeat rules - for text labels, defaults to tile size
         draw.repeat_distance = StyleParser.createPropertyCache(
-            draw.repeat_distance != null ? draw.repeat_distance : Geo.tile_size,
+            draw.repeat_distance,
             StyleParser.parsePositiveNumber
         );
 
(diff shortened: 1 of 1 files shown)


bcamper:
@matteblair @meetar @nvkelso we talked about this a few months ago... finally getting around to it.

nvkelso:
LGTM!

PR #720, opened 2019-06-09
More robust support for `feature.id`

bcamper:
All data formats supported by Tangram have a `feature.id` property (at the top `feature` level, outside `feature.properties`; some data sets may separately include an id or similar property within `feature.properties`), but we haven't had full support for it. This PR fixes that:

- Ensures `feature.id` persists wherever features are internally copied/synthesized/etc. internally
- Provides access to `feature.id` in user-authored JS scene functions with a new `$id` variable
- Includes `feature.id` in feature object results returned from `scene.queryFeatures()`
- Includes `feature.id` in the default uniqueness check for `scene.queryFeatures()`, with `$id` syntax used if specifying a list of properties to uniqueify on, e.g.:
  - For default `scene.queryFeatures({ unique: true })`, `feature.id` will be included in uniqueness check
  - For explicit unique property list `scene.queryFeatures({ unique: ['$id'] })`, `feature.id` can be included in uniqueness check with `$id` (un-prefixed names refer to `feature.properties`)

### DIFF
--- a/src/scene/scene.js
+++ b/src/scene/scene.js
@@ -844,13 +844,15 @@ export default class Scene {
         // Optional uniqueify criteria
         // Valid values: true, false/null, single property name, or array of property names
         unique = (typeof unique === 'string') ? [unique] : unique;
+        const uniqueify_on_id = (unique === true || (Array.isArray(unique) && unique.indexOf('$id') > -1));
         const uniqueify = unique && (obj => {
-            const props = Array.isArray(unique) ? sliceObject(obj.properties, unique) : obj.properties;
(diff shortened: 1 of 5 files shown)


bcamper:
@matteblair @tallytalwar I think the most salient thing here for ES is addition of `$id` syntax for JS functions (which was the most natural way I could imagine access for this, by extending the `$` syntax for other variables outside of `feature.properties`).

bcamper:
Discussed in chat.

Updated to ensure `$id` works in filter objects (e.g. `filter: { $id: 5 }`) in addition to filter JS functions (`filter: function(){ return $id == 5; }`).

PR #721, opened 2019-06-09
Default to WebGL "high performance" mode

bcamper:
Defaults to using `high-performance` for WebGL's `powerPreference` setting. This will ask for use of the discrete GPU where possible. The user can override as with other GL context options:

```
const layer = Tangram.leafletLayer({ scene, { webGLContextOptions: { powerPreference: 'low-power' } } });
```


bcamper:
Discussed in chat and further with @meetar.

Issue #741, opened 2020-01-08
Setting color to be transparent also makes dash_background_color transparent


rwrx:
I would like to kindly ask you a question regarding color transparency. I have this use case:

The first color `track_background_color` is set as transparent. The second color `track_color` is not transparent, this color is used in style `dashed_grade5` as `dash_background_color`. When this style is applied, all lines are transparent although the `dash_background_color` is not set as transparent and we want to have it opaque. Should it be like this or is it an error which can be fixed? Here are snippets from actual scene file:

```
global:
    ...
    track_background_color: [1.0, 1.0, 1.0, 0.3]
    track_color: [0.6, 0.4, 0.0]
    ...
   
styles:
    ...
    dashed_grade5:
        base: lines
        blend: translucent
        dash: [1.5, 2.5]
        dash_background_color: global.track_color
        material:
            diffuse: 1.0
            ambient: 1.0
    ...
   
layers:
    ...
    roads:
        ...
        grade5:
            filter: { tracktype: grade5 }
            draw:
                lines:
                    style: dashed_grade5
                    color: global.track_background_color
                    width: [ [13, 1px], [16, 3px]]
                    visible: true
                    outline:
                        color: global.track_background_color
                        dash_background_color: global.track_background_color
                        width: 5px
```

![dashed](https://user-images.githubusercontent.com/2378743/72020571-5d378d80-326c-11ea-9699-c9e3eae95420.png)

bcamper:
Hello, thanks for your example styling -- I do think this is a bug. Based on your description and style, I think this is what you would have expected, can you confirm?

![tangram-@17 8875 37 7936 -122 4628_2020-01-10_00 54 50](https://user-images.githubusercontent.com/16733/72209008-f8e91980-3477-11ea-8dfd-8f5923822b6f.png)

rwrx:
Yes, this is what I expected to see with that styling. I am also not sure if this is also present in Tangram-ES.

bcamper:
Thanks, I've opened the #742 PR to fix this (and related logic issues with dash/line texture/blend mode behavior). This issue **does** appear to exist in Tangram ES as well, we will take care of opening a corresponding issue there.

rwrx:
Thank you a lot.

bcamper:
Fix released in [v0.20.1](https://github.com/tangrams/tangram/releases/tag/v0.20.1)

PR #742, opened 2020-01-12
Fix line dash color handling

bcamper:
This PR revamps the logic for applying line dash patterns and textures. The prior logic had some holes and inconsistencies, including #741. It implements the following logic in the fragment shader for lines (and also replaces some conditional `if/else` shader branching with nested `mix` functions):

- A line can have an active texture sampler in two cases:
  - Implicitly, through a `dash` pattern (the texture is automatically generated from the dash pattern).
    - The dash pattern defines a foreground color (via the regular `color` parameter, available in the shader as the vertex `color`) and background color (via the `dash_background_color` parameter, available in the fragment shader as `u_dash_background_color`).
    - The dash texture is a monochrome representation of the pattern, with opaque white for the foreground and transparent black for the background. By using a monochrome texture, the same texture can be reused for the same dash pattern with different foreground/background color combinations.
  - Explicitly, assigned via the `texture` parameter (in `style` or `draw` group).
    - The color values in the texture are meant to be used directly (similar to other style texture sampling behaviors).
- If a line has a dash texture, then in the fragment shader:
  - A uniform `u_has_dash` indicates this, with a float value of `0` or `1`.
  - We determine whether the fragment is in the foreground or background of the dash pattern, and assign the corresponding color, using a `mix` function controlled by the line texture alpha channel (following the monochrome texture described above).
    - `mix(u_dash_background_color, color, _line_color.a), // choose dash foreground or background color`
- If the line has an explicit texture, then in the fragment shader:
  - Apply the vertex `color` to the texture color ("tinting" it), consistent with other similar behaviors, e.g. raster textures on polygons.
    - `color * _line_color, // no dash: tint the line texture with the vertex color`
- The determination of which texture type (dash pattern or explicit texture) to use is accomplished through another `mix` function, controlled by the `u_has_dash` uniform described above.
- Finally, for alpha/blending behavior:
  - When `opaque` blending is active, a simple "alpha cutout" behavior is applied, in which any alpha pixel with an alpha of less than 0.5 (`TANGRAM_ALPHA_TEST`) is discarded. This enables basic transparency-like behavior for dashes and textures when using opaque blending.
  - When other blend modes are active, no special logic is applied, and the alpha value resulting from the above color logic will be used as-is.

### DIFF
--- a/src/styles/lines/lines.js
+++ b/src/styles/lines/lines.js
@@ -391,6 +391,7 @@ Object.assign(Lines, {
 
                     if (variant.dash) {
                         uniforms.u_v_scale_adjust = Geo.tile_scale * DASH_SCALE;
+                        uniforms.u_has_dash = (variant.dash_background_color != null ? 1 : 0);
                         uniforms.u_dash_background_color = variant.dash_background_color || [0, 0, 0, 0];
                     }
 
--- a/src/styles/polygons/polygons_fragment.glsl
+++ b/src/styles/polygons/polygons_fragment.glsl
@@ -58,7 +58,7 @@ void main (void) {
             // which either don't support alpha, or would cause transparent pixels to write to the depth buffer,
             // obscuring geometry underneath.
             #ifdef TANGRAM_HAS_MASKED_RASTERS   // skip masking logic if no masked raster sources
-            #ifndef TANGRAM_ALL_MASKED_RASTERS  // skip conditional if *only* masked raster sources (always true)
+            #ifndef TANGRAM_ALL_MASKED_RASTERS  // skip source check for masking if *all* raster sources are masked
             if (u_raster_mask_alpha) {
             #else
             {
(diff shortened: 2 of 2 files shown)


bcamper:
@matteblair It looks like the same issue described in #741 also exists in Tangram ES (because it looks like it more or less copies the previously faulty Tangram JS logic... 😅). I took a couple stabs at fixing it, and found some other edge cases (the alpha cutout test was too early, causing inconsistent alpha handling depending on texture + blend mode combinations), and found it easier to just revamp to the logic in this PR, which hopefully is clearer (but using nested `mix` which isn't always intuitive either). Please give it a look and let me know if it makes sense!

We should open a corresponding issue to #741 for ES as well.

bcamper:
Actually, I don't think Tangram ES suffers from the same specific behavior reported in #741, current logic here:

https://github.com/tangrams/tangram-es/blob/master/core/shaders/polyline.fs#L67-L82

But, I still think it has related issues, particularly with the order of the alpha cutout test vs. the tinting (vertex color multiplied by line texture color), and would benefit from a review in conjunction with this PR.

bcamper:
Hey @matteblair, any further thoughts here? I'd like to cut a minor release with @meetar's geojson-vt upgrade. I tried to be complete with the logic description in the PR, but the actual shader code is pretty concise ;)

PR #744, opened 2020-01-24
update geojson-vt to ^3.2.1

meetar:
Updating geojson-vt for performance and edge-case handling improvements. Passed all Differ tests with only very minor changes to text – if anything it's made text sharper?

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "csscolorparser": "1.0.3",
     "earcut": "2.1.1",
     "fontfaceobserver": "2.0.7",
-    "geojson-vt": "2.4.0",
+    "geojson-vt": "^3.2.1",
     "gl-mat3": "1.0.0",
     "gl-mat4": "1.1.4",
     "gl-shader-errors": "1.0.3",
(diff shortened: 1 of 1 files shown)


bcamper:
Thanks @meetar! As a practice, we keep dependencies pegged to exact versions (https://github.com/tangrams/tangram/pull/744/commits/ab08178ea724fd061af8f1aeffa0113b9d46b39e), to avoid bugs that slip through in minor/patch upgrades (they aren't supposed to, but we've been bitten by it before, so better to explicitly test each version before upgrading).

bcamper:
Re: text rendering, those don't go through geojson-vt at all (those quads are always generated internally), it's some kind of non-deterministic issue, I think with the text "pixel snapping" feature, that broke for Differ tests at some point, and I have never spent enough time to track it down... but it would be nice to fix, since it's misleading. OTOH, since label *placement* itself is now non-deterministic, as of the addition of main thread / cross-data-source collisions, these tests will continue to be as well, even if the rendering issue is fixed :/

PR #745, opened 2020-01-27
wireframe mode

meetar:
This is a naive approach to wireframe rendering, which manually edits the VBOMesh.element_data indices to produce three line segments for every triangle.

Textures and lighting are disabled, but vertex colors are still applied.

PR for discussion – would be cool to have this as a debugging option.

### DIFF
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -3,7 +3,6 @@ import log from '../utils/log';
 import Utils from '../utils/utils';
 import subscribeMixin from '../utils/subscribe';
 import WorkerBroker from '../utils/worker_broker';
-import debugSettings from '../utils/debug_settings';
 
 // GL texture wrapper object for keeping track of a global set of textures, keyed by a unique user-defined name
 export default class Texture {
@@ -87,7 +86,7 @@ export default class Texture {
     }
 
     bind(unit = 0) {
-        if (!this.valid || debugSettings.wireframe) {
+        if (!this.valid) {
             return;
         }
 
--- a/src/gl/texture.js
+++ b/src/gl/texture.js
@@ -3,6 +3,7 @@ import log from '../utils/log';
 import Utils from '../utils/utils';
 import subscribeMixin from '../utils/subscribe';
 import WorkerBroker from '../utils/worker_broker';
+import debugSettings from '../utils/debug_settings';
 
 // GL texture wrapper object for keeping track of a global set of textures, keyed by a unique user-defined name
 export default class Texture {
@@ -86,7 +87,7 @@ export default class Texture {
     }
 
     bind(unit = 0) {
-        if (!this.valid) {
+        if (!this.valid || debugSettings.wireframe) {
             return;
         }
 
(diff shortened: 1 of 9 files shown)


meetar:
image

src/gl/texture.js
```diff
@@ -86,7 +87,7 @@ export default class Texture {
     }
 
     bind(unit = 0) {
-        if (!this.valid) {
+        if (!this.valid || debugSettings.wireframe) {
```
bcamper: Can you tell me what this check is about? Shouldn't any textures used by the geometry still be valid, even if they are not being used for rendering?

bcamper: I'm getting hundreds of these WebGL errors in Chrome with this:

```
WebGL: INVALID_OPERATION: texImage2D: no texture bound to target
WebGL: INVALID_OPERATION: texParameter: no texture bound to target
[.WebGL-0x7f8b71f16e00]RENDER WARNING: there is no texture bound to the unit 0
```

I'm not getting any errors with this check removed though, e.g. reverted to original code.

src/scene/scene.js
```diff
@@ -1152,6 +1153,11 @@ export default class Scene {
     // Create lighting
     createLights() {
         this.lights = {};
+
+        if (debugSettings.wireframe) {
+            Light.enabled = false; // disable lighting for wireframe mode
```
bcamper: Turns out this is a more succinct way to disable lighting for wireframe mode, rather than having to create a synthetic light.

src/styles/style.js
```diff
@@ -411,7 +411,10 @@ export var Style = {
 
     makeMesh (vertex_data, vertex_elements, options = {}) {
         let vertex_layout = this.vertexLayoutForMeshVariant(options.variant);
-        return new VBOMesh(this.gl, vertex_data, vertex_elements, vertex_layout, options);
+        return new VBOMesh(this.gl,
+            vertex_data, vertex_elements, vertex_layout,
+            { ...options, wireframe: debugSettings.wireframe }
```
bcamper: Moved wireframe option enable here, rather than having it hard-coded inside the `VBOMesh`, which (ideally) is more generic (though wireframe mode is itself hard-coded to specific data/draw mode format...).

src/gl/texture.js
bcamper: OK I think I see what you were trying to achieve with this: to get textured points to render as quads, instead of wireframes masked against a sprite?

I will try to think of another way to accomplish this, but we can't disable texture binding this low-level.

meetar: Yep that's the goal – though tbh I've never needed to inspect sprite geometry, I just wanted to try this for completeness

bcamper: OK this is my attempt to make these more visible, think it works alright: https://github.com/tangrams/tangram/pull/745/commits/da98a78c01051600ed17c7df8c8ab08a4344a088

bcamper:
Thought about this a bit more, decided to move wireframe building code to its own file, to isolate from "normal" code and remove any special wireframe logic from `VBOMesh` entirely.

bcamper:
Thanks @meetar! This is awesome, wish we'd had it ~4 years ago...

PR #758, opened 2020-06-10
make preserve_tiles_within_zoom an option in tile source config

bdon:
I didn't see anywhere `preserve_tiles_within_zoom` was being set other than to 0 for raster layers. Is the intention that this can be customized through the YAML? This PR adds it as an option in the YAML like so:

    sources:
        tilezen:
            type: MVT
            url: https://tile.nextzen.org/tilezen/vector/v1/512/all/{z}/{x}/{y}.mvt
            tile_size: 512
            preserve_tiles_within_zoom: 4

### DIFF
--- a/src/sources/data_source.js
+++ b/src/sources/data_source.js
@@ -26,6 +26,9 @@ export default class DataSource {
             });
         }
 
+        // Optional setting to keep tiles within a zoom range (higher requires more memory)
+        this.preserve_tiles_within_zoom = config.preserve_tiles_within_zoom;
+
         // Optional function to preprocess source data
         this.preprocess = config.preprocess;
         if (typeof this.preprocess === 'function') {


bcamper:
It's been an internal setting, but I don't mind exposing it as an option. The name has always been a bit awkward, so if we're going to make it public... have any suggestions for a simpler one? :) If not we'll keep as is.

Re: the behavior you mentioned seeing: with the default value of 1, I think the expected behavior is that this already *should* retain any tiles that are a full zoom level in/out from current view, and only prune those 2 or more levels out. Are you seeing cases where that doesn't seem to be true? If so we should also investigate. The main reason to not set this value higher is that it can theoretically retain a *lot* of tiles... but I suppose if the rest of the view pruning is working, it should stay manageable.

This also reminds me, @burritojustice has also wanted an option to control the `[view.buffer](https://github.com/tangrams/tangram/blob/master/src/scene/view.js#L40)`, which can be expanded to preload more tiles around the view area. That could be a follow-up to this.

bcamper:
@bdon what are your current thoughts on exposing this publicly, given your most recent gitter comments?

bdon:
Yeah, this can be closed as I'd rather investigate making some changes to TileManager.

Issue #724, opened 2019-06-19
Add support for text highlight boxes


nvkelso:
Similar to https://github.com/tangrams/tangram/issues/723 which proposes text underlines, I'd like to a new text treatment to callout country and regional capitals in map styles. We adding better support for both in the new v1.8 Tilezen schema, but right now the only way to call them out visually is to change the color. Another common technique in cartography (especially in Europe) is to "highlight" the text with a yellow or orange box shape behind the text.

There are two options for the shape – if there is letter spacing (or spaces) the highlight should optionally continue as one shape... but the default is usually to only do it behind word blocks.

bdon:
A cheap way to do this may be to draw a rectangle behind the text when it's being rasterized on canvas. This would depend on getting the metrics right with letter-spacing etc applied so the border around the text is consistent. It would also mean the background cannot have an alpha independent of the text, but that seems like an acceptable compromise.

Example screen:

<img width="272" alt="Screen Shot 2020-06-08 at 3 00 10 PM" src="https://user-images.githubusercontent.com/77501/84001366-03f99480-a999-11ea-9530-4646925d08e0.png">

bcamper:
Funny, I was looking at exactly this last night too (re-familiarizing myself with the text rendering code thinking about vertical CJK text, and remembered this issue).

I agree the Canvas background box approach is a good tradeoff for this (and we use a similar approach for texture/collision box debug rendering already). It's not the most optimized (would increase the label atlas size for different color background boxes that otherwise share the same label text + font), but that is unlikely to happen frequently and it's a low code footprint to implement. It is less straightforward for curved labels though, due to order issues with the background boxes overlapping nearby segments, so limiting it to point labels -- which I believe is the main use case -- would also make for an easier initial implementation.

I've pushed my work on this here: https://github.com/tangrams/tangram/compare/text-box

![tangram-1591579291300](https://user-images.githubusercontent.com/16733/84112017-fcb4a400-a9f5-11ea-86f9-90411c1bb76d.png)

Note that while it does apply the recently added `alpha` parameter in the `font` block (#740) to both the `fill` and `background`, it *is* also possible to set a different alpha for the two directly, e.g.:

```
font:
  fill: [1, 1, 1, 0.75] # white text w/75% opacity
  background: [1, 0.5, 0, 0.5] # orange background w/50% opacity
  ...
```

@nvkelso how does this look to you?

bdon:
This is a much better implementation than mine :) 

[...]

Issue #723, opened 2019-06-19
Add support for underline text style


nvkelso:
I'd like to add underline text treatment to callout country and regional capitals in map styles. We adding better support for both in the new v1.8 Tilezen schema, but right now the only way to call them out visually is to change the color. One common technique in cartography is to underline capital text.

There are two options for the underline – if there is letter spacing (or spaces) the underline should optionally continue as one "line"... but the default is usually to only do it under word blocks. The weight of the line (thickness or stroke) should match the weight of the text.

nvkelso:
Similar text highlight boxes: https://github.com/tangrams/tangram/issues/724

bcamper:
released in v0.21.0

PR #761, opened 2020-07-05
Add support for text highlight boxes + underlines

bcamper:
See #724. Adds support for an optional box behind text labels, which adds the following parameters, in a new `font.background` block:

- Text box **fill color**
  - `font.background.color`
  - Optionally vary alpha independently: `font.background.alpha`
- Text box **stroke color**
  - `font.background.stroke.color`
  - Optionally vary alpha independently: `font.background.stroke.alpha`
- Text box **stroke width**:
  - `font.background.stroke.width`
  - Values in pixels, as with `font.stroke.width`
    - e.g. `font.background.stroke.width: 2px` (explicit) or `font.background.stroke.width: 2` (implicit)
  - Defaults to `1px` when `font.background.stroke.color` is not null
- Text background fill and stroke are both incorporated into the label's collision box.

**Examples**

Fill only:

```
draw:
  text:
    ...
    font:
      ...
      background:
        color: red
```

tangram-1593906832593

Fill and stroke:

```
draw:
  text:
    ...
    font:
      ...
      background:
        color: red
        stroke:
          color: black
          width: 2px
```

tangram-1593906900223

Stroke only:

```
draw:
  text:
    ...
    font:
      ...
      background:
        stroke:
          color: black
```

tangram-1593907009480

All parameters can accept JS functions (alpha and stroke width shown here):

```
draw:
  text:
    ...
    font:
      ...
      background:
        color: red
        alpha: function() { return Math.min(Math.max(feature.kind_tile_rank - 4, 0), 20) / 8 + .35 }
        stroke:
          color: black
          width: function() { return Math.min(Math.max(feature.kind_tile_rank - 4, 1), 8) }
```

tangram-1593907059790

**Open Questions**

- For terminology, I was unsure of the following choices:
  - I used `font.background.color` instead of `font.background.fill`. Generally, `color` is more common than `fill` in Tangram, and the latter always felt like a bit of an outlier. That said, there's an argument for using it here for consistency with `font.fill`.
  - I used `font.background.stroke`, with `font.background.stroke.{color|width}`, which does follow `font.stroke.{color|width}`. An alternative would be `outline`, to match shape outlines for `lines` and `points`.
- I included support for `alpha` overrides, to match recently added functionality for all other color properties. Seems like potential overkill, but it's consistent.
- There's no limit on `font.background.stroke.width`. There's some implementation-specific limit, but for instance `stroke: 64px` does work:
tangram-1593919222805

### DIFF
--- a/src/styles/style_parser.js
+++ b/src/styles/style_parser.js
@@ -494,6 +494,18 @@ StyleParser.evalCachedColorProperty = function(val, context = {}) {
     }
 };
 
+// Evaluate color cache object and apply optional alpha override (alpha arg is a single value cache object)
+StyleParser.evalCachedColorPropertyWithAlpha = function (val, alpha_prop, context) {
+    const color = StyleParser.evalCachedColorProperty(val, context);
+    if (color != null && alpha_prop != null) {
+        const alpha = StyleParser.evalCachedProperty(alpha_prop, context);
+        if (alpha != null) {
+            return [color[0], color[1], color[2], alpha];
+        }
+    }
+    return color;
+};
+
 StyleParser.parseColor = function(val, context = {}) {
     if (typeof val === 'function') {
         val = tryEval(val, context);
--- a/src/styles/style_parser.js
+++ b/src/styles/style_parser.js
@@ -223,6 +223,10 @@ StyleParser.createPointSizePropertyCache = function (obj, texture) {
 };
 
 StyleParser.evalCachedPointSizeProperty = function (val, sprite_info, texture_info, context) {
+    if (val == null) {
+        return;
+    }
+
     // no percentage-based calculation, one cache for all sprites
     if (!val.has_pct && !val.has_ratio) {
         return StyleParser.evalCachedProperty(val, context);
@@ -420,7 +424,10 @@ StyleParser.colorForString = function(string) {
 // (caching the result for future use)
 // { value: original, static: [r,g,b,a], zoom: { z: [r,g,b,a] }, dynamic: function(){...} }
 StyleParser.evalCachedColorProperty = function(val, context = {}) {
-    if (val.dynamic) {
+    if (val == null) {
+        return;
+    }
+    else if (val.dynamic) {
         let v = tryEval(val.dynamic, context);
 
         if (typeof v === 'string') {
--- a/src/styles/text/text_canvas.js
+++ b/src/styles/text/text_canvas.js
@@ -181,50 +181,65 @@ export default class TextCanvas {
     }
 
     // Draw multiple lines of text
-    drawTextMultiLine (lines, [x, y], size, { stroke, stroke_width = 0, transform, align, supersample }, type) {
-        let line_height = size.line_height;
+    drawTextMultiLine (lines, [x, y], size, { stroke, stroke_width = 0, background, transform, align, supersample }, type) {
+        // optional background box
+        if (background) {
+            const texture_size = size.texture_size;
+            const dpr = Utils.device_pixel_ratio * supersample;
+            const horizontal_buffer = dpr * (this.horizontal_text_buffer + stroke_width);
+            const vertical_buffer = dpr * this.vertical_text_buffer;
+            const collision_size = size.collision_size;
+            const box_buffer = 4 * dpr; // pixel buffer on each side of label box
+
+            this.context.save();
+            this.context.fillStyle = background;
+            this.context.fillRect(
+                x + horizontal_buffer + (type === 'curved' ? texture_size[0] : 0) - box_buffer,
+                y + vertical_buffer - box_buffer,
+                dpr * collision_size[0] + box_buffer * 2,
+                dpr * collision_size[1] + box_buffer * 2
+            );
+            this.context.restore();
+        }
+
+        // draw text
         let height = y;
         for (let line_num=0; line_num < lines.length; line_num++) {
             let line = lines[line_num];
             this.drawTextLine(line, [x, height], size, { stroke, stroke_width, transform, align, supersample }, type);
-            height += line_height;
+            height += size.line_height;
         }
 
         // Draw bounding boxes for debugging
         if (debugSettings.draw_label_collision_boxes) {
+            const dpr = Utils.device_pixel_ratio * supersample;
+            const horizontal_buffer = dpr * (this.horizontal_text_buffer + stroke_width);
+            const vertical_buffer = dpr * this.vertical_text_buffer;
+            const collision_size = size.collision_size;
+            const lineWidth = 2;
+
             this.context.save();
-
-            let dpr = Utils.device_pixel_ratio * supersample;
-            let horizontal_buffer = dpr * (this.horizontal_text_buffer + stroke_width);
-            let vertical_buffer = dpr * this.vertical_text_buffer;
-            let collision_size = size.collision_size;
-            let lineWidth = 2;
-
             this.context.strokeStyle = 'blue';
             this.context.lineWidth = lineWidth;
             this.context.strokeRect(x + horizontal_buffer, y + vertical_buffer, dpr * collision_size[0], dpr * collision_size[1]);
             if (type === 'curved'){
                 this.context.strokeRect(x + size.texture_size[0] + horizontal_buffer, y + vertical_buffer, dpr * collision_size[0], dpr * collision_size[1]);
             }
-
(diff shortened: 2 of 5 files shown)


bcamper:
cc @nvkelso @bdon @meetar @matteblair @tallytalwar thoughts on functionality, syntax, and naming appreciated!

matteblair:
Nice! I'm excited to use this to make some sick Supreme-branded maps.
- How are the dimensions of the background box determined from the text layout?
- What happens for curved labels?

nvkelso:
Most excellent!

bdon:
Looking good!

* IMHO curved labels can be excluded from this
* I'm getting a `WebGL: INVALID_VALUE: texImage2D: no canvas` when there is no `stroke` key defined in background
* I think there should be a way to control the padding of the background box.

nvkelso:
Padding of background box?

For curved text, I agree can be for later. Usually there is option to stop
the box behind whole string, words, and characters when a threshold of
letter space or word space is reached.

bcamper:
OK, getting back to this again and trying to wrap up...

- **Which labels apply:** Sorry if this was lost here, but based on discussion in original issue #724, the approach here was to limit to **point labels only**, e.g. exclude both curved and angled straight line labels. Curved labels present the implementation difficulty here, so angled straight labels (e.g. usually straight road labels) were excluded largely to match the omission of curved labels, since they are usually labelling the same set of features, and it seemed it would be weird to have it on some but not others. This could be revisited in the future, but I wanted to manage this initial implementation complexity, and believe point labels are the primary use case. cc @matteblair 

- **Background box size:** The background box dimensions are currently calculated as `4px` on each side of the text, but in practice as you can see from screenshots, it creates a bit more padding vertically than horizontally, because font heights extend above. This size was mostly arbitrary, we can adjust the default if people have other suggestions.

- **Background box size configurability:** I was initially going to suggest we defer making the background box dimension configurable, because there's already a lot going on here and I don't want to re-create the CSS box model ;) However, it was only a few lines of code now that everything else is setup, so... here you go, e.g. `font.background.width = 8px` 6cf2090.

- **Text underlines:** I know I said I wanted to manage complexity... I do, but I was already in process of adding the companion feature to this, #723. I've added that here now, with similar syntax. As with the text box, a lot these calculations for positioning and spacing have just been manually tweaked here until they more or less look right under a variety of conditions, because canvas text font metrics are not complex and/or widely available enough to support all these cases. Example:
  ```
  draw:
    text:
      font:
        underline:
          color: blue
          # alpha: 0.5 # optional alpha override
          width: 2px # defaults to 1px if only `color` is defined
  ```
  Screen Shot 2020-07-11 at 4 22 33 PM
  Optionally combined w/background box:
  Screen Shot 2020-07-11 at 4 23 06 PM

- @bdon I believe this should have been addressed by e0d3bde, if you're still seeing on this branch, could you please send me a scene YAML reduction example I can work off of.
  > I'm getting a WebGL: INVALID_VALUE: texImage2D: no canvas when there is no stroke key defined in background

As always, we can make further enhancements or adjustments in the future, but given the scope of this PR, I'd love to wrap it up and proceed. There looks to be general consensus. If you have any further comments on **default background box sizing**, **text underlines**, or anything else, please speak up now and then we can get this into Tangram v0.21.0! 🚀

matteblair:
Thanks for the clarifications 👍 

A few more questions:
- The `stroke` on the background boxes: Does this follow the SVG model of "stroke" where the line extends both into and out of the shape? If so, is the specified `width` just the exterior extent or the full width?
- How are the elements in these formations blended together when multiple colors have transparency?
- So `font.background.width` and `font.background.height` specify the padding that's added to both sides of each axis? Should we call it `padding_width`? Maybe `padding` by itself to set both? Hm. The possibilities start to get out of hand pretty quickly...
- Is there a way to add an underline that inherits the font color? Would `underline` with a `width` but no `color` do that?

nvkelso:
Fantastic!

bcamper:
> * The `stroke` on the background boxes: Does this follow the SVG model of "stroke" where the line extends both into and out of the shape? If so, is the specified `width` just the exterior extent or the full width?

In this case, the full stroke is drawn "outside of"/beyond the background box edge, so the stroke `width` does not affect the visible portion of the background box fill, and the visible portion of the stroke around the background box should be the full `width`. So the full width of the box should be:

`stroke_width + background_width + text_width + background_width + stroke_width` =
`text_width + 2 * (stroke_width + background_width)`

Where `text_width` is implementation-specific between Tangram JS/ES, with some small variations due to different text rendering technologies. (And same applies for height as `text_height + 2 * (stroke_width + background_width)`.)

> * How are the elements in these formations blended together when multiple colors have transparency?

Since for Tangram JS these elements are all drawn as part of the label texture, they are composited using Canvas default `source-over` blending with ["simple alpha compositing"](https://drafts.fxtf.org/compositing/#simplealphacompositing), which if I'm reading right looks equivalent to `glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)`. That texture is then composited with the Tangram scene, with the same blending when the default Tangram `overlay` blend mode is used.

> * So `font.background.width` and `font.background.height` specify the padding that's added to both sides of each axis? Should we call it `padding_width`? Maybe `padding` by itself to set both? Hm. The possibilities start to get out of hand pretty quickly...

Only `font.background.width` is currently defined (no `font.background.height`), because it was intended as _stroke width_ of the line being drawn -- so it applies to both the sides and top/bottom of the box (see notes on box dimensions above). However, this misunderstanding confirms my discomfort with `width` in this context.

**I like `padding` instead**, it's a new term for Tangram but fits the CSS definition. I also don't think we need support for specifying different padding for width vs. height at this point, so I would suggest we stick with a single scalar for now, e.g. `padding: 8px`. If we wanted to expand this to a 2D value in the future, we would simply start accepting array values as we do for similar props like `size`, `offset`, etc.

Similarly, I haven't loved the use of `stroke` here overall. `outline` would be another Tangram term that fits, e.g. from `line.outline`. Or, if we use `padding`, the CSS equivalent for this would be `border`. I don't have strong feelings on this though.

> * Is there a way to add an underline that inherits the font color? Would `underline` with a `width` but no `color` do that?

There is not but that makes sense, I think that's a better default, to have `underline.color` inherit from `font.fill`.

In this case, would there then **not** be a default for `underline.width` (currently `1px`)? Or would we continue to have a default `width` but require the `underline` block to specify _either_ `width` _or_ `color`? Otherwise, you get an empty `underline: {}` object drawing an underline, and that feels a little _too_ implicit to be intuitive. Maybe I'm overcomplicating things though? I don't have strong opinions on this one either.

matteblair:
- Those stroke metrics make sense to me. I think for points strokes do the inside/outside thing so I wanted to make sure.
- Compositing makes sense too. In Tangram ES we'll probably implement this as two separate primitives, but I think as long as it's all using the same blending technique it should be equivalent.. 
- I totally misinterpreted `background.width` haha. `padding` sounds good to me.
- I feel like underlines inheriting text color is worth supporting - or at least that its absence might be frustrating. Seems like underlines with a default width and color would cover the vast majority of use cases. Maybe we can just start with `underline: true`?

bdon:
> @bdon I believe this should have been addressed by e0d3bde

Yes, the background boxes are fixed now for my scene.yaml.

bcamper:
Makes sense to me to use `underline: true` to indicate defaults for inherited font color and default stroke width (`1px`? may be a bit too narrow), and follows similar Tangram conventions. I had been thinking of it as a shortcut, rather than removing the option to set the `color` and `width` properties.

@nvkelso @bdon see above, @matteblair is proposing we start with fixed values for text underline color and width, to manage complexity (especially as they are not 100% sure yet how they would implement this in Tangram ES). Does this make sense as an initial limitation to you? @matteblair pointed out that even most text editing software doesn't offer that level of customization. The code is already written on the Tangram JS side, so not a big difference for me either way :)

nvkelso:
I like `underline: true` with some sensible default for the line weight (1, 1.5, 2?) and inheriting the line color from the text's color. I think that boolean matches what we might see for bold or italic.

bdon:
+1 to fixed values for underline color and width

bcamper:
Sounds good! I set the underline width at `1.5px` which I think gives good results across device pixel ratios. We can adjust if needed.
Screen Shot 2020-07-20 at 9 10 59 PM

PR #763, opened 2020-07-13
If no explicit attribution in leafletLayer, use attribution in YAML sources

bdon:
This is more similar to the behavior of Leaflet raster tiles where each individual tile layer can have an attribution. Because a TangramLayer can use multiple tile sources, this lets the attribution for each source be specified in YAML, and then these are added together in the case that no explicit 'attribution' prop is passed to LeafletLayer.

### DIFF
--- a/src/leaflet_layer.js
+++ b/src/leaflet_layer.js
@@ -166,6 +166,12 @@ function extendLeaflet(options) {
                         blocking: false
                     }).then(() => {
 
+                    if (!this.options.attribution) {
+                        for (const [key, value] of Object.entries(this.scene.config.sources)) {
+                            if (value.attribution) map.attributionControl.addAttribution(value.attribution)
+                        }
+                    }
+
                     this._updating_tangram = true;
 
                     this.updateSize();
--- a/src/leaflet_layer.js
+++ b/src/leaflet_layer.js
@@ -167,8 +167,10 @@ function extendLeaflet(options) {
                     }).then(() => {
 
                     if (!this.options.attribution) {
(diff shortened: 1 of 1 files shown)


bcamper:
I like this, it reminds me that I have wanted to have more formal scene-level metadata properties defined (though source-level makes sense here).

@matteblair @tallytalwar I don't know if there's anything we would need explicitly here on the ES side, minimally to expose these values in the API? We don't otherwise have a built-in attribution display do we?

matteblair:
Tangram ES still doesn't have any integrated form of attribution - but we should! Per-source and per-scene attribution strings make sense to me. We can certainly ingest those strings from the scene configuration, we just don't have anywhere to show them right now. A simple approach could be to create a native text window for the attribution and let the user customize it from their app.

tallytalwar:
I like it too.. I have a vague memory for doing this or at least discussing
approaches for eraser-map..

bcamper:
I made some very minor changes in 6ee8598 to match Tangram code style (some aspects of which I wouldn't pick for a new project today, but... it's the style we have aimed for consistency in for the project's lifespan).

bcamper:
Switched Nextzen demo attribution to use this in 365bb6e65d6bc6713123de908231984e587d73c3 🙌

nvkelso:
Could you also add in `Data from OpenStreetMap and Who's On First`, as well?

bcamper:
> Could you also add in `Data from OpenStreetMap and Who's On First`, as well?

Done! We should add directly to the Nextzen basemaps as well.

